PR #9, opened 2024-08-28
health dashboards: remove path templatization logic

Augustyniak:
The plan is to make this logic remote configuration configurable and move the actual implementation of the feature to shared-core repo instead.

### DIFF
--- a/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/network/HttpResponseInfo.kt
+++ b/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/network/HttpResponseInfo.kt
@@ -37,6 +37,9 @@ data class HttpResponseInfo @JvmOverloads constructor(
 
     internal val fields: InternalFieldsMap =
         run {
+            // Collect out-of-the-box fields specific to HTTPResponse logs. The list consists of
+            // HTTP specific fields such as host, path, or query and HTTP request performance
+            // metrics such as DNS resolution time.
             val fields = buildMap {
                 this.put(SpanField.Key.TYPE, FieldValue.StringField(SpanField.Value.TYPE_END))
                 this.put(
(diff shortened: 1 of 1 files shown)


platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/network/HttpResponseInfo.kt
```diff
@@ -36,60 +35,75 @@ data class HttpResponseInfo @JvmOverloads constructor(
...
+                    )
+                    this.put(
+                        "_response_headers_bytes_count",
+                        FieldValue.StringField(it.responseHeadersBytesCount.toString()),
+                    )
+                    putOptional("_dns_resolution_duration_ms", it.dnsResolutionDurationMs)
+                }
             }
+            // If the path between request and response did not change and an explicit path
+            // template was provided as part of a request use it as path template on a response.
```
murki, line 101: does this comment belong here?

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/network/HttpResponseInfo.kt
```diff
@@ -36,60 +35,75 @@ data class HttpResponseInfo @JvmOverloads constructor(
...
+                            request.path.template
+                        } else {
+                            null
+                        }
 
-                @Suppress("SwallowedException")
-                val normalized: String? = requestPathTemplate?.let { it }
-                    ?: it.template?.let { it }
-                    ?: try {
-                        CaptureJniLibrary.normalizeUrlPath(it.value)
```
murki, line 67: it's great to be able to get rid of this call

platform/jvm/capture/src/test/kotlin/io/bitdrift/capture/CaptureLoggerTest.kt
```diff
@@ -125,7 +125,6 @@ class CaptureLoggerTest {
             "_host" to "api.bitdrift.io",
             "_method" to "GET",
             "_path" to "/my_path/12345",
-            "_path_template" to "/my_path/<id>",
```
murki, line 128: can we now remove the `CaptureJniLibrary.load()` and maybe even make it a plain unit test (vs robolectric one)?

murki (approved):

PR #12, opened 2024-08-28
add app launch TTI log

Augustyniak:
Add predefined App Launch TTI log that can be emitted by customers using Capture SDK. In the future, we will create Capture SDK integration that emits this log automatically, in the current version we expect Capture SDK customers to call the method manually.

### DIFF
--- a/test/platform/swift/unit_integration/mocks/MockCoreLogging.swift
+++ b/test/platform/swift/unit_integration/mocks/MockCoreLogging.swift
@@ -124,7 +124,7 @@ extension MockCoreLogging: CoreLogging {
         self.logAppUpdateExpectation?.fulfill()
     }
 
-    func logAppLaunchTTI(_ duration: TimeInterval) {}
+    func logAppLaunchTTI(_: TimeInterval) {}
 
     func addField(withKey _: String, value _: String) {}
 
--- a/test/platform/swift/unit_integration/mocks/MockLoggerBridging.swift
+++ b/test/platform/swift/unit_integration/mocks/MockLoggerBridging.swift
@@ -94,7 +94,7 @@ extension MockLoggerBridging: LoggerBridging {
         self.logAppUpdateExpectation?.fulfill()
     }
 
-    func logAppLaunchTTI(_ duration: TimeInterval) {}
+    func logAppLaunchTTI(_: TimeInterval) {}
 
     func addField(withKey _: String, value _: String) {}
 
--- a/test/platform/swift/unit_integration/mocks/MockLogging.swift
+++ b/test/platform/swift/unit_integration/mocks/MockLogging.swift
@@ -109,7 +109,7 @@ extension MockLogging: Logging {
         self.onLog(log)
     }
 
-    func logAppLaunchTTI(_ duration: TimeInterval) {}
+    func logAppLaunchTTI(_: TimeInterval) {}
 
     func addField(withKey _: String, value _: FieldValue) {}
 


examples/android/MainActivity.kt
```diff
@@ -121,6 +124,8 @@ class MainActivity : ComponentActivity() {
 
         Logger.addField("field_container_field_key", "field_container_field_value")
         Logger.logInfo(mapOf("key" to "value")) { "MainActivity onCreate called" }
+
+        Logger.logAppLaunchTTI(1.toDuration(DurationUnit.SECONDS))
```
murki, line 128: this is not accurate /trolling

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/CaptureJniLibrary.kt
```diff
@@ -229,6 +229,19 @@ internal object CaptureJniLibrary {
...
+     * Writes an app launch TTI log. The method should be called only once per logger Id. Consecutive calls
+     * have no effect.
+     *
+     * @param loggerId the ID of the logger to write to.
+     * @param durationS the time between a user's intent to launch the app and when the app becomes
+     *                  interactive.
+     */
+    external fun writeAppLaunchTTILog(
+        loggerId: Long,
+        durationS: Double,
```
murki, line 242: why the `S` in the param name? Is it in Seconds?

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/ILogger.kt
```diff
@@ -72,6 +73,7 @@ interface ILogger {
 
     /**
      * Logs a message at a specified level.
+     *
```
murki, line 76: fighting the good fight

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/CaptureJniLibrary.kt
murki: I see below that it is. Is this a standard, to use seconds in Doubles to track time? and if so we should add the seconds unit part to the docstring

platform/shared/src/lib.rs
```diff
@@ -158,6 +162,39 @@ impl LoggerHolder {
     holder.shutdown(false);
     drop(holder);
   }
+
+  /// Logs an out-of-the-box app launch TTI log event. The method should be called only once.
+  /// Consecutive calls have not effect.
+  pub fn log_app_launch_tti(&self, duration: time::Duration) {
+    self.app_launch_tti_log.call_once(|| {
```
murki, line 169: I'm just learning about this primitive and their docs say 
>This method will block the calling thread if another initialization routine is currently running.
which seems to be aimed to help with
>It is also guaranteed that any memory writes performed by the executed closure can be reliably observed by other threads at this point

This sounds like an overkill to me since there are no shared resources that we'd like to guard, but maybe I'm misunderstanding something

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/CaptureJniLibrary.kt
murki: also worth mentioning that the value should be positive otherwise will be ignored

platform/shared/src/lib.rs
```diff
@@ -158,6 +162,39 @@ impl LoggerHolder {
...
+          key: "_duration_ms".into(),
+          value: duration_ms.to_string().into(),
+        },
+        kind: LogFieldKind::Ootb,
+      }];
+
+      self.log(
+        log_level::INFO,
+        LogType::Lifecycle,
+        "AppLaunchTTI".into(),
```
murki, line 190: It could be argued that this value should be just a field in the existing app launch event, but I see how exposing it in our public api would make that complicated, so I guess this will have to do for now

murki (approved):

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/CaptureJniLibrary.kt
murki: Anything that an API allows users will do. My opinion is that we should never allow for invalid operations using the type system itself e.g. our Span API doesn't allow negative durations. 

So at least we should document that this particular usage is invalid

platform/shared/src/lib.rs
murki: sg

PR #24, opened 2024-09-04
Fix `ProvidersTest#network_attributes()` test `TODO`

pablo-guardiola:
As I continue my journey exploring `capture-sdk`'s repo / learning how bitdrift's Capture SDK works, I noticed that `ProvidersTest` had an _easy_ `TODO` that I could tackle, so this is a quick fix that addresses that tech debt. While working on that, I also went ahead and refactored `ProvidersTest` and added some more coverage to `ClientAttributes`, `DeviceAttributes` and `NetworkAttributes`.

---

Fixes `ProvidersTest#network_attributes()` test `TODO`:

- Removes `@Ignore` annotation
- Refactors `ClientAttributes`, `DeviceAttributes` and `NetworkAttributes` tests out of `ProvidersTest` and into their respective test classes
- Adds more coverage to `ClientAttributes`, `DeviceAttributes` and `NetworkAttributes`
- Bumps Robolectric dependency version to [`4.13`](https://github.com/robolectric/robolectric/releases/tag/robolectric-4.13)

---

Refactored the tests to have only 1 assertion per test and one test per scenario 👀 https://speakerdeck.com/guardiola31337/elegant-unit-testing-mobilization-2016?slide=57

Basically, if one of them fails you'll know at a glance what in the code might have caused that failure and in which part (class, method, etc.).

The problems of having multiple assertions are:
- It's annoying to have to look at the stacktrace to know what's failing (which assertion failed?)
- You don't have information about the remaining assertions (because when an assertion fails it breaks the execution of the test)

I rarely write tests with multiple assertions (at least I try 😅), only if they're tearing apart the result object to verify everything e.g. if there's no `equals` (which is another problem although could happen if it's coming from a 3rd party library object) or if the assertions are atomic to the object.

This is a technique that I learnt from the book [Working Effectively With Unit Tests](https://leanpub.com/wewut). I really like the techniques described on it in favor of DAMP (Descriptive And Maintainable Procedures). Mainly because they increase maintainability and readability of tests. Another great one (already followed in the project ❤️) is to structure the tests following the triple `A` - `A`rrange / `A`ct / `A`ssert (I removed the explicit comments as IMO are unnecessary).

BTW, I highly recommend you reading that book (if you haven't already). It's 🔝 💯


pablo-guardiola:

murki:
Thanks @pablo-guardiola ! I'll do a quick pass and take a closer look tomorrow

platform/jvm/capture/src/test/kotlin/io/bitdrift/capture/ProvidersTest.kt
```diff
@@ -7,120 +7,22 @@
...
-
-        // ACT
-        val clientAttributes = ClientAttributes(context, lifecycleOwner).invoke()
-
-        // ASSERT
-        assertThat(clientAttributes).containsEntry("foreground", "0")
-    }
-
-    @Test
-    @Ignore("Robolectric throwing ClassNotFoundException: android.telephony.TelephonyCallback")
```
murki, line 103: what was the actual fix of the TODO/failure? Was it just bumping Robolectric to 4.13?

platform/jvm/gradle/libs.versions.toml
```diff
@@ -19,7 +19,7 @@ mockitoCore = "4.9.0"
 mockitoKotlin = "2.2.0"
 mockitoKotlinVersion = "4.1.0"
 okhttp = "4.12.0"
-robolectric = "4.9"
```
murki, line 22: Until we finish the migration from bazel to gradle the integration (robolectric) tests are mostly run via CI using bazel, which means you might need to also bump https://github.com/bitdriftlabs/capture-sdk/blob/a5e8d0fbd98d219be8998148ff31e5b5cc0bc3d7/bazel/capture_dependencies.bzl#L21

As you can tell the versions were already out of sync :(

murki:
There might be a bug in our CI checks where we didn't trigger the android jobs cause we didn't detect any "meaningful" android files had changed, will need to take a closer look

platform/jvm/capture/src/test/kotlin/io/bitdrift/capture/ProvidersTest.kt
pablo-guardiola: That's correct. That issue was fixed in https://github.com/robolectric/robolectric/pull/7786 and released as part of [Robolectric `4.10`](https://github.com/robolectric/robolectric/releases/tag/robolectric-4.10) - it seems that the test was marked as ignored at some point and never revisited ¯\\\_(ツ)\_/¯

That being said, I bumped to the latest available to avoid `ClientAttributesTest#package_info_android_tiramisu()`'s issue `Skip package_info_android_tiramisu because Robolectric doesn't support legacy resources mode after P` (which gets fixed using Robolectric 4.11+).

platform/jvm/gradle/libs.versions.toml
pablo-guardiola: Great catch! I tested both Bazel (`./bazelw test //platform/jvm/...`) and Gradle locally. I was lucky that Robolectric's Bazel version was set to `4.11` and everything was passing 😅 (refs. https://github.com/bitdriftlabs/capture-sdk/pull/24#discussion_r1744750393). Will bump Bazel dependency version to `4.13` to have them in sync 👍

pablo-guardiola: > Will bump Bazel dependency version to 4.13 to have them in sync 👍

Done 👉 https://github.com/bitdriftlabs/capture-sdk/pull/24/commits/b3ac74c4cd4cd66ef0fc9d1469a5740447ee411c

pablo-guardiola:
> There might be a bug in our CI checks where we didn't trigger the android jobs cause we didn't detect any "meaningful" android files had changed

Great catch @murki 👍 

`files_changed` script is failing for external contributions. As you pointed out, root cause seems having `origin` hardcoded 👀 https://github.com/bitdriftlabs/capture-sdk/blob/a5e8d0fbd98d219be8998148ff31e5b5cc0bc3d7/ci/files_changed.sh#L10 

Another great point you brought was that the script is currently implemented in "fail-open" mode rather than "fail-close". That should be fixed as well to report red if something fails - currently it feels that everything worked as the checks passed.

Will try to take a stab at these and report back here.

murki:
> Will try to take a stab at these and report back here.
Thanks @pablo-guardiola, yeah ideally we should make the fixes on this same PR so we can progressively test that they work.
### Re: "fail-closed":
- Remove the true at [the end of the github run step command](https://github.com/bitdriftlabs/capture-sdk/blob/a5e8d0fbd98d219be8998148ff31e5b5cc0bc3d7/.github/workflows/android.yaml#L28)
- Potentially simplify [ci/files_changed.sh](https://github.com/bitdriftlabs/capture-sdk/blob/a5e8d0fbd98d219be8998148ff31e5b5cc0bc3d7/ci/files_changed.sh) so it's more explicit. Something like this:
```bash
if git rev-parse --abbrev-ref HEAD | grep -q ^main$; then
  exit 0
else
  git diff --name-only "origin/$GITHUB_BASE_REF" | grep -E "$1"
fi
```
### Re: external contributions:
- We might need to fiddle with the [ci/files_changed.sh](https://github.com/bitdriftlabs/capture-sdk/blob/a5e8d0fbd98d219be8998148ff31e5b5cc0bc3d7/ci/files_changed.sh) script further to make it recognize forks not in `origin`. The final result could look something like this:
```bash
set -euo pipefail

# Fetch the base branch from the original repository using its URL
git fetch "https://github.com/${GITHUB_REPOSITORY}.git" "${GITHUB_BASE_REF}"

# Check if the current branch is `main`, exit if it is
if git rev-parse --abbrev-ref HEAD | grep -q ^main$; then
  exit 0
fi

# Check if relevant files have changed between the current branch and the base branch
git diff --name-only "FETCH_HEAD" | grep -E "$1"
```

pablo-guardiola:
Hey @murki thanks for the suggestions! They're exactly what I was working on and the same approach I ended up taking.

I've just pushed the changes in https://github.com/bitdriftlabs/capture-sdk/pull/24/commits/d2fc88247545e6cf4a05c54ba3cd01e6d3997e13 Could you please kick-off CI when you have a chance so we can see if everything is working now?

pablo-guardiola:
https://github.com/bitdriftlabs/capture-sdk/actions/runs/10741877498/job/29793722002?pr=24 

```
Run ./ci/check_bazel.sh //examples/android:android_app || ./ci/files_changed.sh .github/workflows/android.yaml || ./ci/files_changed.sh "^platform/jvm/gradle-test-app/.*\.(gradle|kts|kt|xml)$" && ./ci/run_tests.sh
  ./ci/check_bazel.sh //examples/android:android_app || ./ci/files_changed.sh .github/workflows/android.yaml || ./ci/files_changed.sh "^platform/jvm/gradle-test-app/.*\.(gradle|kts|kt|xml)$" && ./ci/run_tests.sh
  shell: /usr/bin/bash -e {0}
fatal: Not a valid object name origin/pg-fix-providers-network-attributes-test
.github/workflows/android.yaml
Tests will run
```

Looking `fatal: Not a valid object name origin/pg-fix-providers-network-attributes-test` 🤔

pablo-guardiola:
I think I know what's going on. It seems that we're not fetching the remote base branch. Apparently, even though `fetch-depth: 0` does fetch the entire history of the repository, it doesn't automatically fetch all remote branches so fetching it explicitly may fix it. Let me quickly try that.

pablo-guardiola:
I've just pushed @murki https://github.com/bitdriftlabs/capture-sdk/pull/24/commits/6ac9cb8fc41a25fd41ca0866bcbfda802b34d5e5

pablo-guardiola:
@murki got some time to implement the "fail-closed" mode properly and now should be working fine (including external contributions / forks) 🚀 

Please, give it a try when you have a chance 🙏

murki (approved):
Thanks @pablo-guardiola for your contributions and your steadfastness towards improving our CI along the way. Changes look good to me

PR #25, opened 2024-09-05
[sdk] Include 3XX http responses in success results

murki:


mattklein123 (approved):
LGTM but I would run it by @Reflejo to see if he has any opinions.

PR #32, opened 2024-09-10
[ci] Fix `files_changed` `merge` main issues and `files_changed` and `check_bazel` script's `base_sha` to point to `HEAD~1`

pablo-guardiola:
Follow up from https://github.com/bitdriftlabs/capture-sdk/pull/24 and https://github.com/bitdriftlabs/capture-sdk/pull/28

This PR fixes:

- https://github.com/bitdriftlabs/capture-sdk/actions/runs/10781042749/job/29898099793 👇

```
Check for workflow file changes
Run ./ci/files_changed.sh .github/workflows/android.yaml
fatal: ambiguous argument 'origin/': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
No relevant changes found.
```

```
Check for relevant Gradle changes
Run ./ci/files_changed.sh "^platform/jvm/gradle-test-app/.*\.(gradle|kts|kt|xml)$"
fatal: ambiguous argument 'origin/': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
No relevant changes found.
```

This is solved by fixing `files_changed` to differentiate when you are in a PR or `main`, as well as ensuring `git diff` failures are caught and reported accordingly.

- `check_bazel` and `files_changed` script's `base_sha` to point to `HEAD~1` when running on a `main` merge. I noticed that no changes were detected even though they were present as part of the PR. Root cause was that we were diffing pointing to the same references and hence no changes. Pointing to the commit before the current one (`HEAD~1`) fixes this. I've thoroughly tested locally and now changes are detected too after merging a PR.

cc @murki

### DIFF
--- a/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/Capture.kt
+++ b/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/Capture.kt
@@ -320,9 +320,9 @@ object Capture {
                 val result = block()
                 span?.end(SpanResult.SUCCESS)
                 return result
-            } catch (e: Throwable) {
+            } catch (exception: Throwable) {
                 span?.end(SpanResult.FAILURE)
-                throw e
+                throw exception
             }
         }
 
--- a/.github/workflows/android.yaml
+++ b/.github/workflows/android.yaml
@@ -21,6 +21,7 @@ jobs:
         with:
           fetch-depth: 0
 
+      # Check for Bazel build file changes
       - name: Check for Bazel build file changes
         id: bazel_check
         run: ./ci/check_bazel.sh //examples/android:android_app
--- a/ci/check_bazel.sh
+++ b/ci/check_bazel.sh
@@ -17,7 +17,7 @@ if [ -n "$GITHUB_BASE_REF" ]; then
   base_sha=$(git rev-parse "$GITHUB_BASE_REF")
 else
   echo "Not in a pull request, skipping base ref fetch."
-  base_sha=$(git rev-parse HEAD)
+  base_sha=$(git rev-parse HEAD~1)
 fi
 
 # Get the latest commit SHA for the PR branch (the head ref in the forked repository)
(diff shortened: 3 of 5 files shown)


murki:
Thanks @pablo-guardiola , would you mind add some inconsequential code change somewhere that would be picked-up by `files_changed.sh` to verify it works? I can then approve and run the checks

pablo-guardiola:
@murki 
> would you mind add some inconsequential code change somewhere that would be picked-up by files_changed.sh to verify it works? I can then approve and run the checks

I already did that locally but I can push some "empty" changes for you to double check, not a problem.

It's important that the changes land on `main` to test if everything works after merging PRs and I didn't want to pollute the repo's commit history ¯\\\_(ツ)\_/¯

But you can probably remove them quickly after confirming 👍

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/Capture.kt
```diff
@@ -83,6 +83,8 @@ object Capture {
             dateProvider: DateProvider? = null,
             apiUrl: HttpUrl = defaultCaptureApiUrl,
         ) {
+            // TODO Adding a no-op for testing, please remove after confirming everything is good
+            val test = true
```
murki, line 87: this might break detekt, plus I agree with your statement about not polluting the repo with these and TODOS since they will go into `main`. How about something inconsequential that can actually stay in teh codebase? Something like a variable rename for example

pablo-guardiola: Sounds good to me. Let me quickly fix 👍

pablo-guardiola: Done 👉 https://github.com/bitdriftlabs/capture-sdk/pull/32/commits/ecd11acfd5c74888a7285fef6753700982d81bf5

murki (approved):

PR #40, opened 2024-09-11
update NDK version to 27.1.12297006

Augustyniak:
Updating NDK version to 27.1.12297006 since GH runners dropped 27.0.12077973 in https://github.com/actions/runner-images/releases/tag/ubuntu22%2F20240908.1


murki (approved):
welp, that was fast... just don't forget to revert the check.sh change

PR #47, opened 2024-09-24
document .sdk_version string in workflow prompt

snowp:


murki (approved):

.github/workflows/release.yaml
```diff
@@ -4,7 +4,7 @@ on:
   workflow_dispatch:
     inputs:
       version:
-        description: 'The new version to release, ex: 0.12.0'
+        description: 'The new version to release, ex: 0.12.0. This must match the value in platform/shared/.sdk_version'
```
murki, line 7: might as well update it below in the `workflow_call` branch as well

PR #52, opened 2024-09-27
handle logger creation failures

Augustyniak:
Make platform layer handle cases where a Rust call to create a logger fails.

The way the logic works is that after the first time `Logger.configure` fails the consecutive attempts to configure the logger are no-ops. There is a public API change introduces in this PR on Android. `Logger.configure(...)` method returns an optional `LoggerIntegrator` object that used to be non-optional. For this reason, the next release will be 0.15.0.

Fixes BIT-3752

### DIFF
--- a/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/Capture.kt
+++ b/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/Capture.kt
@@ -147,26 +147,24 @@ object Capture {
             }
 
             // Ideally we would use `getAndUpdate` in here but it's available for API 24 and up only.
-            when (default.getAndSet(LoggerState.ConfigurationStarted)) {
-                is LoggerState.NotConfigured -> {
-                    try {
-                        val logger = LoggerImpl(
-                            apiKey = apiKey,
-                            apiUrl = apiUrl,
-                            fieldProviders = fieldProviders,
-                            dateProvider = dateProvider ?: SystemDateProvider(),
-                            configuration = configuration,
-                            sessionStrategy = sessionStrategy,
-                            bridge = bridge,
-                        )
-                        default.set(LoggerState.Configured(logger))
-                    } catch (e: Throwable) {
-                        Log.w("capture", "Capture initialization failed", e)
-                        default.set(LoggerState.ConfigurationFailure)
-                    }
-                } else -> {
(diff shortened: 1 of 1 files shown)


platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/Capture.kt
```diff
@@ -344,7 +386,14 @@ object Capture {
...
         @JvmStatic
         fun log(httpResponseInfo: HttpResponseInfo) {
-            default.get()?.log(httpResponseInfo)
+            logger()?.log(httpResponseInfo)
+        }
+
+        /**
+         * Used for testing purposes.
+         */
+        internal fun resetShared() {
```
murki, line 395: where do you use this?

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/Capture.kt
```diff
@@ -96,22 +147,28 @@ object Capture {
...
+                            apiUrl = apiUrl,
+                            fieldProviders = fieldProviders,
+                            dateProvider = dateProvider ?: SystemDateProvider(),
+                            configuration = configuration,
+                            sessionStrategy = sessionStrategy,
+                            bridge = bridge,
+                        )
+                        default.set(LoggerState.Configured(logger))
+                    } catch (e: Throwable) {
+                        Log.w("capture", "Capture initialization failed")
```
murki, line 165:
```suggestion
                        Log.w("capture", "Capture initialization failed", e)
```

murki: ^ we should attach the root cause so at least they have some info about what happened

murki (approved):
by always catching at construction time you might've "fixed" other cases where the `context.getSystemService` might be null (see BIT-3712). Should be ok although we now incur the cost of shutting down the entire logger when the better solution would be to only shutdown the dependent `ListenerLogger`

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/Capture.kt
murki: this is to distinguish other possible root causes that we're not anticipating (see my other comment)

PR #64, opened 2024-10-04
[andr][releases] Include sources.jar artifact in maven releases

murki:
Since the source is now available we should include the sources jar for ease of development and debugging on consumer apps.


ci/capture_android_release.sh
```diff
@@ -64,6 +64,7 @@ function release_capture_sdk() {
       "$sdk_repo/ci/NOTICE.txt" \
       "$name.pom" \
       "$name-javadoc.jar" \
+      "$name-sources.jar" \
```
murki, line 67: @Augustyniak I assume this is all I need to do since by looking at the logs it seems we do download and unzip the sources from gh here

PR #67, opened 2024-10-07
ci: fix release android script

Augustyniak:
Extra whitespaces added in https://github.com/bitdriftlabs/capture-sdk/pull/64 broke android release pipeline.

Adding a shell script to help to prevent cases like this in the future.


murki (approved):
🙃

PR #92, opened 2024-10-31
[ANDR][Tooling] Assorted improvements

murki:
- gradle cargo build broke after upgrading to sequoia (needs python below 3.13) see: https://github.com/mozilla/rust-android-gradle/issues/153
- central maven requires scm section in pom and artifact signing as pre-requisites: https://central.sonatype.org/publish/requirements/


platform/jvm/capture/build.gradle.kts
```diff
@@ -84,7 +84,7 @@ cargo {
     module = "../.."
     targetDirectory = "../../../target"
     targets = listOf("arm64", "x86_64")
-    pythonCommand = "python3"
+    pythonCommand = "/opt/homebrew/bin/python3.12"
```
murki, line 87: I actually I think this will probably will break in CI 🤔

PR #93, opened 2024-11-01
session replay: use new apis for screen capture, add support for screenshot capture

Augustyniak:
Fixes BIT-4005

1. Expose callbacks that are used by shared-core to inform the platform layer about a need to take a screenshot.
2. Remove session replay runtime variables from platform layer - the feature is controlled by the rRust layer now.
3. Breaking API change - remove an option to specify the rate at which session replay screens are captured. This is controlled with the use of runtime variables now. Another change is that customers can no longer disable session replay by passing `null` as `SessionReplayConfiguration`.


Reflejo (approved):
ios part lgtm

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/events/SessionReplayTarget.kt
```diff
@@ -27,37 +26,32 @@ import io.bitdrift.capture.replay.internal.EncodedScreenMetrics
...
-        when (event) {
-            Lifecycle.Event.ON_CREATE -> replayModule.create(context)
-            Lifecycle.Event.ON_START -> replayModule.start()
-            Lifecycle.Event.ON_STOP -> replayModule.stop()
-            else -> {}
-        }
+    override fun captureScreenshot() {
+        // This implementation is required for the SDK to support screenshot capture on Android.
+        // As currently implemented, the function must emit a session replay screenshot log.
+        // Without this emission, the SDK is blocked from requesting additional screenshots.
```
murki, line 54: missing a `//TODO`?

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/events/SessionReplayTarget.kt
```diff
@@ -27,37 +26,32 @@ import io.bitdrift.capture.replay.internal.EncodedScreenMetrics
...
+        context,
+        mainThreadHandler,
+    )
 
-    fun stop() {
-        mainThreadHandler.run {
-            processLifecycleOwner.lifecycle.removeObserver(this)
-        }
+    override fun captureScreen() {
+        val skipReplayComposeViews = runtime?.isEnabled(RuntimeFeature.SESSION_REPLAY_COMPOSE)
```
murki, line 46: why is runtime nullable? shouldn't it just be passed at construction time

platform/jvm/replay/src/main/kotlin/io/bitdrift/capture/replay/SessionReplayConfiguration.kt
```diff
@@ -9,10 +9,8 @@ package io.bitdrift.capture.replay
 
 /**
  * A configuration used to configure Bitdrift session replay feature.
- * @param captureIntervalMs The number of milliseconds between consecutive screen replay captures.
  * @param replayMapperConfiguration Map used to matching third party Android views to Bitdrift view types.
  */
 data class SessionReplayConfiguration @JvmOverloads constructor(
-    val captureIntervalMs: Long = 3000,
     val replayMapperConfiguration: ReplayMapperConfiguration? = null,
```
murki, line 15: 🤔  apparently this is the only part of the configuration left, which makes the entire nested structure unnecessary complicated. not much to do now but leaving this message for posterity: we should simplify this

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/events/SessionReplayTarget.kt
murki: That 2nd TODO seems more relevant now that none of the recurrent loggers are managed at the platform level (resource utilization and now replay). Hopefully we can get to that todo next.

murki (approved):

PR #99, opened 2024-11-05
Capture pixel screenshots implementation on iOS

Reflejo:


platform/swift/source/replay/SessionReplayTarget.swift
```diff
@@ -41,14 +42,36 @@ extension SessionReplayTarget: CapturePassable.SessionReplayTarget {
...
+        DispatchQueue.main.async {
+            guard let window = UIApplication.shared.sessionReplayWindows().first else {
+                // According to the documentation for
+                // `CapturePassable.SessionReplayTarget.captureScreenshot()`, this method must emit
+                // a screenshot log. Without this, the Rust logger will ignore any further screenshot
+                // capture actions.
+                // To handle cases where no window is available, an assertion is included here.
+                // TODO: Add a log to inform the Rust layer if the screenshot log emission fails, allowing it
+                // to proceed to the next screenshot (if any).
+                assertionFailure("no window to take screenshot of")
```
Reflejo, line 54: If that’s the case might be better to send an empty data object back for now, wdyt?

platform/swift/source/replay/SessionReplayTarget.swift
```diff
@@ -41,14 +42,32 @@ extension SessionReplayTarget: CapturePassable.SessionReplayTarget {
...
+                    screen: nil,
+                    duration: 0
+                )
+                return
+            }
+
+            let layer = window.layer
+            let bounds = UIScreen.main.bounds.size
+
+            self.queue.async { [weak self] in
```
murki, line 57: @Reflejo for my android-side understanding: would these two operations happen on separate threads? are any of these new worker threads or do we re-use existing ones?

murki: similarly, those these threads get spawned even when the runtime flag `session_replay.screenshots.enabled` is off?

Reflejo: Queues do not map 1:1 to threads, we actually use a default queue for the [target here](https://github.com/bitdriftlabs/capture-sdk/pull/99/files#diff-99302345e23541f201ed448d77a2f32870b5c5e13ea6aa09581181a838db19f8R13) meaning this won't create an additional thread.

The lines that are under the `DispatchQueue.main.async` but below `self.queue.async` will be dispatched to the main thread (basically just to access the layer object) then the `self.queue.async` closure runs that in our serial queue mentioned before

murki: I'm more confused than before? then what's the diff between queue and thread? what does default mean?

PR #109, opened 2024-11-07
android, session replay: fix session replay compose view parsing

Augustyniak:
The condition was flipped by accident in https://github.com/bitdriftlabs/capture-sdk/pull/93


murki (approved):

murki:
unit test?

PR #126, opened 2024-11-18
[andr][replay] Handle hybrid AndroidViews inside Compose

murki:
This took an embarrassingly long time to figure out, but after following the rabbit hole in https://github.com/androidx/androidx/blob/48f6421f6615572456ee94984209cd110a4694d4/compose/ui/ui/src/androidMain/kotlin/androidx/compose/ui/platform/accessibility/android_a11y_implementation_notes.md#androidview-children I realized that the Compose LayoutNode has an (internal) reference to the legacy AndroidView if it exists.

In the example below we're creating an hybrid view via: https://github.com/bitdriftlabs/capture-sdk/blob/f6edcc166a5f4bb1c9a5c6e1d2f6bcc1ddf89c1f/platform/jvm/gradle-test-app/src/main/java/io/bitdrift/gradletestapp/ComposeScreen.kt#L82-L91

|Screen|Before|After|
|---|---|---|
|Screenshot 2024-11-18 at 10 52 38 AM|Screenshot 2024-11-18 at 10 52 51 AM|Screenshot 2024-11-18 at 10 52 59 AM|

### DIFF
--- a/platform/jvm/gradle-test-app/src/androidTest/java/io/bitdrift/gradletestapp/ComposeReplayTest.kt
+++ b/platform/jvm/gradle-test-app/src/androidTest/java/io/bitdrift/gradletestapp/ComposeReplayTest.kt
@@ -8,6 +8,7 @@
 package io.bitdrift.gradletestapp
 
 import android.view.ViewGroup
+import android.widget.Button
 import android.widget.TextView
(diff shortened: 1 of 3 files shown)


snowp (approved):
So simple!

PR #128, opened 2024-11-19
Proposal for shutting down logger

murki:


platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/LoggerImpl.kt
```diff
@@ -265,6 +266,16 @@ internal class LoggerImpl(
         appExitSaveCurrentSessionId()
     }
 
+    override fun shutdown() {
+        runCatching {
+            appExitLogger.uninstallAppExitLogger()
+            eventsListenerTarget.stop()
+            // TODO(murki): Should we run this? in a bg thread?
+            CaptureJniLibrary.flush(this.loggerId, true)
+            CaptureJniLibrary.destroyLogger(this.loggerId)
```
murki, line 275: Need to confirm that calling `destroyLogger()` will destroy the threads used by `resource_utilization_target` and `session_replay_target` and `events_listener_target`

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/Capture.kt
```diff
@@ -206,6 +209,12 @@ object Capture {
             logger()?.startNewSession()
         }
 
+        @JvmStatic
+        fun shutdown() {
+            logger()?.shutdown()
```
snowp, line 214: I wonder if these need to be flipped, I think we'd crash if something tried to do something with the loggerid while shutdown is being called, so seems safer to pretend like we are shut down earlier?

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/LoggerImpl.kt
snowp: It will do so asynchronously, calling destroyLogger will trigger the shutdown notification to be sent to active tasks which should have them wind down as they are able to respond to the signal, then once all the primary tasks shut down we will turn down the runtime executor. 

All of those tasks seem to subscribe to the shutdown signal so I believe those tasks should wind down in response to calling shutdown

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/LoggerImpl.kt
```diff
@@ -265,6 +266,18 @@ internal class LoggerImpl(
...
     }
 
+    override fun shutdown() {
+        runCatching {
+            appExitLogger.uninstallAppExitLogger()
+            eventsListenerTarget.stop()
+            // TODO(murki): stop the rest of the threads (replay, resource utilization, etc.)
+            processingQueue.shutdown()
+            // TODO(murki): Should we run this? in a bg thread?
+            CaptureJniLibrary.flush(this.loggerId, true)
```
snowp, line 276: Running it blocking will potentially mean that it can take a while, maybe we just start with running it async and see how that goes?

snowp: Another option is to move this into the Rust layer so that it can be sequenced as part of shutting down, though it would mean having to rework the shutdown logic slightly

murki: yeah I think for now I'll remove explicitly calling it and later we can rework it into the shutdown sequence if needed

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/LoggerImpl.kt
```diff
@@ -265,6 +262,36 @@ internal class LoggerImpl(
...
+    override fun shutdown() {
+        runCatching {
+            Thread {
+                val shutdownOperations = listOf(
+                    { appExitSaveCurrentSessionId(this.sessionId) },
+                    { eventsListenerTarget.stop() },
+                    { captureExecutors.shutdown() },
+                    { apiClient.shutdown() },
+                    { captureNetwork.shutdown() },
+                    { CaptureJniLibrary.destroyLogger(this.loggerId) },
```
murki, line 274: 🔒 This PR should be locked until we can make this call to rust not fail and crash the app

PR #131, opened 2024-11-26
remove magic okhttp interceptors

snowp:


murki (approved):
i'm a little wary of potentially affecting customers' interceptors but we can test whether that's true

snowp:
This should only really affect the OkHttpClient that we create internally to the API client, so the main thing we are protecting against is extra interceptors added to *our* client that we have no guarantee will work correctly

PR #143, opened 2024-12-08
send device model in handshake and error reports

mattklein123:
Also bump shared core to bring in previous disconnection reason change in API.

Fixes BIT-4172


platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/LoggerImpl.kt
```diff
@@ -172,6 +173,7 @@ internal class LoggerImpl(
                 eventsListenerTarget,
                 clientAttributes.appId,
                 clientAttributes.appVersion,
+                Build.MODEL,
```
murki, line 176: This is already abstracted via `DeviceAttributes` we should expose it there

platform/swift/source/Logger.swift
```diff
@@ -128,6 +128,7 @@ public final class Logger {
         let appStateAttributes = AppStateAttributes()
         let clientAttributes = ClientAttributes()
         let deviceAttributes = DeviceAttributes()
+        let hardwareVersion = deviceAttributes.hardwareVersion
```
murki, line 131: is this allocation necessary? seems like we can call `model: deviceAttributes.hardwareVersion` below

mattklein123: I'm not very familiar with how swift works and what allocates or not. I will read more and clean it up.

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/LoggerImpl.kt
mattklein123: OK I will change this around.

murki (approved):
couple of nits, looks fine otherwise

PR #144, opened 2024-12-08
support two modes of okhttp instrumentation via plugin + wire up release pipeline

snowp:
Release pipeline updated to support multiple additional artifacts, in this case:
- capture-plugin
- capture-apollo3


ci/gh_prepare_android_artifacts.sh
```diff
@@ -43,33 +43,15 @@ function prepare_capture_sdk() {
...
-    files=(\
-      "$name.aar" \
-      "$name.module" \
-      "$name.pom" \
-      "$name-javadoc.jar" \
-      "$name-sources.jar" \
-      "$sdk_repo/ci/LICENSE.txt" \
-      "$sdk_repo/ci/NOTICE.txt" \
-    )
+  cp "$sdk_repo/capture-timber.zip" "$sdk_repo/capture-timber-$version.android.zip"
```
snowp, line 46: This has the implication that we will now include checksum files in the GH release zips for timber/plugin which isn't ideal but I don't think anyone relies on the GH releases so seems fine until we can simplify the deployment pipeline further

murki: that should be fine, arguably any artifact bundle can benefit from including checksums

.github/workflows/integrations_android.yaml
```diff
@@ -35,13 +35,23 @@ jobs:
...
         SKIP_PROTO_GEN: 1
-    - name: Compress artifacts
+    - name: Compress Timber artifacts
       run: |
         readonly dir=$(pwd)
         (cd capture-timber/build/repos/releases/io/bitdrift/capture-timber/${{ inputs.version }} && zip -r "$dir/capture-timber.zip" ./*)
+    - name: Compress Plugin artifacts
+      run: |
+        readonly dir=$(pwd)
+        (cd capture-plugin/build/repos/releases/io/bitdrift/capture-plugin/${{ inputs.version }} && zip -r "$dir/capture-plugin.zip" ./*)
```
murki, line 45: I'm pretty sure the upload artifacts job already compresses all the files in a folder, not sure why we had this compression step to being with

murki (approved):
lgtm, I'm gonna take over to merge it and build on top of it the similar changes required to releasing capture-apollo lib

.github/workflows/integrations_android.yaml
murki: also aren't you missing a step to run `run: ./gradlew :capture-plugin:publish` ??

ci/capture_android_release.sh
```diff
@@ -78,36 +79,28 @@ function release_capture_sdk() {
...
-    for file in "${files[@]}"; do
-      upload_file "$remote_location_prefix/$version" "$file"
-    done
-
-    generate_maven_file "$remote_location_prefix"
+    aws s3 cp "$version" "$remote_location_prefix/$version/" --recursive --region us-east-1
   popd
 }
 
-release_capture_sdk
```
murki, line 112: wrong method call removed

PR #154, opened 2024-12-10
[andr][ci] Make more fixes for the side libraries release

murki:
The .zip file is redundant since the upload/download actions automatically use zip for directories


jacksonhardaker (approved):
I trust you

PR #155, opened 2024-12-10
[andr][ci] Fix release jobs

murki:
Fixing issues introduced in #144
Tested it works in brach used here: https://github.com/bitdriftlabs/capture-sdk/actions/runs/12261648491

### DIFF
--- a/.github/workflows/release_gh.yaml
+++ b/.github/workflows/release_gh.yaml
@@ -203,19 +203,16 @@ jobs:
         uses: actions/download-artifact@v4
         with:
           name: capture-timber-${{ inputs.version }}.android.zip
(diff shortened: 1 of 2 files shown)


.github/workflows/release_gh.yaml
```diff
@@ -179,38 +179,48 @@ jobs:
         with:
           ref: ${{ inputs.sha }}
       # Download all artifacts to current working directory
-      - uses: actions/download-artifact@v4
```
Reflejo, line 182: are these downloading all artifacts? if so, wouldn't it be easier to just put it on a path and do
```      
    - name: Download all artifacts
        uses: actions/download-artifact@v4
        with:
          path: whatever
```

Reflejo (approved):

PR #158, opened 2024-12-11
allow enabling the ios network integratin through the ObjC interface

snowp:
This will enable us to call this from the RN plugin


Reflejo (approved):

snowp:
Found a workaround for the RN case - introduced a Swift indirection there instead of having to expose everything via the "official" objective-c version

PR #167, opened 2024-12-20
fix publishing of capture plugin

snowp:
The new plugin directory structure has a nested artifact within the artifact directory for the capture-plugin which messes up the script that tries to pull all the published releases. This should ignore that nested directory 

Fixes BIT-4242


murki (approved):
let's try it

PR #187, opened 2025-01-17
Redo networking swizzling

Reflejo:
This change moves from swizzling `URLSession` + `URLSessionTask` + using an `URLProtocol` to just swizzle one method in `URLSessionTask`. It's probably not going to cover all cases and only works on iOS 15.0 but at the time of writing 14.x market share is [about 0.09%](https://mixpanel.com/trends/#report/ios_14/from_date:-29,report_unit:day,to_date:0)

A nice advantage of this approach, additionally to increased safety is that it can be enabled at any point of the application lifecycle and will start capturing forward.


murki (approved):
we should a note to our docs about the version support

mattklein123 (approved):

PR #188, opened 2025-01-17
expose logCrash function that can be called before SDK init

snowp:
This PR introduces a way to report a crash event even before the SDK has started. No real event information is currently included in the event, but this should be sufficient to trigger an upload should another crash framework notify us that there has been a crash. 

By being able to be called before SDK init, this eliminates the ordering requirement between the initializing bitdrift before whatever crash handling may want to report a crash. 

Since this is limited to a simple crash notification, we can take a lot of shortcuts, making this not suitable for general logging prior to init. 

For both iOS and Android this uses a state machine guarded by synchronized access to make sure that we don't race against the SDK init. On Android this means that we are introducing a synchronized block during regular logger access where we previously could get away with atomics, though this just ends up matching the existing behavior on iOS. 

Fixes BIT-4335


platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/Capture.kt
```diff
@@ -92,6 +93,9 @@ object Capture {
         // This is a lazy property to avoid the need to initialize the main thread handler unless needed here
         private val mainThreadHandler by lazy { MainThreadHandler() }
 
+        @Volatile
+        private var crashReported = false
```
snowp, line 97: Having this tracked /w a separate volatile means that we're risking some race conditions but the only other option I see is to synchronize it all which introduces a mutex to the logging path

FranAguilera:

PR #173, opened 2025-01-08
[andr] Bump android deps to API level 35 (OS 15)

murki:
Below are tests conducted for Support 16 KB page sizes. Followed guide from: https://developer.android.com/guide/practices/page-sizes

Validated it's properly packaged and aligned after building with AGP >= 8.5.1 and NDK >= r27 (using linker flags `-Wl,-z,max-page-size=16384`)

## Gradle: gradle-test-app

before:
```
% ./toolchains/llvm/prebuilt/darwin-x86_64/bin/llvm-objdump -p ~/Desktop/tmp/16kb/gradle-test-app-debug/lib/arm64-v8a/libcapture.so | grep LOAD
    LOAD off    0x0000000000000000 vaddr 0x0000000000000000 paddr 0x0000000000000000 align 2**12
    LOAD off    0x0000000000503ea8 vaddr 0x0000000000504ea8 paddr 0x0000000000504ea8 align 2**12
    LOAD off    0x0000000000ffa0a0 vaddr 0x0000000000ffc0a0 paddr 0x0000000000ffc0a0 align 2**12
    LOAD off    0x00000000010acb78 vaddr 0x00000000010afb78 paddr 0x00000000010afb78 align 2**12
```

after:
```
% ./toolchains/llvm/prebuilt/darwin-x86_64/bin/llvm-objdump -p ~/Desktop/tmp/16kb/gradle-test-app-debug/lib/arm64-v8a/libcapture.so | grep LOAD
    LOAD off    0x0000000000000000 vaddr 0x0000000000000000 paddr 0x0000000000000000 align 2**14
    LOAD off    0x0000000000503ef0 vaddr 0x0000000000507ef0 paddr 0x0000000000507ef0 align 2**14
    LOAD off    0x0000000000ff9ff0 vaddr 0x0000000001001ff0 paddr 0x0000000001001ff0 align 2**14
    LOAD off    0x00000000010acaf8 vaddr 0x00000000010b8af8 paddr 0x00000000010b8af8 align 2**14
```

>Run the following [zipalign](https://developer.android.com/tools/zipalign) command to verify that your app is 16 KB-aligned
```
% ./build-tools/35.0.0/zipalign -c -P 16 -v 4 ~/Desktop/tmp/16kb/gradle-test-app-debug.apk
Verification succesful
```

Validated app runs and SDK works on session: https://timeline.bitdrift.dev/session/dd127004-4c49-42fe-b2dd-4506fc6c3a73

## Bazel: HelloWorld app

before:
```
[...]

PR #193, opened 2025-01-23
[ANDR] Adding linker flags to only android platform

FranAguilera:
# Issue
Potential fix for building issues on previous 0.16.9 attempt (see previous failure on https://github.com/bitdriftlabs/capture-sdk/actions/runs/12916680190/job/36022967272)

# Test plan

Will follow similar steps listed on prior PR's description https://github.com/bitdriftlabs/capture-sdk/pull/173 

- [ ] Validate locally bazel android builds
- [ ] Verify that the Android app is 16 KB-aligned
- [ ] Validate app runs and the SDK is reporting last session  
Check that Android test app is https://developer.android.com/guide/practices/page-sizes

### DIFF
--- a/platform/jvm/BUILD.bazel
+++ b/platform/jvm/BUILD.bazel
@@ -15,7 +15,10 @@ cc_binary(
         "-lm",  # Required to avoid dlopen runtime failures unrelated to rust
         "-lz",  # Link against system zlib library
     ] + select({
-        "@platforms//os:android": ["-Wl,-z,max-page-size=16384,--retain-symbols-file=$(location :jni_symbols.lds)"],
+        "@platforms//os:android": [
+            "-Wl,-z,max-page-size=16384", # nable 16 KB ELF alignment on Android to support API 35+
+            "-Wl,--retain-symbols-file=$(location :jni_symbols.lds)",
+            ],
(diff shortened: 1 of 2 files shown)


platform/jvm/BUILD.bazel
```diff
@@ -14,9 +14,8 @@ cc_binary(
     linkopts = [
         "-lm",  # Required to avoid dlopen runtime failures unrelated to rust
         "-lz",  # Link against system zlib library
-        "-Wl,-z,max-page-size=16384",  # enable 16 KB ELF alignment
     ] + select({
-        "@platforms//os:android": ["-Wl,--retain-symbols-file=$(location :jni_symbols.lds)"],
+        "@platforms//os:android": ["-Wl,-z,max-page-size=16384,--retain-symbols-file=$(location :jni_symbols.lds)"],
```
murki, line 18: this should work, the best way to test this still works would be not to test before/after the pr chances but with and without this flag

murki (approved):

murki:
also for testing this shouldn't affect the gradle test app (that flag is set somewhere else) but only bazel

murki (approved):
I tested this work locally on my 16-kb enabled emulator

PR #205, opened 2025-02-03
[Draft] [Android] Add BackgroundExecutor

FranAguilera:
Explorations for not blocking caller thread doing initial setup of sdk via Capture.Logger.start() 

Will link doc with more context soon

### DIFF
--- a/platform/jvm/common/src/main/kotlin/io/bitdrift/capture/common/BackgroundExecutor.kt
+++ b/platform/jvm/common/src/main/kotlin/io/bitdrift/capture/common/BackgroundExecutor.kt
@@ -1,0 +1,25 @@
+package io.bitdrift.capture.common
+
+import java.util.concurrent.ExecutorService
+import java.util.concurrent.Executors
+
+/**
+ * A single threaded [ExecutorService] that can be used to run an specific block of code
+ */
+object BackgroundExecutor {
+
+    private const val SINGLE_THREAD_EXECUTOR_NAME = "io.bitdrift.capture.single-background-executor"
+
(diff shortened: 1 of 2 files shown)


platform/jvm/common/src/main/kotlin/io/bitdrift/capture/common/BackgroundExecutor.kt
```diff
@@ -0,0 +1,25 @@
+package io.bitdrift.capture.common
+
+import java.util.concurrent.ExecutorService
+import java.util.concurrent.Executors
+
+/**
+ * A single threaded [ExecutorService] that can be used to run an specific block of code
+ */
+object BackgroundExecutor {
```
murki, line 9: I like the idea of having executors definition on a separate place, I think this could be a good opportunity to organize them all, maybe something similar to what I was playing with here https://github.com/bitdriftlabs/capture-sdk/pull/128/files#diff-118bdebccae8c1bde50314ef9d78f897d272e1f222d57a5d8a44d4f0f348b5d4R13

murki: another example `https://github.com/measure-sh/measure/blob/0721674bbba592e32b3b66d2e5a390e6c594cf89/android/measure/src/main/java/sh/measure/android/executors/ExecutorServiceRegistry.kt`

FranAguilera: totally agree, will update

PR #207, opened 2025-02-03
[Android] Consolidate existing ExecutorServices

FranAguilera:
# What

This is a follow up from @murki on https://github.com/bitdriftlabs/capture-sdk/pull/128/files#diff-118bdebccae8c1bde50314ef9d78f897d272e1f222d57a5d8a44d4f0f348b5d4R13 that proposes to consolidates all existing ExecutorService

This will be useful later on for https://github.com/bitdriftlabs/capture-sdk/pull/205 

# Verification

Verify sessions are captured in same manner and existing thread names remain the same:

- io.bitdrift.capture.event-listener
- io.bitdrift.capture.network.okhttp
- io.bitdrift.capture.session-replay

### DIFF
--- a/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/LoggerImpl.kt
+++ b/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/LoggerImpl.kt
@@ -45,7 +45,7 @@ import io.bitdrift.capture.providers.FieldValue
 import io.bitdrift.capture.providers.MetadataProvider
 import io.bitdrift.capture.providers.session.SessionStrategy
 import io.bitdrift.capture.providers.toFields
-import io.bitdrift.capture.threading.CaptureDispatcher
+import io.bitdrift.capture.threading.CaptureDispatchers
 import okhttp3.HttpUrl
 import java.io.File
 import kotlin.time.Duration
@@ -76,7 +76,7 @@ internal class LoggerImpl(
     private var deviceCodeService: DeviceCodeService = DeviceCodeService(apiClient),
     private val activityManager: ActivityManager = context.getSystemService(Context.ACTIVITY_SERVICE) as ActivityManager,
     private val bridge: IBridge = CaptureJniLibrary,
-    private val eventListenerDispatcher: CaptureDispatcher = CaptureDispatcher.EventListener
+    private val eventListenerDispatcher: CaptureDispatchers.EventListener = CaptureDispatchers.EventListener
 ) : ILogger {
     private val metadataProvider: MetadataProvider
     private val memoryMonitor = MemoryMonitor(context)
--- a/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/LoggerImpl.kt
+++ b/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/LoggerImpl.kt
@@ -76,6 +76,7 @@ internal class LoggerImpl(
     private var deviceCodeService: DeviceCodeService = DeviceCodeService(apiClient),
     private val activityManager: ActivityManager = context.getSystemService(Context.ACTIVITY_SERVICE) as ActivityManager,
     private val bridge: IBridge = CaptureJniLibrary,
+    private val eventListenerDispatcher: CaptureDispatcher = CaptureDispatcher.EventListener
 ) : ILogger {
     private val metadataProvider: MetadataProvider
     private val memoryMonitor = MemoryMonitor(context)
@@ -93,8 +94,6 @@ internal class LoggerImpl(
     private val eventsListenerTarget = EventsListenerTarget()
 
     private val sessionReplayTarget: SessionReplayTarget?
-
-    private val processingQueue = CaptureDispatcher.EventListener.executorService
 
     @VisibleForTesting
     internal val loggerId: LoggerId
@@ -158,7 +157,7 @@ internal class LoggerImpl(
                         diskUsageMonitor,
                         errorHandler,
                         this,
-                        processingQueue,
+                        eventListenerDispatcher.executorService,
                     )
 
                 val sessionReplayTarget =
@@ -207,7 +206,7 @@ internal class LoggerImpl(
                         this,
                         ProcessLifecycleOwner.get(),
                         runtime,
-                        processingQueue,
+                        eventListenerDispatcher.executorService,
                     ),
                 )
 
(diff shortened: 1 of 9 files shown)


platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/LoggerImpl.kt
```diff
@@ -65,10 +64,6 @@ internal class LoggerImpl(
     fieldProviders: List<FieldProvider>,
     dateProvider: DateProvider,
     private val errorHandler: ErrorHandler = ErrorHandler(),
-    processingQueue: ExecutorService =
```
murki, line 68: we should still pass it via the constructor so it can be mocked for tests if necessary

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/network/okhttp/OkHttpNetwork.kt
```diff
@@ -100,10 +99,8 @@ internal class OkHttpNetwork(
                 .build()
         }
 
-    private val executor: ExecutorService =
-        Executors.newSingleThreadExecutor {
-            Thread(it, "io.bitdrift.capture.network.okhttp")
-        }
+    private val networkDispatcher: CaptureDispatcher.Network = CaptureDispatcher.Network
```
murki, line 102: here too we should use this opportunity to pass it in the ctor

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/threading/CaptureDispatcher.kt
```diff
@@ -0,0 +1,51 @@
...
+
+    /**
+     * [ExecutorService] to be used for Session-Replay
+     */
+    object SessionReplay: CaptureDispatcher("session-replay")
+
+    /**
+     * Tears down the existing service
+     */
+    fun shutDown() {
```
murki, line 38:
```suggestion
    fun shutdown() {
```

murki: ^ nit to make it consistency with naming below

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/threading/CaptureDispatcher.kt
```diff
@@ -0,0 +1,51 @@
...
+        executorService.shutdown()
```
murki, line 37: how does this work? IS this supposed to shutdown all initialized executor services? i'm not sure i understand how

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/LoggerImpl.kt
FranAguilera: good callout, will update

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/network/okhttp/OkHttpNetwork.kt
FranAguilera: agree, will update similar as above

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/threading/CaptureDispatcher.kt
FranAguilera: given we are going to specify the type will can manually call this later on. I will remove it for now to avoid noise and added back when we do the proper shutdown as in your prior proposal PR

murki: the intent of that other PR is to have a global shutdown for all initialized executors, but i agree that that's probably out of scope for this PR if you believe the current design makes it easy to add it later on (i'm not convinced it does)

FranAguilera: totally agree, I was thinking to keep track of a register via some sort of register function that is apply each time a new type is created. I can add in this diff to clarify the shutdownAll

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/threading/CaptureDispatcher.kt
```diff
@@ -0,0 +1,63 @@
...
+ *
+ * The initial implementation exposes an [executorService]
+ */
+sealed class CaptureDispatcher private constructor(private val threadName: String) {
+
+    /**
+     * Returns the associated [ExecutorService] for the [CaptureDispatcher] type
+     */
+    val executorService: ExecutorService = buildExecutorService(threadName)
+            .also { register(this) }
```
FranAguilera, line 17: @murki Register happening here now

murki (approved):
lgtm thanks for doing this sorry for the scope creep

FranAguilera:
> lgtm thanks for doing this sorry for the scope creep

thanks! np at all

murki:
@FranAguilera For the failing test my guess is you need to inject something like `Executors.newSingleThreadExecutor()` to the classes used in the tests

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/threading/CaptureDispatchers.kt
```diff
@@ -0,0 +1,68 @@
...
+     */
+    companion object {
+        private const val CAPTURE_EXECUTOR_SERVICE_NAME = "io.bitdrift.capture"
+
+        private val initializedDispatchers = mutableSetOf<CaptureDispatchers>()
+
+        /**
+         * Shuts down all used dispatchers
+         */
+        fun shutdownAll() {
```
murki, line 59: I would bind this to a button in the test-app ui to test it works properly (example https://github.com/bitdriftlabs/capture-sdk/pull/128/files#diff-7c8515dbaf1f92c15ce3488493f51b09d0c1252d7ba4e8d92b57f8e0ac8520b1R131)

FranAguilera: forgot I had auto merge, but will create a followup small PR for this soon

PR #208, opened 2025-02-05
Add platform server/cpp dumb lib

Reflejo:


platform/server/cpp/src/logger.rs
```diff
@@ -0,0 +1,355 @@
...
+      network: Box::new(handle),
+      static_metadata,
+    })
+    .with_client_stats(true)
+    .build()?;
+
+    LoggerBuilder::run_logger_runtime(async {
+      // Make sure we hold onto the shutdown handle to avoid an immediate shutdown.
+      #[allow(clippy::no_effect_underscore_binding)]
+      let _shutdown = shutdown;
```
mattklein123, line 119: nit: to avoid the clippy lint usually I just `drop(shutdown)` at the end of the block.

platform/server/cpp/src/logger.rs
```diff
@@ -0,0 +1,355 @@
...
+  }
+
+  fn path(&self, key: &str) -> PathBuf {
+    let mut hasher = DefaultHasher::new();
+    key.hash(&mut hasher);
+    self.root.join(hasher.finish().to_string())
+  }
+}
+
+impl Storage for DiskStorage {
```
mattklein123, line 302: This is not your doing but this should be an async trait and use async file APIs. I would just add a TODO here. I can look into this when I get back into the SDK code next.

mattklein123 (approved):

FranAguilera:

PR #209, opened 2025-02-05
[Android] [Enrich perf data] Add memory attributes into App Exit and UncaughtExceptionHandler

FranAguilera:
# What

Adding memory attributes into `AppExit` and `UncaughtExceptionHandler`

# Verification

https://timeline.bitdrift.dev/session/6ff292fe-7237-4a7f-92aa-c6e0fb1e96a1?pages=1&utilization=0&expanded=-355182893748571615

## AppExit - Before (Without additional memory fields)
app_exit_before

## AppExit  - After (With memory fields)
app_exit_with_memory_class

## UncaughtExceptionHandler - Before (Without additional memory fields)
crash_without_memory

## UncaughtExceptionHandler  - After (With memory fields)
crash_with_memory

### DIFF
--- a/platform/jvm/capture/src/test/kotlin/io/bitdrift/capture/AppExitLoggerTest.kt
+++ b/platform/jvm/capture/src/test/kotlin/io/bitdrift/capture/AppExitLoggerTest.kt
@@ -154,18 +154,14 @@ class AppExitLoggerTest {
         // ASSERT
         val expectedFields =
             buildMap {
-                putAll(
-                    mapOf(
-                        "_app_exit_source" to "ApplicationExitInfo",
-                        "_app_exit_process_name" to "test-process-name",
-                        "_app_exit_reason" to "ANR",
-                        "_app_exit_importance" to "FOREGROUND",
-                        "_app_exit_status" to "0",
-                        "_app_exit_pss" to "1",
-                        "_app_exit_rss" to "2",
-                        "_app_exit_description" to "test-description",
-                    ),
-                )
+                put("_app_exit_source", "ApplicationExitInfo")
+                put("_app_exit_process_name", "test-process-name")
+                put("_app_exit_reason", "ANR")
+                put("_app_exit_importance", "FOREGROUND")
+                put("_app_exit_status", "0")
+                put("_app_exit_pss", "1")
+                put("_app_exit_rss", "2")
+                put("_app_exit_description", "test-description")
                 putAll(DEFAULT_MEMORY_ATTRIBUTES_MAP)
             }.toFields()
         verify(logger).log(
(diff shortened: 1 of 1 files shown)


platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/events/lifecycle/AppExitLogger.kt
```diff
@@ -141,20 +148,46 @@ internal class AppExitLogger(
...
     }
 
+    private fun buildCrashAndMemoryFieldsMap(
+        thread: Thread,
+        throwable: Throwable,
+    ): InternalFieldsMap {
+        val rootCause = throwable.getRootCause()
+        return buildMap {
+            putAll(
+                mapOf(
```
murki, line 158: can avoid this extra allocation by just individual puts inside buildMap { }

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/events/performance/MemoryMetricsProvider.kt
```diff
@@ -0,0 +1,18 @@
...
+// Use of this source code is governed by a source available license that can be found in the
+// LICENSE file or at:
+// https://polyformproject.org/wp-content/uploads/2020/06/PolyForm-Shield-1.0.0.txt
+
+package io.bitdrift.capture.events.performance
+
+/**
+ * Provides Memory related attributes such as Memory class, Total JVM memory, used JVM memory, etc
+ */
+fun interface MemoryMetricsProvider {
```
murki, line 13: /nit: I know we don't always follow conventions everywhere but the latest agreement was to prefix interfaces with I: `IMemoryMetricsProvider` and then rename the impl to just `MemoryMetricsProvider` (from `MemoryMonitor` in this case)

FranAguilera: got it! will update

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/events/lifecycle/AppExitLogger.kt
FranAguilera: good point, will update

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/events/performance/MemoryMonitor.kt
```diff
@@ -15,10 +15,10 @@ private const val KB = 1024L
 
 internal class MemoryMonitor(
     context: Context,
-) {
+) : MemoryMetricsProvider {
```
murki, line 18: comment for below: in retrospective I think that computing the memory class value as `activityManager.memoryClass * 1024` was a mistake, since it deviates from that the android sdk gives and what developers expect. I think we should bite the bullet and just change it to be reported in megabytes

FranAguilera:
> lgtm -after the change of moving to report memory class in mb

Thanks, discussed offline will be taking this as a follow up task in BIT-4465 given this will require backend changes

PR #213, opened 2025-02-06
[Android] Add thread count metadata to AppExit and UncaughtExceptionHandler

FranAguilera:
For now adding current thread count to AppExit and UncaughtExceptionHandler.

For OOM crashes due to Too many threads this will be useful information for Crash reporting (this thread count can be comparable to the number of threads at initial cold launch via AppExit report)

# Verification

https://timeline.bitdrift.dev/session/a39246be-9910-4348-8dfe-23bac740ffdb?pages=1&utilization=0&expanded=2918627673516284925

Screenshot 2025-02-06 at 7 13 50 PM

Screenshot 2025-02-06 at 7 16 29 PM

### DIFF
--- a/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/events/performance/IThreadMetricsProvider.kt
+++ b/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/events/performance/IThreadMetricsProvider.kt
@@ -1,0 +1,18 @@
+// capture-sdk - bitdrift's client SDK
+// Copyright Bitdrift, Inc. All rights reserved.
+//
+// Use of this source code is governed by a source available license that can be found in the
+// LICENSE file or at:
+// https://polyformproject.org/wp-content/uploads/2020/06/PolyForm-Shield-1.0.0.txt
+
+package io.bitdrift.capture.events.performance
+
+/**
+ * Provides attributes for threads running in the app
+ */
+interface IThreadMetricsProvider {
+    /**
+     * Returns Thread information (e.g. Thread count)
+     */
+    fun getThreadAttributes(): Map<String, String>
+}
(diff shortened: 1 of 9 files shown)


platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/events/performance/ThreadMetricsProvider.kt
```diff
@@ -0,0 +1,21 @@
...
+
+package io.bitdrift.capture.events.performance
+
+/**
+ * Concrete implementation of [IMemoryMetricsProvider]
+ */
+internal class ThreadMetricsProvider : IThreadMetricsProvider {
+    override fun getThreadAttributes(): Map<String, String> = mapOf(THREAD_COUNT_KEY to getCurrentThreadCount())
+
+    private fun getCurrentThreadCount(): String = Thread.getAllStackTraces().size.toString()
```
murki, line 16: shouldn't the call be `Thread.activeCount()` ? or what's the difference

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/events/performance/IThreadMetricsProvider.kt
```diff
@@ -0,0 +1,18 @@
...
+package io.bitdrift.capture.events.performance
+
+/**
+ * Provides attributes for threads running in the app
+ */
+interface IThreadMetricsProvider {
+    /**
+     * Returns Thread information (e.g. Thread count)
+     */
+    fun getThreadAttributes(): Map<String, String>
```
murki, line 17: As BIT-371 say could you add this to resource utilization logs as well? without it this information can only be available for users that have crashed vs every single user

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/events/performance/ThreadMetricsProvider.kt
murki: also is this safe to be called from the main thread?

FranAguilera:
Will reopen when this is being prioritized again

PR #215, opened 2025-02-07
[DO NOT REVIEW] Adding Performance library dep

FranAguilera:
TBF


FranAguilera:
Will explore this again after jank stats

PR #217, opened 2025-02-10
Ensure sdk directory has no file protections

Reflejo:
Without this, apps that set the protection entitlement of
`NSURLFileProtectionNone` are likely to crash when the device is locked
since the ring buffer mmap will keep trying to write to disk.


platform/swift/source/LoggerBridge.swift
```diff
@@ -30,6 +55,14 @@ final class LoggerBridge: LoggerBridging {
         network: Network?,
         errorReporting: RemoteErrorReporting
     ) {
+        do {
+            try bufferDirectoryPath.map(makeDirectoryAndDisableProtection(at:))
+        } catch {
+            // To be safe we don't initialize the logger if we can't create the directory or set
+            // the file protection policy.
+            return nil
```
mattklein123, line 63: Should we warn here or something so it's obvious at least in local dev if this fails for some reason?

platform/swift/source/LoggerBridge.swift
```diff
@@ -30,6 +55,14 @@ final class LoggerBridge: LoggerBridging {
...
+        }
+
         let loggerID = capture_create_logger(
```
mattklein123, line 66: I don't think so but just checking that the platform code doesn't make any files before we get here? I know there is https://github.com/bitdriftlabs/capture-sdk/blob/main/platform/swift/source/src/key_value_storage.rs but this is used by the rust code only AFAIK to store things like device ID, etc. On that note, does whatever that use need any permission changes or it's a separate thing?

mattklein123 (approved):
Nice. From looking at capture-es, it looks like that code uses this code? Or do we need changes over there as well?

platform/swift/source/LoggerBridge.swift
Reflejo: I'll follow up on a diff PR, the reason I didn't do that is we don't have a nice way of logging stuff right now and there are more cases where we return `nil` silently

platform/swift/source/LoggerBridge.swift
Reflejo: separate thing. I think we should be good with userdefaults. will double check though

mattklein123 (approved):

PR #218, opened 2025-02-10
[Android] Move expensive AppExitLogger methods off the main thread

FranAguilera:
## What

Following up on https://linear.app/bitdrift/issue/BIT-4410 proposing to make expensive AppExit methods a non-blocking call. This is given the potential impact on main thread as Capture.Logger.start() is going to be usually called on Application/Main activity onCreate

Local measurements shows that building and initializing AppExit is the main contributor to Capture.Logger.start() time.

# Verification

- Verified that crash/app exit terminations are reported successfully
- Verified that the session id is correct
- Validate improvements on main thread of Capture.Logger.start() call on low end device
- Modify existing microbenchmarks to measure with installAppExitLogger on main vs background

**NOTE: Please note that the whole method doesn't get executed given that there are no exit reasons so we early exit without appending app exit data and memory attributes

On current main
image

On this diff
image

Device used
image

### DIFF
--- a/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/LoggerImpl.kt
+++ b/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/LoggerImpl.kt
@@ -76,7 +76,7 @@ internal class LoggerImpl(
     private var deviceCodeService: DeviceCodeService = DeviceCodeService(apiClient),
     private val activityManager: ActivityManager = context.getSystemService(Context.ACTIVITY_SERVICE) as ActivityManager,
     private val bridge: IBridge = CaptureJniLibrary,
-    private val eventListenerDispatcher: CaptureDispatchers.EventListener = CaptureDispatchers.EventListener,
+    private val eventListenerDispatcher: CaptureDispatchers.CommonBackground = CaptureDispatchers.CommonBackground,
 ) : ILogger {
     private val metadataProvider: MetadataProvider
     private val memoryMetricsProvider = MemoryMetricsProvider(context)
--- a/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/events/lifecycle/AppExitLogger.kt
+++ b/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/events/lifecycle/AppExitLogger.kt
@@ -20,6 +20,7 @@ import io.bitdrift.capture.LogLevel
 import io.bitdrift.capture.LogType
 import io.bitdrift.capture.LoggerImpl
 import io.bitdrift.capture.common.ErrorHandler
+import io.bitdrift.capture.common.IBackgroundThreadHandler
 import io.bitdrift.capture.common.Runtime
 import io.bitdrift.capture.common.RuntimeFeature
 import io.bitdrift.capture.events.performance.IMemoryMetricsProvider
@@ -37,7 +38,7 @@ internal class AppExitLogger(
     private val crashHandler: CaptureUncaughtExceptionHandler = CaptureUncaughtExceptionHandler(),
     private val versionChecker: BuildVersionChecker = BuildVersionChecker(),
     private val memoryMetricsProvider: IMemoryMetricsProvider,
-    private val commonBackgroundDispatcher: CaptureDispatchers.CommonBackground = CaptureDispatchers.CommonBackground,
+    private val backgroundThreadHandler: IBackgroundThreadHandler = CaptureDispatchers.CommonBackground,
 ) {
     companion object {
         const val APP_EXIT_EVENT_NAME = "AppExit"
@@ -59,7 +60,8 @@ internal class AppExitLogger(
         if (!runtime.isEnabled(RuntimeFeature.APP_EXIT_EVENTS)) {
             return
         }
-        commonBackgroundDispatcher.launchAsync {
+//        checkIsNotOnMainThread()
+        backgroundThreadHandler.runAsync {
             crashHandler.install(this)
             saveCurrentSessionId()
             logPreviousExitReasonIfAny()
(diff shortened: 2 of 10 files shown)


platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/threading/CaptureDispatchers.kt
```diff
@@ -17,17 +17,43 @@ import java.util.concurrent.Executors
...
+    }
 
     /**
      * Returns the associated [ExecutorService] for the [CaptureDispatchers] type
      */
-    val executorService: ExecutorService
-        get() = _executorService
+    fun asExecutorService(): ExecutorService {
+        if (executorService.isShutdown || executorService.isTerminated) {
+            executorService = buildExecutorService(threadName)
```
FranAguilera, line 53: @murki The idea here is to automatically handle restart of executorService in case there was a prior call to shutdownAll

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/Capture.kt
```diff
@@ -153,26 +157,28 @@ object Capture {
...
-                            dateProvider = dateProvider ?: SystemDateProvider(),
-                            configuration = configuration,
-                            sessionStrategy = sessionStrategy,
-                            bridge = bridge,
-                        )
-                    default.set(LoggerState.Started(logger))
-                } catch (e: Throwable) {
-                    Log.w("capture", "Failed to start Capture", e)
-                    default.set(LoggerState.StartFailure)
+            sdkInitializationDispatcher.launchAsyncAndShutDown {
```
murki, line 160: It seems wasteful to create and teardown a whole executor service just for one operation. Could we instead re-use one of the existing ones? say `event-listener` for example (and feel free to rename).

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/Capture.kt
```diff
@@ -153,26 +157,28 @@ object Capture {
...
+                // Ideally we would use `getAndUpdate` in here but it's available for API 24 and up only.
+                if (default.compareAndSet(LoggerState.NotStarted, LoggerState.Starting)) {
```
murki, line 162: don't we run the risk to race conditions here (multiple loggers being initialized if the start() method is called in rapid succession)? I'd think that starting the background operation _after_ this check would avoid that

murki:
@FranAguilera can we also make sure we test when an actual appExitInfo is retrieved (if you haven't)? want to make sure there are no issues in calling `activityManager.setProcessStateSummary()` and `activityManager.getHistoricalProcessExitReasons()` from a bg thread

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/Capture.kt
FranAguilera: good idea, I wasn't very happy with this approach as well. Will rename the existing event-listener to handle this as well

FranAguilera:
> @FranAguilera can we also make sure we test when an actual appExitInfo is retrieved (if you haven't)? want to make sure there are no issues in calling `activityManager.setProcessStateSummary()` and `activityManager.getHistoricalProcessExitReasons()` from a bg thread

Good point, I've verified on this [session](https://timeline.bitdrift.dev/session/6a4f79ac-6271-4791-80aa-6a318c570147?pages=1&utilization=0&expanded=-2263628960630733080) . It doesn't seem to be an issue by calling activityManager.getHistoricalProcessExitReasons() off main thread

FranAguilera:
Will do further changes to support testing async the initialization

FranAguilera:
Will fix tests on AppExitLogger before publishing again

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/LoggerImpl.kt
```diff
@@ -241,19 +242,9 @@ internal class LoggerImpl(
...
-                        errorHandler,
-                        memoryMetricsProvider = memoryMetricsProvider,
-                    )
-
-                // Install the app exit logger before the Capture logger is started to ensure
-                // that logs emitted during the installation are the first logs emitted by the
-                // Capture logger.
-                appExitLogger.installAppExitLogger()
+                commonBackgroundDispatcher.launchAsync {
+                    setupAppExitInfo(commonBackgroundDispatcher)
```
murki, line 246: I would just move the `appExitLogger.installAppExitLogger()` to the bg, there's not benefit of moving the construction plus it creates potential race conditions of it being null

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/events/lifecycle/AppExitLogger.kt
```diff
@@ -114,30 +144,34 @@ internal class AppExitLogger(
...
-        // explicitly letting it run in the caller thread
-        logger.log(
-            LogType.LIFECYCLE,
-            LogLevel.ERROR,
-            buildCrashAndMemoryFieldsMap(thread, throwable),
-            blocking = true, // this ensures we block until the log has been persisted to disk
-        ) {
-            APP_EXIT_EVENT_NAME
-        }
+        runOnBackgroundThread {
```
murki, line 152: we should not make this call async... as you see below we want it to be blocking (`// this ensures we block until the log has been persisted to disk`), this is important because we will attempt to upload this log on next app launch so we should try to guarantee it will be indeed persisted

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/events/lifecycle/AppExitLogger.kt
```diff
@@ -60,29 +76,40 @@ internal class AppExitLogger(
...
         crashHandler.uninstall()
     }
 
     @TargetApi(Build.VERSION_CODES.R)
+    @WorkerThread
     fun saveCurrentSessionId(sessionId: String? = null) {
-        if (!runtime.isEnabled(RuntimeFeature.APP_EXIT_EVENTS) || !versionChecker.isAtLeast(Build.VERSION_CODES.R)) {
-            return
-        }
+        runOnBackgroundThread {
```
murki, line 88: I think it's overkill to move this operation to the bg, moreso this could also lead to some unpredictable race conditions (the sessionId might be be persisted on time to be read after)

murki (approved):
i won't block on this approach since i think it's reasonable, but the more i think about i think we would get the most bang for our buck on reducing the blast radius of this change to just move `logPreviousExitReasonIfAny()` to the bg. up to you

FranAguilera:
Pending to add results from microbenchmarks with/without the change

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/events/lifecycle/AppExitLogger.kt
```diff
@@ -51,13 +54,16 @@ internal class AppExitLogger(
...
+    @WorkerThread
     fun installAppExitLogger() {
         if (!runtime.isEnabled(RuntimeFeature.APP_EXIT_EVENTS)) {
             return
         }
-        crashHandler.install(this)
-        saveCurrentSessionId()
-        logPreviousExitReasonIfAny()
+        commonBackgroundDispatcher.launchAsync {
+            crashHandler.install(this)
```
murki, line 63: This class method internally calls `Thread.setDefaultUncaughtExceptionHandler(this)` -I'm not sure this static call would attach the handler to the _current_ thread (doubtful?) or just globally regardless of the thread. I would verify it just in case

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/events/lifecycle/AppExitLogger.kt
```diff
@@ -82,36 +88,39 @@ internal class AppExitLogger(
...
-        val exits =
-            try {
-                // a null packageName means match all packages belonging to the caller's process (UID)
-                // pid should be 0, a value of 0 means to ignore this parameter and return all matching records
-                // maxNum should be 1, The maximum number of results to be returned, as we need only the last one
-                activityManager.getHistoricalProcessExitReasons(null, 0, 1)
-            } catch (error: Throwable) {
-                errorHandler.handleError("Failed to retrieve ProcessExitReasons from ActivityManager", error)
-                emptyList()
+        commonBackgroundDispatcher.launchAsync {
```
murki, line 97: `logPreviousExitReasonIfAny()` is only being called in one place, by `installAppExitLogger()` above. what's the implication of running nested `commonBackgroundDispatcher.launchAsync {` operations? is that necessary?

platform/jvm/capture/src/test/kotlin/io/bitdrift/capture/AppExitLoggerTest.kt
```diff
@@ -38,25 +40,28 @@ class AppExitLoggerTest {
...
-            errorHandler,
-            crashHandler,
-            versionChecker,
-            memoryMetricsProvider,
-        )
+    private lateinit var appExitLogger: AppExitLogger
 
     @Before
     fun setUp() {
+        CaptureDispatchers.setTestExecutorService(MoreExecutors.newDirectExecutorService())
```
murki, line 52: this seems to be a roundabout way to inject a test injector? can't you just pass `MoreExecutors.newDirectExecutorService()` as the last param of `AppExitLogger()` ctor? that's how we do it in all of our other tests

platform/jvm/common/src/main/kotlin/io/bitdrift/capture/common/MainThreadHandler.kt
```diff
@@ -25,4 +25,27 @@ class MainThreadHandler {
     fun run(run: () -> Unit) {
         mainHandler.post { run() }
     }
+
+    /**
+     * Common main thread checkers
+     */
+    companion object {
```
murki, line 32: is this used anywhere?

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/events/lifecycle/AppExitLogger.kt
FranAguilera: Good catch, I've verified and there is no issue but agree this is unnecessary and will revert

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/events/lifecycle/AppExitLogger.kt
FranAguilera: Just verified it doesn't seem to be an issue here

platform/jvm/capture/src/test/kotlin/io/bitdrift/capture/fakes/FakeBackgroundThreadHandler.kt
```diff
@@ -0,0 +1,22 @@
...
+package io.bitdrift.capture.fakes
+
+import com.google.common.util.concurrent.MoreExecutors
+import io.bitdrift.capture.common.IBackgroundThreadHandler
+
+/**
+ * Fake [IBackgroundThreadHandler] that relies on newDirectExecutorService
+ */
+class FakeBackgroundThreadHandler : IBackgroundThreadHandler {
+    private val fakeExecutorService = MoreExecutors.newDirectExecutorService()
```
murki, line 17: 👍

murki (approved):
🚀

PR #220, opened 2025-02-11
[Android] Provide listener for Capture.Logger.start

FranAguilera:
## What

Provides an optional listener that reports Logger status upon a call to Capture start (in the future shutdown, etc)

## Verification

Tested on sample app. Now toaster and logcat will be shown if API key is invalid

https://github.com/user-attachments/assets/4fbac99d-46b4-466e-b454-2bd7210893b6

https://github.com/user-attachments/assets/1dfbdcad-a76e-4a04-8ef5-aa8d92d469a4

NOTE: As a followup will update related sdk docs


platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/Capture.kt
```diff
@@ -22,7 +22,20 @@ import okhttp3.HttpUrl
...
 
-internal sealed class LoggerState {
+/**
+ * Emits changes to [LoggerState] upon a call to Capture.Logger.start
+ */
+fun interface LoggerStateListener {
+    /**
+     * Reports on each [LoggerState] changes
+     */
+    fun onLoggerStateUpdate(loggerState: LoggerState)
```
murki, line 32: is this only supposed to be called once? If so the name Update sounds misleading. We should also probably mention it in the kdocs if that's the case

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/Capture.kt
```diff
@@ -37,13 +50,18 @@ internal sealed class LoggerState {
      * The logger has been successfully started and is ready for use. Subsequent attempts to start the logger will be ignored.
      */
     class Started(
-        val logger: LoggerImpl,
+        internal val logger: ILogger,
```
murki, line 53: this is kinda cool, the callback now contains a fully initiated version of the ILogger which gives access to non-nullable pieces of data like the `sesssionId`, `deviceId`, etc 👍

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/Capture.kt
```diff
@@ -37,13 +50,18 @@ internal sealed class LoggerState {
...
 
     /**
      * An attempt to start the logger was made but failed. Subsequent attempts to start the logger will be ignored.
      */
-    data object StartFailure : LoggerState()
+    class StartFailure(
+        /**
+         * The reason for Capture.Logger.start() failure
+         */
+        val throwable: Throwable,
```
murki, line 63: for functional programming style I wouldn't return throwables (since they can't really be caught). Take a look at our existing `sealed class Error` for a better way to return async errors

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/Capture.kt
murki: for example, do we ever expect to return `Starting`? seems like that would be confusing

FranAguilera: good callout, was also debating this given probably we don't need the states othen that Started/Failure

Nice, will check that typealias

FranAguilera:
We won't be pursuing this at this time but we can reopen in the future

PR #221, opened 2025-02-12
[Android] Make Capture.Logger.start() return LoggerState

FranAguilera:
Capture.Logger.start now returns current LoggerState. This could be useful in case of Logger.start initialization errors

See demo below on gradle test app

https://github.com/user-attachments/assets/fd6cfa27-2f76-47ee-b394-b1ef01df7b06

### DIFF
--- a/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/Capture.kt
+++ b/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/Capture.kt
@@ -19,14 +19,18 @@ import io.bitdrift.capture.providers.FieldProvider
 import io.bitdrift.capture.providers.SystemDateProvider
 import io.bitdrift.capture.providers.session.SessionStrategy
 import okhttp3.HttpUrl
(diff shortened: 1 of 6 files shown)


platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/Capture.kt
```diff
@@ -19,14 +19,18 @@ import io.bitdrift.capture.providers.FieldProvider
 import io.bitdrift.capture.providers.SystemDateProvider
 import io.bitdrift.capture.providers.session.SessionStrategy
 import okhttp3.HttpUrl
+import org.jetbrains.annotations.VisibleForTesting
 import java.util.concurrent.atomic.AtomicReference
 import kotlin.time.Duration
 
-internal sealed class LoggerState {
+/**
+ * Represents the different Logger States
```
FranAguilera, line 27: lol this is not great comment, will update it

FranAguilera:
Will reopen after we discuss changes to setup api

PR #222, opened 2025-02-12
[Android] [Dropped Frames] Basic integration with JankStats

FranAguilera:
Basic Drop Frames/Non terminating ANR reporting via JankStats

- Bump library deps
- Moving existing WindowManager used by Replay into commons
- Basic reporting as UX type event and Error event

### **Please NOTE: Controlling the jankHeuristicMultiplier via Runtime will be added in a separate PR. See BIT-4665**

Below demo with local changes (Pending to polish UX/icon/colors/etc)

image


platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/events/performance/JankStatsMonitor.kt
```diff
@@ -0,0 +1,81 @@
...
+ * Reports Jank Frames and its duration in ms
+ *
+ * This will be a no-op when `client_feature.android.jank_stats_reporting` kill switch is disabled
+ */
+internal class JankStatsMonitor(
+    private val logger: LoggerImpl,
+    runtime: Runtime,
+) : IWindowListener,
+    JankStats.OnFrameListener {
+    private val shouldDisableMonitorViaKillSwitch = !runtime.isEnabled(RuntimeFeature.JANK_STATS_EVENTS)
```
murki, line 31: should probably check the runtime value every time the value needs to be read (in case it has changed in the meantime)

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/events/performance/JankStatsMonitor.kt
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    private fun FrameData.durationToMilli(): Long = this.frameDurationUiNanos / TO_MILLI
+
+    private fun FrameData.shouldReportJankyFrame(): Boolean =
+        this.isJank && this.durationToMilli() >= MIN_THRESHOLD_JANKY_FRAME_DURATION_IN_MILLI
+
+    private companion object {
+        private const val TO_MILLI = 1000000L
+        private const val MIN_THRESHOLD_JANKY_FRAME_DURATION_IN_MILLI = 16L
```
murki, line 77: we could make this a runtime configurable value so it's not hardcoded in the aar

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/LoggerImpl.kt
```diff
@@ -207,6 +208,8 @@ internal class LoggerImpl(
                         ProcessLifecycleOwner.get(),
                         runtime,
                         eventListenerDispatcher.executorService,
+                        errorHandler = errorHandler,
+                        windowListener = JankStatsMonitor(this, runtime),
```
murki, line 212: no need to check api level 24 and up?

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/events/performance/JankStatsMonitor.kt
```diff
@@ -0,0 +1,81 @@
...
+
+    /**
+     * Called when we have a new window available
+     */
+    override fun onWindowAvailable(window: Window) {
+        if (shouldDisableMonitorViaKillSwitch || currentWindow == window) {
+            return
+        }
+        currentWindow = window
+        jankStats = JankStats.createAndTrack(window, this)
```
murki, line 50: might need to handle error case. from their docs:
>Note: The window must be active with a non-null DecorView, otherwise it returns an IllegalStateException.

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/events/performance/JankStatsMonitor.kt
```diff
@@ -0,0 +1,81 @@
...
+    private fun sendJankStatData(frameData: FrameData) {
+        val message = "$ROW_MESSAGE ${frameData.durationToMilli()} ms"
+        val fields = mapOf(JANKY_FRAME_DURATION_KEY to "${frameData.durationToMilli()}")
+        logger.log(LogType.LIFECYCLE, LogLevel.WARNING, fields.toFields()) { message }
+    }
+
+    private fun FrameData.durationToMilli(): Long = this.frameDurationUiNanos / TO_MILLI
+
+    private fun FrameData.shouldReportJankyFrame(): Boolean =
+        this.isJank && this.durationToMilli() >= MIN_THRESHOLD_JANKY_FRAME_DURATION_IN_MILLI
```
murki, line 73: neither the docs example or documentation seem to check `durationToMilli` to determine whether there is jank as the boolean seem to be sufficient?
>Whether this frame was determined to be janky, meaning that its duration exceeds the duration determined by the system to indicate jank (@see [JankStats.jankHeuristicMultiplier](https://developer.android.com/reference/androidx/metrics/performance/JankStats#jankHeuristicMultiplier())).

platform/jvm/capture/src/test/kotlin/io/bitdrift/capture/AppLifecycleListenerLoggerTest.kt
```diff
@@ -32,13 +36,14 @@ class AppLifecycleListenerLoggerTest {
     fun testLogsAreFlushedOnStop() {
         // ARRANGE
         whenever(runtime.isEnabled(RuntimeFeature.APP_LIFECYCLE_EVENTS)).thenReturn(true)
-        val listener = AppLifecycleListenerLogger(logger, processLifecycleOwner, runtime, executor, handler)
+        val listener = AppLifecycleListenerLogger(logger, processLifecycleOwner, runtime, executor, handler, errorHandler, windowListener)
```
murki, line 39: pls add tests

murki:
@FranAguilera I know this is still WIP so my comments are just an early look, feel free to disregard

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/events/performance/JankStatsMonitor.kt
FranAguilera: good call, will update soon

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/events/performance/JankStatsMonitor.kt
FranAguilera: totally agree, I was wondering how we handle configuration of non boolean values there but we can sync offline

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/LoggerImpl.kt
FranAguilera: Good catch, that's definitely needed. Will update

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/events/performance/JankStatsMonitor.kt
FranAguilera: yup, but was wondering if this value should be configurable. Maybe 16 ms it's too much and we care mostly about Frozen frames etc for ANR case initially. Let's discuss it offline

platform/jvm/capture/src/test/kotlin/io/bitdrift/capture/AppLifecycleListenerLoggerTest.kt
FranAguilera: Yup, was planning to add those soon. Still prototyping a bunch and stuff can change so was going to add after this is more consolidated

FranAguilera:
> @FranAguilera I know this is still WIP so my comments are just an early look, feel free to disregard

np! appreciate the early feedback

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/LoggerImpl.kt
murki: actually after further reading [their docs](https://developer.android.com/topic/performance/jankstats#capabilities) i'm not sure it's needed? you should definitely test in an emulator with api < 24 to verify how it works:
>JankStats builds on top of the existing Android platform capabilities, including the [FrameMetrics API](https://developer.android.com/reference/android/view/FrameMetrics) on Android 7 (API level 24) and higher or [OnPreDrawListener](https://developer.android.com/reference/android/view/ViewTreeObserver.OnPreDrawListener) on earlier versions.

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/events/performance/JankStatsMonitor.kt
FranAguilera: Planning this in follow up PR

murki: Ok but let's just make sure to _not_ release a new sdk version from main until it has been added

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/events/lifecycle/AppLifecycleListenerLogger.kt
```diff
@@ -68,5 +72,16 @@ internal class AppLifecycleListenerLogger(
...
+        emitWindowChanges(event)
+    }
+
+    private fun emitWindowChanges(event: Lifecycle.Event) {
+        if (event == Lifecycle.Event.ON_RESUME) {
+            windowManager.getCurrentWindow()?.let {
+                lifecycleWindowListener.onWindowAvailable(it)
+            }
+        } else if (event == Lifecycle.Event.ON_STOP) {
+            lifecycleWindowListener.onWindowRemoved()
```
murki, line 84: do we need to call this method on the stop() method above? (assuming we would want jank stats to stop being reported in that case)

murki: if you want a clear separation of concerns I would just make the Jank Stat Monitor be its own `IEventListenerLogger, LifecycleEventObserver` vs attaching the operations to the `AppLifecycleListenerLogger`

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/events/performance/JankStatsMonitor.kt
```diff
@@ -0,0 +1,169 @@
...
+            stopCollection()
+            return
+        }
+
+        if (volatileFrameData.isJank) {
+            // Below API 24 [onFrame(volatileFrameData)] call happens on the main thread
+            if (Build.VERSION.SDK_INT < Build.VERSION_CODES.N) {
+                backgroundThreadHandler.runAsync { volatileFrameData.sendJankFrameData() }
+            } else {
+                // For >= 24 this happens on `FrameMetricsAggregator` thread
```
murki, line 65: nice catch

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/events/performance/JankStatsMonitor.kt
```diff
@@ -0,0 +1,169 @@
...
+                volatileFrameData.sendJankFrameData()
+            }
+        }
+    }
+
+    private fun setJankStatsForCurrentWindow(window: Window) {
+        try {
+            jankStats = JankStats.createAndTrack(window, this)
+
+            // BIT-4665 To update Runtime to provide heuristics value from config
```
murki, line 75: can you mark these with `// TODO(FranAguilera): ...` that's the current format we follow (plus allows you to list all the todos in android studio)

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/events/performance/JankStatsMonitor.kt
```diff
@@ -0,0 +1,169 @@
...
+                    LogLevel.ERROR,
+                    ANR_MESSAGE_ID,
+                )
+            }
+        }
+
+    private fun FrameData.toJankType(): JankFrameType =
+        if (this.durationToMilli() < FROZEN_FRAME_THRESHOLD_MS) {
+            JankFrameType.SLOW
+        } else if (this.durationToMilli() < ANR_FRAME_THRESHOLD_MS) {
```
murki, line 128: these 2 thresholds should also be runtime-configurable (or add TODOs)

platform/jvm/common/src/main/kotlin/io/bitdrift/capture/common/Runtime.kt
```diff
@@ -62,6 +62,11 @@ sealed class RuntimeFeature(
      * Whether the logger should be flushed on crash.
      */
     data object LOGGER_FLUSHING_ON_CRASH : RuntimeFeature("client_feature.android.logger_flushing_on_force_quit", defaultValue = true)
+
+    /**
+     * Whether Jank Stats reporting is enabled
+     */
+    data object JANK_STATS_EVENTS : RuntimeFeature("client_feature.android.jank_stats_reporting")
```
murki, line 69: could we rename this to something like "dropped_frames_..."

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/events/performance/JankStatsMonitor.kt
```diff
@@ -0,0 +1,169 @@
...
+    private val backgroundThreadHandler: IBackgroundThreadHandler = CaptureDispatchers.CommonBackground,
+) : ILifecycleWindowListener,
+    JankStats.OnFrameListener {
+    private var jankStats: JankStats? = null
+
+    override fun onWindowAvailable(window: Window) {
+        stopCollection()
+
+        if (runtime.isEnabled(RuntimeFeature.JANK_STATS_EVENTS)) {
+            setJankStatsForCurrentWindow(window)
```
murki, line 46: is there any sanity checks that we needs to do on the returned phoneWindow? I remember not being super reliable

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/events/performance/JankStatsMonitor.kt
```diff
@@ -0,0 +1,169 @@
...
+    override fun onFrame(volatileFrameData: FrameData) {
+        if (!runtime.isEnabled(RuntimeFeature.JANK_STATS_EVENTS)) {
+            stopCollection()
+            return
+        }
+
+        if (volatileFrameData.isJank) {
+            // Below API 24 [onFrame(volatileFrameData)] call happens on the main thread
+            if (Build.VERSION.SDK_INT < Build.VERSION_CODES.N) {
+                backgroundThreadHandler.runAsync { volatileFrameData.sendJankFrameData() }
```
murki, line 63: according to the docs for volatileFrameData:
>The FrameData object will be modified internally after returning from the listener; any data that needs to be retained should be copied before returning.

do we need to so something in the async case to make sure we don't have race conditions?

murki (approved):
lgtm - with the condition that we make the configurable runtime changes before releasing a new version

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/events/lifecycle/AppLifecycleListenerLogger.kt
FranAguilera: I agree, will update shortly

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/events/performance/JankStatsMonitor.kt
FranAguilera: I will double check on existing Replay impl

murki:
ios ci error could be a flake, re-running...

platform/jvm/common/src/main/kotlin/io/bitdrift/capture/common/Runtime.kt
FranAguilera: nice catch, forgot to update after proposal review

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/events/performance/JankStatsMonitor.kt
FranAguilera: sounds good, will add TODO for BIT-4665

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/events/performance/JankStatsMonitor.kt
FranAguilera: I thought about it earlier and did some manual verification on API 21 and so far seems safe. 

JankStats instance needs to be created on main thread so that doesn't change either way and we are not storing state of frameData either

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/events/performance/JankStatsMonitor.kt
FranAguilera: so far I saw that JankStats create just needs a valid decorView. This is covered by error log as well. Either way will keep and eye and do further manual verification after landing and before release

PR #228, opened 2025-02-19
[Android] [Dropped Frames] Adding Runtime.ConfigValue

FranAguilera:
## What

Follow up from https://github.com/bitdriftlabs/capture-sdk/pull/222 where now these fields are controlled by RuntimeConfig as listed on the proposal doc

- `android.jank_frame_heuristics_multiplier` with default of 2
- `android.frozen_frame_threshold` with default of 700ms
- `android.anr_frame_threshold` with default of 5000ms

### DIFF
--- a/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/events/performance/JankStatsMonitor.kt
+++ b/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/events/performance/JankStatsMonitor.kt
@@ -24,6 +24,7 @@ import io.bitdrift.capture.common.IBackgroundThreadHandler
 import io.bitdrift.capture.common.IWindowManager
 import io.bitdrift.capture.common.MainThreadHandler
 import io.bitdrift.capture.common.Runtime
+import io.bitdrift.capture.common.RuntimeConfig
 import io.bitdrift.capture.common.RuntimeFeature
 import io.bitdrift.capture.events.IEventListenerLogger
 import io.bitdrift.capture.events.span.SpanField
@@ -48,6 +49,14 @@ internal class JankStatsMonitor(
     LifecycleEventObserver,
     JankStats.OnFrameListener {
     private var jankStats: JankStats? = null
+
+    private val frozenFrameThreshold by lazy {
+        runtime.getConfigValue(RuntimeConfig.FROZEN_FRAME_THRESHOLD_IN_MILLI_SECONDS)
+    }
+
+    private val anrFrameThreshold by lazy {
+        runtime.getConfigValue(RuntimeConfig.ANR_FRAME_THRESHOLD_IN_MILLI_SECONDS)
+    }
 
     override fun start() {
         mainThreadHandler.run {
@@ -98,9 +107,7 @@ internal class JankStatsMonitor(
             }
 
             jankStats = JankStats.createAndTrack(window, this)
-
-            // TODO(FranAguilera): BIT-4665 To update Runtime to provide heuristics value from config
-            jankStats?.jankHeuristicMultiplier = DEFAULT_JANK_HEURISTICS_MULTIPLIER
+            jankStats?.jankHeuristicMultiplier = runtime.getConfigValue(RuntimeConfig.JANK_FRAME_HEURISTICS_MULTIPLIER).toFloat()
         } catch (illegalStateException: IllegalStateException) {
             errorHandler.handleError(
                 "Couldn't create JankStats instance",
@@ -150,9 +157,9 @@ internal class JankStatsMonitor(
         }
 
     private fun FrameData.toJankType(): JankFrameType =
-        if (this.durationToMilli() < DEFAULT_FROZEN_FRAME_THRESHOLD_MS) {
+        if (this.durationToMilli() < frozenFrameThreshold) {
             JankFrameType.SLOW
-        } else if (this.durationToMilli() < DEFAULT_ANR_FRAME_THRESHOLD_MS) {
+        } else if (this.durationToMilli() < anrFrameThreshold) {
             JankFrameType.FROZEN
         } else {
             JankFrameType.ANR
@@ -189,12 +196,5 @@ internal class JankStatsMonitor(
         private const val TO_MILLI = 1_000_000L
         private const val DROPPED_FRAME_MESSAGE_ID = "DroppedFrame"
         private const val ANR_MESSAGE_ID = "ANR"
-
-        // TODO(FranAguilera): BIT-4665 To update Runtime to provide heuristics value from config
-        private const val DEFAULT_JANK_HEURISTICS_MULTIPLIER = 2.0F
-
-        // TODO(FranAguilera): BIT-4665 To update Runtime to provide these thresholds from config
-        private const val DEFAULT_FROZEN_FRAME_THRESHOLD_MS = 700L
-        private const val DEFAULT_ANR_FRAME_THRESHOLD_MS = 5000L
     }
 }
--- a/platform/jvm/capture/src/test/kotlin/io/bitdrift/capture/JankStatsMonitorTest.kt
+++ b/platform/jvm/capture/src/test/kotlin/io/bitdrift/capture/JankStatsMonitorTest.kt
@@ -21,6 +21,7 @@ import com.nhaarman.mockitokotlin2.verify
 import com.nhaarman.mockitokotlin2.whenever
 import io.bitdrift.capture.common.IWindowManager
 import io.bitdrift.capture.common.Runtime
+import io.bitdrift.capture.common.RuntimeConfig
 import io.bitdrift.capture.common.RuntimeFeature
 import io.bitdrift.capture.events.performance.JankStatsMonitor
 import io.bitdrift.capture.fakes.FakeBackgroundThreadHandler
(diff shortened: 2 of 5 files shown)


platform/jvm/common/src/main/kotlin/io/bitdrift/capture/common/Runtime.kt
```diff
@@ -69,6 +69,43 @@ sealed class RuntimeFeature(
...
+ * @param defaultValue The value
+ */
+sealed class RuntimeConfig(
+    val configName: String,
+    val defaultValue: Int,
+) {
+    /**
+     * The configured value for [jankHeuristicMultiplier] from [JankStatsMonitoring]. More info at https://developer.android.com/topic/performance/jankstats#jank-heuristics
+     */
+    data object JANK_FRAME_HEURISTICS_MULTIPLIER : RuntimeConfig("client_feature.android.jank_frame_heuristics", 2)
```
FranAguilera, line 84: probably better to name this `client_config` but maybe this breaks some stuff

platform/jvm/src/jni.rs
```diff
@@ -1181,6 +1181,31 @@ pub extern "system" fn Java_io_bitdrift_capture_Jni_isRuntimeEnabled(
   .into()
 }
 
+#[no_mangle]
+pub extern "system" fn Java_io_bitdrift_capture_Jni_captureRuntimeIntValue(
```
FranAguilera, line 1185: Pending to test how can I update this config for real

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/events/performance/JankStatsMonitor.kt
```diff
@@ -49,6 +50,14 @@ internal class JankStatsMonitor(
     JankStats.OnFrameListener {
     private var jankStats: JankStats? = null
 
+    private val frozenFrameThreshold by lazy {
+        runtime.getConfigValue(RuntimeConfig.FROZEN_FRAME_THRESHOLD_IN_MILLI_SECONDS)
```
FranAguilera, line 54: @murki let me know what you think. These 2 are lazy and instance class variable intentionally just to avoid constant reads to JNI config value

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/events/performance/JankStatsMonitor.kt
```diff
@@ -98,9 +107,7 @@ internal class JankStatsMonitor(
             }
 
             jankStats = JankStats.createAndTrack(window, this)
-
-            // TODO(FranAguilera): BIT-4665 To update Runtime to provide heuristics value from config
-            jankStats?.jankHeuristicMultiplier = DEFAULT_JANK_HEURISTICS_MULTIPLIER
+            jankStats?.jankHeuristicMultiplier = runtime.getConfigValue(RuntimeConfig.JANK_FRAME_HEURISTICS_MULTIPLIER).toFloat()
```
FranAguilera, line 110: This is also intentional so we can change the multiplier with cold launch

platform/jvm/src/jni.rs
```diff
@@ -1181,6 +1181,31 @@ pub extern "system" fn Java_io_bitdrift_capture_Jni_isRuntimeEnabled(
...
+) -> jint {
+  bd_client_common::error::with_handle_unexpected_or(
+    || {
+      let logger = unsafe { LoggerId::from_raw(logger_id) };
+      let binding = unsafe { env.get_string_unchecked(&variable_name) }?;
+      let variable_name = binding.to_str()?;
+      Ok(
+        logger
+          .runtime_snapshot()
+          .get_integer(variable_name, default_value as u32) as jint,
```
FranAguilera, line 1200: Probably there is a better way to convert this type, can investigate further

platform/jvm/src/jni.rs
murki, line 1185: oh and you also have to add this method signature to https://github.com/bitdriftlabs/capture-sdk/blob/b2d27d34f7951772101eb7016b806f325dadcc07/platform/jvm/jni_symbols.lds

platform/jvm/common/src/main/kotlin/io/bitdrift/capture/common/Runtime.kt
```diff
@@ -69,6 +69,43 @@ sealed class RuntimeFeature(
...
+    /**
+     * The upper bound threshold that defines what constitutes an ANR frame reported via [JankStats]
+     *
+     * The default value is 5000ms
+     *
+     * Slow Frame: >=16ms to < FROZEN_FRAME_THRESHOLD_IN_MILLI_SECONDS
+     * Frozen Frame: >=FROZEN_FRAME_THRESHOLD_IN_MILLI_SECONDS to <ANR_FRAME_THRESHOLD_IN_MILLI_SECONDS
+     * ANR Frame: >=ANR_FRAME_THRESHOLD_IN_MILLI_SECONDS
+     */
+    data object ANR_FRAME_THRESHOLD_IN_MILLI_SECONDS : RuntimeConfig("client_feature.android.frozen_frame.threshold_ms, 5000)
```
murki, line 106:
```suggestion
    data object ANR_FRAME_THRESHOLD_IN_MILLI_SECONDS : RuntimeConfig("client_feature.android.anr.threshold_ms, 5000)
```

copypasta typo

platform/jvm/common/src/main/kotlin/io/bitdrift/capture/common/Runtime.kt
```diff
@@ -69,6 +69,43 @@ sealed class RuntimeFeature(
...
+    /**
+     * The upper bound threshold that defines what constitutes a FROZEN frame reported via [JankStats]
+     *
+     * The default value is 700ms
+     *
+     * Slow Frame: >=16ms to < FROZEN_FRAME_THRESHOLD_IN_MILLI_SECONDS
+     * Frozen Frame: >=FROZEN_FRAME_THRESHOLD_IN_MILLI_SECONDS to <ANR_FRAME_THRESHOLD_IN_MILLI_SECONDS
+     * ANR Frame: >=ANR_FRAME_THRESHOLD_IN_MILLI_SECONDS
+     */
+    data object FROZEN_FRAME_THRESHOLD_IN_MILLI_SECONDS : RuntimeConfig("client_feature.android.frozen_frame.threshold_ms", 700)
```
murki, line 95: I would just shorten to `FROZEN_FRAME_THRESHOLD_MS` which is common usage of suffixes for units

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/events/performance/JankStatsMonitor.kt
murki: I mean we're already checking `runtime.isEnabled(RuntimeFeature.DROPPED_EVENTS_MONITORING)` on every `onFrame()` (and everywhere else in the sdk code), so I'm not sure we'll get much benefit from this micro-optimization

platform/jvm/src/jni.rs
murki: were you able to run the app and get jank frames reported? seems like that should be test for this (it should crash now since you haven't enumerated the new symbol in jni_symbols.lds as mentioned below)

otherwise we could add some tmp logs at the rust layer and I can show you how to view them

platform/jvm/src/jni.rs
murki: i think this should be fine right @snowp ?

platform/jvm/capture/src/test/kotlin/io/bitdrift/capture/JankStatsMonitorTest.kt
```diff
@@ -111,6 +115,20 @@ class JankStatsMonitorTest {
...
+        whenever(runtime.getConfigValue(RuntimeConfig.ANR_FRAME_THRESHOLD_IN_MILLI_SECONDS)).thenReturn(2000)
+        val jankDurationInMilli = 2000L
+
+        triggerLifecycleEvent(
+            lifecycleEvent = Lifecycle.Event.ON_RESUME,
+            isJankyFrame = true,
+            durationInMilli = jankDurationInMilli,
+        )
+
+        assertLogDetails(jankDurationInMilli, LogLevel.ERROR, "ANR")
```
murki, line 129: 👍

platform/jvm/src/jni.rs
snowp: It's fine but a little bit shady - the Java interface allows you to pass a negative number (jint is signed 32 bit) so if you were to pass -1 `get_integer()` would return `4294967295` but then since you are casting it back to `jint` it would go back to being -1

The perhaps misleading thing is that with the current setup the only way you can get a negative number is via the default parameter, as the underlying runtime system operates on unsigned values

As for better ways of converting you can use `try_from` if you want to specify a default handling when conversion fails but it doesn't seem like a big deal

platform/jvm/common/src/main/kotlin/io/bitdrift/capture/common/Runtime.kt
FranAguilera: good catch, I was renaming this at some point to match iOS but I'm going to keep it as anr_frame for clarity

platform/jvm/src/jni.rs
FranAguilera: Validate on both bazel/gradle apps. Will include session on PR description

murki (approved):

murki (changes requested):
Actually @FranAguilera as part of this PR could you also include a change for adding the `_duration_ms` field to the existing ANR event on iOS? https://github.com/bitdriftlabs/capture-sdk/blob/494518a0cf13ca79dde983538fd842ed04a62d69/platform/swift/source/default/events/listeners/ANRReporter.swift#L91
should be a simple change

FranAguilera:
> Actually @FranAguilera as part of this PR could you also include a change for adding the `_duration_ms` field to the existing ANR event on iOS?
> 
> https://github.com/bitdriftlabs/capture-sdk/blob/494518a0cf13ca79dde983538fd842ed04a62d69/platform/swift/source/default/events/listeners/ANRReporter.swift#L91
> 
> 
> should be a simple change

Fyi @kattrali 

Discussed offline with @murki , for just `ANR` report this will be tricky as we don't know when main thread will be responsive again so _duration_ms is not reported. 

But for `ANREnd` we do have that duration and is being reported under _duration_ms. e.g. 
image

PR #231, opened 2025-02-20
[Android] [Dropped Frames] Adding another config value for min threshold

FranAguilera:

Adding another config threshold for what is considered Jank.

On local testing and changing to more strict heuristics value was still getting frames reported as janks below 16 ms (e.g. 4 ms)

In this manner we can also control a bit more the lower threshold in case this becomes super noisy

### DIFF
--- a/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/events/performance/JankStatsMonitor.kt
+++ b/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/events/performance/JankStatsMonitor.kt
@@ -52,8 +52,6 @@ internal class JankStatsMonitor(
     JankStats.OnFrameListener {
     private var jankStats: JankStats? = null
 
-    private val minJankFrameThresholdMs by lazy { runtime.getConfigValue(RuntimeConfig.MIN_JANK_FRAME_THRESHOLD_MS) }
-
     override fun start() {
         mainThreadHandler.run {
             processLifecycleOwner.lifecycle.addObserver(this)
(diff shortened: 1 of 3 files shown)


murki (approved):
i agree be better safe than sorry, though it worries me that we're relying on a library that doesn't seem work as advertised

FranAguilera:
Discussed offline, abandoning this. Will provide more context soon

FranAguilera:
Reopening in order to have more control on the lower bound threshold for the jank duration

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/events/performance/JankStatsMonitor.kt
```diff
@@ -92,6 +89,12 @@ internal class JankStatsMonitor(
         }
     }
 
+    /**
+     * Will only report frames as Jank if duration is at least the configured value at `MIN_JANK_FRAME_THRESHOLD_MS` (which is a default of 16 ms)
+     * This is given that depending on device characteristics, frame can be reported as Jank even the duration is below 16 ms
+     */
+    private fun FrameData.isJankAndMinDurationReached(): Boolean = isJank && minJankFrameThresholdMs > this.durationToMilli()
```
murki, line 96: so we're no longer checking `runtime.isEnabled(RuntimeFeature.DROPPED_EVENTS_MONITORING` here? also if I remember my boolean logic classes that second part of a logical AND shouldn't even be evaluated if the first conditions is false no? That would make the lazy optimization unnecesary

PR #232, opened 2025-02-20
[Android][Dropped Frames] Add API level checks for duration

FranAguilera:
Kudos to @murki for the investigation

In order to retrieve more accurate total jank frame duration, adding API level check


murki (approved):

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/events/performance/JankStatsMonitor.kt
```diff
@@ -157,7 +159,17 @@ internal class JankStatsMonitor(
             JankFrameType.ANR
         }
 
-    private fun FrameData.durationToMilli(): Long = this.frameDurationUiNanos / TO_MILLI
+    private fun FrameData.durationToMilli(): Long {
```
murki, line 162: also an un-related reminder to move the `runtime.isEnabled(RuntimeFeature.DROPPED_EVENTS_MONITORING)` check that happens `onFrame()` inside the `if (volatileFrameData.isJank) {}` check

FranAguilera:
So far looks good on local testing. Auto-merge once CI completes

PR #236, opened 2025-02-26
[Android] OkHttp Event listener filtering

FranAguilera:
More details at https://linear.app/bitdrift/issue/BIT-4704 

Verified session https://timeline.bitdrift.dev/s/b253183b-68d5-440e-9c26-5e2ecc5411e5?utm_source=sdk

### DIFF
--- a/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/network/okhttp/CaptureOkHttpEventListener.kt
+++ b/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/network/okhttp/CaptureOkHttpEventListener.kt
@@ -92,6 +92,9 @@ internal class CaptureOkHttpEventListener internal constructor(
 
     override fun callStart(call: Call) {
         runCatching { targetEventListener?.callStart(call) }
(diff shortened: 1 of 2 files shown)


platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/network/okhttp/CaptureOkHttpEventListener.kt
```diff
@@ -91,6 +91,10 @@ internal class CaptureOkHttpEventListener internal constructor(
     private var lastResponse: Response? = null
 
     override fun callStart(call: Call) {
+        if (call.shouldIgnoreCurrentPath()) {
+            return
```
murki, line 95: IIUC this check is all that would be really needed? since when true it would never initialize requestInfo and those null checks would take care of the rest no?

my suggestion would be to just have this one check (probably move it after the targetEventListener forwarding) and then greatly simplify this change

murki (changes requested):

murki (approved):

PR #237, opened 2025-02-26
include build-id in release android builds

snowp:
This passes -Wl,--build-id during linking for the release build which should make it easier to symbolicate frames coming from the libcapture.so binary


platform/jvm/BUILD.bazel
```diff
@@ -14,6 +14,7 @@ cc_binary(
     linkopts = [
         "-lm",  # Required to avoid dlopen runtime failures unrelated to rust
         "-lz",  # Link against system zlib library
+        "-Wl,--build-id", # Include build id in the binary
```
FranAguilera, line 17: shouldn't this be part of "@platforms//os:androi

FranAguilera (approved):

PR #240, opened 2025-02-26
[Android] Send crash artifact 

FranAguilera:
More details at https://linear.app/bitdrift/issue/BIT-4742

Related PR https://github.com/bitdriftlabs/loop-api/pull/2095

Sample session with multiple native crashes https://timeline.bitdrift.dev/session/f0c2ad55-66ae-498f-9f85-13a88b18eee1?pages=1&utilization=0

### DIFF
--- a/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/events/lifecycle/AppExitLogger.kt
+++ b/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/events/lifecycle/AppExitLogger.kt
@@ -24,6 +24,7 @@ import io.bitdrift.capture.common.IBackgroundThreadHandler
 import io.bitdrift.capture.common.Runtime
 import io.bitdrift.capture.common.RuntimeFeature
 import io.bitdrift.capture.events.performance.IMemoryMetricsProvider
+import io.bitdrift.capture.providers.toFieldValue
 import io.bitdrift.capture.providers.toFields
 import io.bitdrift.capture.threading.CaptureDispatchers
 import io.bitdrift.capture.utils.BuildVersionChecker
@@ -53,6 +54,7 @@ internal class AppExitLogger(
         private const val APP_EXIT_PSS_KEY = "_app_exit_pss"
         private const val APP_EXIT_RSS_KEY = "_app_exit_rss"
         private const val APP_EXIT_DESCRIPTION_KEY = "_app_exit_description"
+        private const val APP_EXIT_CRASH_ARTIFACT_KEY = "_crash_artifact"
     }
 
     @WorkerThread
@@ -116,7 +118,7 @@ internal class AppExitLogger(
         logger.log(
             LogType.LIFECYCLE,
             lastExitInfo.reason.toLogLevel(),
-            buildAppExitAndMemoryFieldsMap(lastExitInfo),
+            buildAppExitInternalFieldsMap(lastExitInfo),
             attributesOverrides = LogAttributesOverrides(sessionId, timestampMs),
         ) { APP_EXIT_EVENT_NAME }
     }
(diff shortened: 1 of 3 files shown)


platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/events/lifecycle/AppExitLogger.kt
```diff
@@ -171,11 +173,27 @@ internal class AppExitLogger(
...
+            putAll(applicationExitInfo.toMap().toFields())
+            putAll(memoryMetricsProvider.getMemoryAttributes().toFields())
+            applicationExitInfo.getNativeCrashByteArray()?.let {
+                put(APP_EXIT_CRASH_ARTIFACT_KEY, it.toFieldValue())
+            }
+        }
+
+    @TargetApi(Build.VERSION_CODES.R)
+    private fun ApplicationExitInfo.getNativeCrashByteArray(): ByteArray? {
+        if (reason == ApplicationExitInfo.REASON_CRASH_NATIVE &&
```
murki, line 187: do we need to check the api level too?

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/events/lifecycle/AppExitLogger.kt
```diff
@@ -171,11 +173,27 @@ internal class AppExitLogger(
...
+            traceInputStream != null &&
+            runtime.isEnabled(RuntimeFeature.SEND_CRASH_ARTIFACT)
+        ) {
+            traceInputStream?.let {
```
murki, line 191: should we use `use` or something to make sure the stream is closed? or does readBytes handles it?

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/events/lifecycle/AppExitLogger.kt
```diff
@@ -171,11 +173,27 @@ internal class AppExitLogger(
...
+                return it.readBytes()
```
murki, line 192: probably need error handling

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/events/lifecycle/AppExitLogger.kt
FranAguilera: I believe is not needed when API below 31 as the traceInputStream will be null 
ref: https://developer.android.com/reference/android/app/ApplicationExitInfo#getTraceInputStream()

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/events/lifecycle/AppExitLogger.kt
FranAguilera: good callout. Will update

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/events/lifecycle/AppExitLogger.kt
FranAguilera: good catch, we should ensure is closed

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/events/lifecycle/AppExitLogger.kt
FranAguilera: Confirmed while testing on API 30

murki (approved):

platform/jvm/common/src/main/kotlin/io/bitdrift/capture/common/Runtime.kt
```diff
@@ -67,6 +67,11 @@ sealed class RuntimeFeature(
      * Whether Dropped Frames reporting is enabled
      */
     data object DROPPED_EVENTS_MONITORING : RuntimeFeature("client_feature.android.dropped_frames_reporting")
+
+    /**
+     * Whether we should send crash artifact data via AppExit
+     */
+    data object SEND_CRASH_ARTIFACT : RuntimeFeature("client_feature.android.send_crash_artifact")
```
FranAguilera, line 74: note for self: This behaves as killswitch and default value is always set to true (if not passed in)

PR #241, opened 2025-02-27
[Android] [GradleTestApp] Add session strategy configuration option

FranAguilera:
Just small tweak on the gradle test app in order to easy tweak SessionStrategy type without having to hardcode it locally (default still remains as Fixed)

config_fragment

session_strategy_options


murki (approved):
good idea

PR #243, opened 2025-03-04
Add some missing new ObjC interfaces

Reflejo:


murki (approved):
objective-c making a comeback!

PR #245, opened 2025-03-04
[Android] Memory Leak fix on JankStatsMonitor

FranAguilera:
Verified with leakcanary enabled on debug build and that dropped frames are still being send upon configuration changes

https://timeline.bitdrift.dev/session/1d4ca361-3037-43d0-b449-3085fbb06b90?pages=1&utilization=0&expanded=1070970770558996757


murki (approved):
even though this is a band-aid i think it's important to ship a version with a fix until we do a more proper fix

PR #248, opened 2025-03-05
[Android] Follow up to bandaid memory leak fix at JankStatsMonitor

FranAguilera:
## What

https://github.com/bitdriftlabs/capture-sdk/pull/245 was a bandaid fix to avoid existing leak on configuration changes

The proposal here is to also rely on `ActivityLifecycleCallbacks` to get accurate signal of onResumed/onPaused activity to make sure JankStats ref is nullified

## Verification 

- Verified on latest on main that leak is no longer reported on configuration changes
- Verified dropped frames still reported upon configuration changes. See session in here 
https://timeline.bitdrift.dev/session/de4c4aef-0cad-4d53-aae1-c2215dae5951?pages=1&utilization=0&expanded=7024628788602156679

image

### DIFF
--- a/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/LoggerImpl.kt
+++ b/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/LoggerImpl.kt
@@ -9,6 +9,7 @@ package io.bitdrift.capture
 
 import android.annotation.SuppressLint
 import android.app.ActivityManager
+import android.app.Application
 import android.content.Context
 import android.content.pm.ApplicationInfo
 import android.system.Os
@@ -89,7 +90,7 @@ internal class LoggerImpl(
     private val diskUsageMonitor: DiskUsageMonitor
     private val appExitLogger: AppExitLogger
     private val runtime: JniRuntime
-    private val jankStatsMonitor: JankStatsMonitor
+    private var jankStatsMonitor: JankStatsMonitor? = null
 
     // we can assume a properly formatted api url is being used, so we can follow the same pattern
     // making sure we only replace the first occurrence
@@ -246,15 +247,7 @@ internal class LoggerImpl(
                     ),
                 )
 
-                jankStatsMonitor =
-                    JankStatsMonitor(
-                        this,
-                        runtime,
-                        windowManager,
-                        errorHandler,
-                    )
-
-                eventsListenerTarget.add(jankStatsMonitor)
+                addJankStatsMonitorTarget(windowManager, context)
 
                 appExitLogger =
                     AppExitLogger(
@@ -322,7 +315,7 @@ internal class LoggerImpl(
     }
 
     override fun logScreenView(screenName: String) {
-        jankStatsMonitor.trackScreenNameChanged(screenName)
+        jankStatsMonitor?.trackScreenNameChanged(screenName)
         CaptureJniLibrary.writeScreenViewLog(this.loggerId, screenName)
     }
 
@@ -512,6 +505,28 @@ internal class LoggerImpl(
                     Os.setenv("RUST_LOG", internalLogLevel, true)
                 }
             }
+        }
+    }
+
+    private fun addJankStatsMonitorTarget(
+        windowManager: IWindowManager,
+        context: Context,
+    ) {
+        if (context is Application) {
+            jankStatsMonitor =
+                JankStatsMonitor(
+                    context,
+                    this,
+                    ProcessLifecycleOwner.get(),
+                    runtime,
+                    windowManager,
+                    errorHandler,
+                )
+            jankStatsMonitor?.let {
+                eventsListenerTarget.add(it)
+            }
+        } else {
+            errorHandler.handleError("Couldn't start JankStatsMonitor", IllegalArgumentException("Invalid application provided"))
         }
     }
 }
--- a/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/events/performance/JankStatsMonitor.kt
+++ b/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/events/performance/JankStatsMonitor.kt
@@ -13,14 +13,17 @@ import android.os.Build
 import android.os.Bundle
 import android.view.Window
 import androidx.annotation.UiThread
+import androidx.annotation.VisibleForTesting
 import androidx.annotation.WorkerThread
+import androidx.lifecycle.Lifecycle
+import androidx.lifecycle.LifecycleEventObserver
+import androidx.lifecycle.LifecycleOwner
 import androidx.metrics.performance.FrameData
 import androidx.metrics.performance.FrameDataApi24
 import androidx.metrics.performance.FrameDataApi31
 import androidx.metrics.performance.JankStats
 import androidx.metrics.performance.PerformanceMetricsState
 import androidx.metrics.performance.StateInfo
-import io.bitdrift.capture.ContextHolder.Companion.APP_CONTEXT
 import io.bitdrift.capture.ErrorHandler
 import io.bitdrift.capture.LogLevel
 import io.bitdrift.capture.LogType
@@ -44,7 +47,9 @@ import io.bitdrift.capture.threading.CaptureDispatchers
  *
  */
 internal class JankStatsMonitor(
+    private val application: Application,
     private val logger: LoggerImpl,
+    private val processLifecycleOwner: LifecycleOwner,
     private val runtime: Runtime,
     private val windowManager: IWindowManager,
     private val errorHandler: ErrorHandler,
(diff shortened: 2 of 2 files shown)


platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/LoggerImpl.kt
```diff
@@ -248,8 +250,8 @@ internal class LoggerImpl(
 
                 jankStatsMonitor =
                     JankStatsMonitor(
+                        context as Application,
```
FranAguilera, line 253: this is the application context which is guarded via Capture.logger.start internals

FranAguilera: if (!ContextHolder.isInitialized) {
    Log.w(
        "capture",
        "Attempted to initialize Capture before androidx.startup.Initializers " +
            "are run. Aborting logger initialization.",
    )
    return
}

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/events/performance/JankStatsMonitor.kt
```diff
@@ -44,48 +44,29 @@ import java.lang.ref.WeakReference
  *
  */
 internal class JankStatsMonitor(
+    private val application: Application,
     private val logger: LoggerImpl,
-    private val processLifecycleOwner: LifecycleOwner,
     private val runtime: Runtime,
     private val windowManager: IWindowManager,
```
FranAguilera, line 50: @murki I can do a follow up to remove windowmanager. Or I can do it here, my idea was to store PerformanceMetricsState.Holder once we get activity at onResume so we don't need to call each time `windowManager.getFirstRootView()`

FranAguilera: actually went ahead and removed IWindowManager  here

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/LoggerImpl.kt
murki: I'm not sure this is guaranteed to be the Application _class_, it's for sure [guarded to be the applicationContext](https://github.com/bitdriftlabs/capture-sdk/blob/2c5fd973fa7d73dc7e43435ae3c37d06f55c9c16/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/ContextHolder.kt#L20), but I'm not sure if it's guaranteed to always be 100% the same.

We should probably do a check to avoid crashing in the odd case

FranAguilera: Sounds good, let me add it

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/events/performance/JankStatsMonitor.kt
```diff
@@ -114,16 +94,56 @@ internal class JankStatsMonitor(
...
             if (!runtime.isEnabled(RuntimeFeature.DROPPED_EVENTS_MONITORING)) {
                 stopCollection()
                 return
             }
-
-            jankStatsWeakRef = WeakReference(JankStats.createAndTrack(window, this))
-            getJankStats()?.jankHeuristicMultiplier = runtime.getConfigValue(RuntimeConfig.JANK_FRAME_HEURISTICS_MULTIPLIER).toFloat()
+            jankStats = JankStats.createAndTrack(window, this)
+            performanceMetricsStateHolder = PerformanceMetricsState.getHolderForHierarchy(window.decorView.rootView)
+            jankStats?.jankHeuristicMultiplier = runtime.getConfigValue(RuntimeConfig.JANK_FRAME_HEURISTICS_MULTIPLIER).toFloat()
```
murki, line 146: oh yeah all this is so much better

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/events/performance/JankStatsMonitor.kt
```diff
@@ -132,11 +152,10 @@ internal class JankStatsMonitor(
...
     }
 
-    private fun getJankStats(): JankStats? = jankStatsWeakRef?.get()
-
     private fun stopCollection() {
-        getJankStats()?.isTrackingEnabled = false
-        jankStatsWeakRef?.clear()
+        jankStats?.isTrackingEnabled = false
+        jankStats = null
+        performanceMetricsStateHolder = null
```
murki, line 158: we should also do `performanceMetricsStateHolder?.state?.removeState(SCREEN_NAME_KEY)` before nullifying it

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/events/performance/JankStatsMonitor.kt
```diff
@@ -44,48 +43,29 @@ import java.lang.ref.WeakReference
...
     private val errorHandler: ErrorHandler,
     private val mainThreadHandler: MainThreadHandler = MainThreadHandler(),
     private val backgroundThreadHandler: IBackgroundThreadHandler = CaptureDispatchers.CommonBackground,
 ) : IEventListenerLogger,
-    LifecycleEventObserver,
+    Application.ActivityLifecycleCallbacks,
     JankStats.OnFrameListener {
-    private var jankStatsWeakRef: WeakReference<JankStats>? = null
+    @VisibleForTesting
+    internal var jankStats: JankStats? = null
```
murki, line 56: why is this internal?

murki (approved):
lgtm - thanks for the quick followup!

murki:
@FranAguilera in the session you linked I'd expect to see the `_screen_name` field on the Slow/Drop Frames log, but [that doesn't seem to be the case](https://timeline.bitdrift.dev/session/d05e1f61-1490-4137-bf1f-a2718caf70c8?pages=1&utilization=0&type=7&h=5520075984124727523), do you know why?

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/events/performance/JankStatsMonitor.kt
FranAguilera: good catch, updating it now

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/events/performance/JankStatsMonitor.kt
FranAguilera: Just for testing purposes in order to ensure stopCollection by checking this is nullified upon onPause

FranAguilera:
> @FranAguilera in the session you linked I'd expect to see the `_screen_name` field on the Slow/Drop Frames log, but [that doesn't seem to be the case](https://timeline.bitdrift.dev/session/d05e1f61-1490-4137-bf1f-a2718caf70c8?pages=1&utilization=0&type=7&h=5520075984124727523), do you know why?

will double check

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/LoggerImpl.kt
murki: I'm not sure we should completely disable the Logger if Context != Application. as far as we know every other part of the logger works today without assuming the context is the application. so for the case where this is not true I would just "disable" the JankStatsMonitor feature, not the whole logger

FranAguilera:
> > @FranAguilera in the session you linked I'd expect to see the `_screen_name` field on the Slow/Drop Frames log, but [that doesn't seem to be the case](https://timeline.bitdrift.dev/session/d05e1f61-1490-4137-bf1f-a2718caf70c8?pages=1&utilization=0&type=7&h=5520075984124727523), do you know why?
> 
> will double check

Confirmed in this session

https://timeline.bitdrift.dev/session/de4c4aef-0cad-4d53-aae1-c2215dae5951?pages=1&utilization=0&expanded=9113230504299786866

image

image

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/LoggerImpl.kt
FranAguilera: oh actually, by checking the ContextHolder impl this should always be Application. 

`APP_CONTEXT = context.applicationContext` [source](https://cs.android.com/android/platform/superproject/+/android15-qpr1-release:frameworks/base/core/java/android/content/Context.java;l=865?q=public%20abstract%20Context%20getApplicationContext())

but can add the check before JankStatsMonitor instantiation if you think there is still some risk

murki: I'm not sure where it says it should always be application. Like I said before Application != applicationContext (which is of type Context). unless i'm missing something from the android source

FranAguilera: I'm not missing anything the source points to always be the application

https://android.googlesource.com/platform/frameworks/base/+/master/core/java/android/app/ContextImpl.java#476

FranAguilera: Discussed offline (thanks for sharing this https://stackoverflow.com/questions/5018545/getapplication-vs-getapplicationcontext/6760019#6760019)

TLDR: This can be tweaked by vendor so adding additional check to prevent runtime crash.

platform/jvm/gradle-test-app/src/main/res/layout/fragment_first.xml
```diff
@@ -1,178 +1,154 @@
 <?xml version="1.0" encoding="utf-8"?>
-<LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
+<ScrollView
```
FranAguilera, line 2: uops bad rebase. let me fix

PR #253, opened 2025-03-07
improve ergonomics of using the test server in platform tests

snowp:
Instead of repeatedly grabbing references from the static, wrap all the calls in one big closure. Using a newtype we can add extra test helpers to the handle that avoids having to lock the mutex every time


murki (approved):

FranAguilera:

PR #254, opened 2025-03-10
[Android] Initial proposal for pre init crash reporting method

FranAguilera:
- Introduce "experimental" `Capture.Logger.initCrashReporting` API.
- Status of that init function is passed to LoggerImpl (called upon first Capture.Logger.start), there will append additional map values to existing SDKConfigured . This contains the status of crash reporter and its duration when applicable
- For now opting for a config format such as 'root_directory_path,file_extension'. 
- Appended the epoch time of the original crash file that will be moved under reports/new
- Demo sessions [bazel sample app](https://timeline.bitdrift.dev/session/f150649a-1da8-4732-bf20-4b3b61226272?pages=1&utilization=0&expanded=4568865902884925366), [gradle sample app](https://timeline.bitdrift.dev/s/dddb3ced-0bc6-4078-9044-b90ef98c4022?utm_source=sdk)

### DIFF
--- a/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/Capture.kt
+++ b/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/Capture.kt
@@ -20,7 +20,8 @@ import io.bitdrift.capture.providers.FieldProvider
 import io.bitdrift.capture.providers.SystemDateProvider
 import io.bitdrift.capture.providers.session.SessionStrategy
 import io.bitdrift.capture.reports.CrashReporter
-import io.bitdrift.capture.reports.CrashReporter.CrashReporterState
+import io.bitdrift.capture.reports.CrashReporter.CrashReporterState.NotInitialized
+import io.bitdrift.capture.reports.CrashReporter.CrashReporterStatus
 import okhttp3.HttpUrl
 import java.util.concurrent.atomic.AtomicReference
 import kotlin.time.Duration
@@ -54,7 +55,7 @@ internal sealed class LoggerState {
  */
 object Capture {
     private val default: AtomicReference<LoggerState> = AtomicReference(LoggerState.NotStarted)
-    private var crashReporterState: CrashReporterState = CrashReporterState.NotInitialized
+    private var crashReporterStatus: CrashReporterStatus = CrashReporterStatus(NotInitialized)
 
     /**
      * Returns a handle to the underlying logger instance, if Capture has been started.
@@ -104,7 +105,7 @@ object Capture {
         @ExperimentalBitdriftApi
         fun initCrashReporting() {
             val crashReporter = CrashReporter()
-            crashReporterState = crashReporter.processCrashReportFile()
+            crashReporterStatus = crashReporter.processCrashReportFile()
         }
 
         /**
@@ -180,7 +181,7 @@ object Capture {
                             configuration = configuration,
                             sessionStrategy = sessionStrategy,
                             bridge = bridge,
-                            crashReporterState = crashReporterState,
+                            crashReporterStatus = crashReporterStatus,
                         )
                     default.set(LoggerState.Started(logger))
                 } catch (e: Throwable) {
--- a/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/Capture.kt
+++ b/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/Capture.kt
@@ -100,7 +100,7 @@ object Capture {
         /**
          * Initializes crash reporting mechanism.
          *
-         * This should be called priorly to Capture.Logger.start()
+         * This should be called prior to Capture.Logger.start()
          */
         @ExperimentalBitdriftApi
         @JvmStatic
--- a/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/Capture.kt
+++ b/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/Capture.kt
@@ -20,8 +20,8 @@ import io.bitdrift.capture.providers.FieldProvider
 import io.bitdrift.capture.providers.SystemDateProvider
 import io.bitdrift.capture.providers.session.SessionStrategy
 import io.bitdrift.capture.reports.CrashReporter
-import io.bitdrift.capture.reports.CrashReporter.CrashReporterState.NotInitialized
-import io.bitdrift.capture.reports.CrashReporter.CrashReporterStatus
+import io.bitdrift.capture.reports.CrashReporterState
+import io.bitdrift.capture.reports.CrashReporterStatus
 import okhttp3.HttpUrl
 import java.util.concurrent.atomic.AtomicReference
 import kotlin.time.Duration
@@ -55,7 +55,7 @@ internal sealed class LoggerState {
  */
 object Capture {
     private val default: AtomicReference<LoggerState> = AtomicReference(LoggerState.NotStarted)
-    private var crashReporterStatus: CrashReporterStatus = CrashReporterStatus(NotInitialized)
+    private var crashReporterStatus: CrashReporterStatus = CrashReporterStatus(CrashReporterState.NotInitialized)
 
     /**
      * Returns a handle to the underlying logger instance, if Capture has been started.
@@ -106,7 +106,11 @@ object Capture {
         @JvmStatic
         fun initCrashReporting() {
             val crashReporter = CrashReporter()
-            crashReporterStatus = crashReporter.processCrashReportFile()
+            if (crashReporterStatus.state is CrashReporterState.NotInitialized) {
+                crashReporterStatus = crashReporter.processCrashReportFile()
+            } else {
+                Log.w("capture", "Crash reporting already being initialized")
+            }
         }
 
         /**
--- a/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/Capture.kt
+++ b/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/Capture.kt
@@ -12,12 +12,15 @@ import com.github.michaelbull.result.Err
 import io.bitdrift.capture.common.MainThreadHandler
 import io.bitdrift.capture.events.span.Span
 import io.bitdrift.capture.events.span.SpanResult
+import io.bitdrift.capture.experimental.ExperimentalBitdriftApi
 import io.bitdrift.capture.network.HttpRequestInfo
 import io.bitdrift.capture.network.HttpResponseInfo
 import io.bitdrift.capture.providers.DateProvider
 import io.bitdrift.capture.providers.FieldProvider
 import io.bitdrift.capture.providers.SystemDateProvider
 import io.bitdrift.capture.providers.session.SessionStrategy
+import io.bitdrift.capture.reports.CrashReporter
+import io.bitdrift.capture.reports.CrashReporter.CrashReporterState
 import okhttp3.HttpUrl
 import java.util.concurrent.atomic.AtomicReference
 import kotlin.time.Duration
@@ -51,6 +54,7 @@ internal sealed class LoggerState {
  */
 object Capture {
     private val default: AtomicReference<LoggerState> = AtomicReference(LoggerState.NotStarted)
+    private var crashReporterState: CrashReporterState = CrashReporterState.NotInitialized
 
     /**
      * Returns a handle to the underlying logger instance, if Capture has been started.
@@ -91,6 +95,17 @@ object Capture {
 
         // This is a lazy property to avoid the need to initialize the main thread handler unless needed here
         private val mainThreadHandler by lazy { MainThreadHandler() }
+
+        /**
+         * Initializes crash reporting mechanism. This should be the first call upon Application.onCreate()
+         */
+        @Synchronized
+        @JvmStatic
+        @ExperimentalBitdriftApi
+        fun initCrashReporting() {
+            val crashReporter = CrashReporter()
+            crashReporterState = crashReporter.processCrashReportFile()
+        }
 
         /**
          * Initializes the Capture SDK with the specified API key, providers, and configuration.
@@ -165,6 +180,7 @@ object Capture {
                             configuration = configuration,
                             sessionStrategy = sessionStrategy,
                             bridge = bridge,
+                            crashReporterState = crashReporterState,
                         )
                     default.set(LoggerState.Started(logger))
                 } catch (e: Throwable) {
--- a/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/Capture.kt
+++ b/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/Capture.kt
@@ -98,11 +98,12 @@ object Capture {
         private val mainThreadHandler by lazy { MainThreadHandler() }
 
         /**
-         * Initializes crash reporting mechanism. This should be the first call upon Application.onCreate()
+         * Initializes crash reporting mechanism.
+         *
+         * This should be called priorly to Capture.Logger.start()
          */
-        @Synchronized
-        @JvmStatic
         @ExperimentalBitdriftApi
+        @JvmStatic
         fun initCrashReporting() {
             val crashReporter = CrashReporter()
             crashReporterStatus = crashReporter.processCrashReportFile()
--- a/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/LoggerImpl.kt
+++ b/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/LoggerImpl.kt
@@ -50,7 +50,6 @@ import io.bitdrift.capture.providers.MetadataProvider
 import io.bitdrift.capture.providers.session.SessionStrategy
 import io.bitdrift.capture.providers.toFields
 import io.bitdrift.capture.reports.CrashReporter.Companion.buildFieldsMap
-import io.bitdrift.capture.reports.CrashReporter.CrashReporterState.NotInitialized
 import io.bitdrift.capture.reports.CrashReporter.CrashReporterStatus
 import io.bitdrift.capture.threading.CaptureDispatchers
 import okhttp3.HttpUrl
@@ -85,8 +84,7 @@ internal class LoggerImpl(
     private val bridge: IBridge = CaptureJniLibrary,
     private val eventListenerDispatcher: CaptureDispatchers.CommonBackground = CaptureDispatchers.CommonBackground,
     windowManager: IWindowManager = WindowManager(errorHandler),
-    // TODO(FranAguilera): To remove default and update tests
-    private val crashReporterStatus: CrashReporterStatus = CrashReporterStatus(NotInitialized),
+    private val crashReporterStatus: CrashReporterStatus,
 ) : ILogger {
     private val metadataProvider: MetadataProvider
     private val memoryMetricsProvider = MemoryMetricsProvider(context)
@@ -271,11 +269,7 @@ internal class LoggerImpl(
                 CaptureJniLibrary.startLogger(this.loggerId)
             }
 
-        CaptureJniLibrary.writeSDKStartLog(
-            this.loggerId,
-            sdkLogStartMap(),
-            duration.toDouble(DurationUnit.SECONDS),
-        )
+        duration.writeSDKStartLog()
     }
 
     override val sessionId: String
@@ -535,11 +529,18 @@ internal class LoggerImpl(
         }
     }
 
-    private fun sdkLogStartMap(): Map<String, FieldValue> {
-        if (!runtime.isEnabled(RuntimeFeature.APPEND_INIT_CRASH_REPORTING_INFO)) {
-            return mapOf()
-        }
-        return crashReporterStatus.buildFieldsMap()
+    private fun Duration.writeSDKStartLog() {
+        val fields =
+            if (runtime.isEnabled(RuntimeFeature.APPEND_INIT_CRASH_REPORTING_INFO)) {
+                crashReporterStatus.buildFieldsMap()
+            } else {
+                mapOf()
+            }
+        CaptureJniLibrary.writeSDKStartLog(
+            loggerId,
+            fields,
+            toDouble(DurationUnit.SECONDS),
+        )
     }
 }
 
--- a/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/LoggerImpl.kt
+++ b/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/LoggerImpl.kt
@@ -526,14 +526,6 @@ internal class LoggerImpl(
             errorHandler.handleError("Couldn't start JankStatsMonitor", IllegalArgumentException("Invalid application provided"))
         }
     }
-
-    private fun writeSDKStartLog(duration: Duration) {
-        CaptureJniLibrary.writeSDKStartLog(
-            loggerId,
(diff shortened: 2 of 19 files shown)


platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/LoggerImpl.kt
```diff
@@ -82,6 +85,8 @@ internal class LoggerImpl(
     private val bridge: IBridge = CaptureJniLibrary,
     private val eventListenerDispatcher: CaptureDispatchers.CommonBackground = CaptureDispatchers.CommonBackground,
     windowManager: IWindowManager = WindowManager(errorHandler),
+    // TODO(FranAguilera): To remove default and update tests
+    private val crashReporterStatus: CrashReporterStatus = CrashReporterStatus(NotInitialized),
```
FranAguilera, line 89: Pending to remove default state and update all LoggerImpl test deps

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/CrashReporter.kt
```diff
@@ -0,0 +1,233 @@
...
+        val rootPath: String,
+        val extensionFileName: String,
+    )
+
+    internal companion object {
+        // TODO(FranAguilera): To rename to /bitdrift_capture/reports/config when shared-core is bumped
+        private const val CONFIGURATION_FILE_PATH = "/bitdrift_capture/reports/directories"
+        private const val DESTINATION_FILE_PATH = "/bitdrift_capture/reports/new"
+        private const val FILE_CREATION_TIME_ATTRIBUTE = "creationTime"
+        private const val CRASH_REPORTING_STATE_KEY = "crash_reporting_state"
```
murki, line 219: for fields that the sdk append internally we use the convention of prefixing them with `_`, e.g. `_crash_reporting_state`

FranAguilera: oh good catch, let me update

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/CrashReporter.kt
```diff
@@ -0,0 +1,233 @@
...
+
+        fun CrashReporterStatus.buildFieldsMap(): Map<String, FieldValue> =
+            buildMap {
+                put(CRASH_REPORTING_STATE_KEY, state.readableType.toFieldValue())
+                put(CRASH_REPORTING_DETAILS_KEY, state.message.toFieldValue())
+                put(CRASH_REPORTING_DURATION_NANO_KEY, getDurationFieldValue().toFieldValue())
+            }
+
+        @VisibleForTesting
+        fun CrashReporterStatus.getDurationFieldValue(): String = duration?.toString(DurationUnit.NANOSECONDS) ?: "n/a"
```
murki, line 231: it's weird to append this method's name with "FieldValue()" if it doesn't return an actual FieldValue, also why is this String and not Long? I noticed in your example session that we're appending "ns" at the end of the number. if it's not always numeric it will make it very hard to compute histograms etc

murki: Also the more I think about this, if we're exposing it to the customer probably ms will be best, (it's less scary than seeing huge numbers, plus our UX supports it already everywhere)

FranAguilera: good catch, this was unintended due to prior refactor. Will update asap

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/CrashReporter.kt
```diff
@@ -0,0 +1,233 @@
...
+import kotlin.time.DurationUnit
+import kotlin.time.measureTime
+
+/**
+ * Handles internal reporting of crashes
+ */
+internal class CrashReporter(
+    private val mainThreadHandler: MainThreadHandler = MainThreadHandler(),
+) {
+    private val appContext by lazy { APP_CONTEXT }
```
murki, line 31: is this to avoid potential race conditions? or why lazy

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/LoggerImpl.kt
```diff
@@ -529,6 +528,20 @@ internal class LoggerImpl(
             errorHandler.handleError("Couldn't start JankStatsMonitor", IllegalArgumentException("Invalid application provided"))
         }
     }
+
+    private fun Duration.writeSDKStartLog() {
```
murki, line 532: it looks awkward to me for this logic to be an extension method of Duration? wouldn't a simple plain method suffice? not everything has to be an extension method 😄

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/Capture.kt
```diff
@@ -92,6 +97,18 @@ object Capture {
...
 
+        /**
+         * Initializes crash reporting mechanism.
+         *
+         * This should be called priorly to Capture.Logger.start()
+         */
+        @ExperimentalBitdriftApi
+        @JvmStatic
+        fun initCrashReporting() {
+            val crashReporter = CrashReporter()
```
murki, line 108: what's the advantage this being an instance class if there's not state that lingers around? otherwise it could just be an static call (unless you'd want to mock it for tests which would mean you'd need to pass this instance at construction time)

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/LoggerImpl.kt
```diff
@@ -529,6 +528,20 @@ internal class LoggerImpl(
...
+        val fields =
+            if (runtime.isEnabled(RuntimeFeature.APPEND_INIT_CRASH_REPORTING_INFO)) {
+                crashReporterStatus.buildFieldsMap()
```
murki, line 535: you could also move the responsibility of checking the runtime flag inside crashReporter (you could pass the runtime object if you don't want to keep a `CrashReporter` instance class around)

murki: this is a nit btw

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/experimental/ExperimentalBitdriftApi.kt
```diff
@@ -0,0 +1,19 @@
...
+@RequiresOptIn(
+    level = RequiresOptIn.Level.WARNING,
+    message = "This API is in experimental phase and may change in the future.",
+)
+@Retention(AnnotationRetention.BINARY)
+@Target(AnnotationTarget.CLASS, AnnotationTarget.FUNCTION, AnnotationTarget.PROPERTY)
+/**
+ * Use it for APIs that are still under experimental phase
+ */
+annotation class ExperimentalBitdriftApi
```
murki, line 19: 👍

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/CrashReporter.kt
```diff
@@ -0,0 +1,233 @@
...
+
+    /**
+     * Process existing crash report files.
+     *
+     * @return The status of the crash reporting processing
+     */
+    fun processCrashReportFile(): CrashReporterStatus {
+        var crashReporterState: CrashReporterState = CrashReporterState.Initializing
```
murki, line 39: if this write is not inside the mainThread block it will most likely require synchronization. as-is there is not guarantees there won't be race conditions if `Logger.initCrashReporting()` gets called multiple times from diff threads

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/CrashReporter.kt
```diff
@@ -0,0 +1,233 @@
...
+            measureTime {
+                if (MainThreadHandler.isOnMainThread()) {
+                    crashReporterState = verifyDirectoriesAndCopyFiles()
+                } else {
+                    mainThreadHandler.run {
+                        crashReporterState = verifyDirectoriesAndCopyFiles()
+                    }
+                }
+            }
+        return CrashReporterStatus(crashReporterState, duration)
```
murki, line 50: shouldn't we just do everything inside the main thread? returning this value in the context of the original call could lead to race conditions. Since `CaptureJniLibrary.writeSDKStartLog()` should already be handled in a background thread we could avoid this extra context switching

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/Capture.kt
FranAguilera: actually now that I think about it, making instance class here will also prevent easier access to `processCrashReportFile` which should be only accessed here.

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/CrashReporter.kt
```diff
@@ -0,0 +1,233 @@
...
+
+    private data class ConfigDetails(
+        val rootPath: String,
+        val extensionFileName: String,
+    )
+
+    internal companion object {
+        // TODO(FranAguilera): To rename to /bitdrift_capture/reports/config when shared-core is bumped
+        private const val CONFIGURATION_FILE_PATH = "/bitdrift_capture/reports/directories"
+        private const val DESTINATION_FILE_PATH = "/bitdrift_capture/reports/new"
```
murki, line 217: we shouldn't assume the root sdk directory path will never change. this logic already lives in `LoggerImpl.getSdkDirectoryPath()` we should move it to a place where it can remain the single source of truth

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/LoggerImpl.kt
FranAguilera: this would require runtime dependency at CrashReporter which is not available until logger impl. Also feel like this is a better place to leave given LoggerImpl is the one calling writeSDKStartLog

murki: ah good call

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/Capture.kt
```diff
@@ -92,6 +97,18 @@ object Capture {
...
+            crashReporterStatus = crashReporter.processCrashReportFile()
```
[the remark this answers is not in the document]
FranAguilera: Yup, will use the stored state to avoid a new call to processCrashReportFile

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/CrashReporter.kt
FranAguilera: good call, will update

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/CrashReporter.kt
FranAguilera: that's good to know, will update

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/CrashReporter.kt
```diff
@@ -0,0 +1,235 @@
...
+        } else {
+            Initialized.WithoutPriorCrash("No prior crashes found")
+        }
+    }
+
+    private fun getConfigDetails(crashConfigFileContent: String): ConfigDetails? =
+        runCatching {
+            val crashConfigDetails = crashConfigFileContent.split(",")
+            ConfigDetails(crashConfigDetails[0], crashConfigDetails[1])
+        }.getOrNull()
```
murki, line 112: wouldn't you prefer to return an Init Error here as well

murki: Ideally we could pass the string of the error back up

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/CrashReporter.kt
```diff
@@ -0,0 +1,235 @@
...
+    private fun getFileCreationTimeEpochInMillis(file: File): Long =
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+            val fileTime =
+                Files.getAttribute(file.toPath(), FILE_CREATION_TIME_ATTRIBUTE) as FileTime
+            fileTime.toMillis()
+        } else {
+            file.lastModified()
+        }
+
+    internal sealed class CrashReporterState(
```
murki, line 136: nit: could move this class to its own file for clarity

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/CrashReporter.kt
```diff
@@ -0,0 +1,235 @@
...
+        return fileCreationEpochTime.toString() + "." + this.extension
+    }
+
+    private fun getFileCreationTimeEpochInMillis(file: File): Long =
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+            val fileTime =
+                Files.getAttribute(file.toPath(), FILE_CREATION_TIME_ATTRIBUTE) as FileTime
+            fileTime.toMillis()
+        } else {
+            file.lastModified()
```
murki, line 133: Shouldn't we choose either creation time or last modified time? Why are we switching between timestamp origin?

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/CrashReporter.kt
```diff
@@ -0,0 +1,235 @@
...
+        }
+
+    internal sealed class CrashReporterState(
+        open val readableType: String,
+    ) {
+        /**
+         * Contains details of the state
+         */
+        abstract val message: String
```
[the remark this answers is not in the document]
FranAguilera: yup, that makes sense. This went through many refactors and now there is no reason to keep it

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/CrashReporter.kt
```diff
@@ -0,0 +1,235 @@
...
+        sourceFile: File,
+        fileExtension: String,
+    ): File? =
+        sourceFile.walk().firstOrNull {
+            it.isFile && it.extension == fileExtension
+        }
+
+    private fun File.toFilenameWithTimeStamp(): String {
+        val fileCreationEpochTime = getFileCreationTimeEpochInMillis(this)
+        return fileCreationEpochTime.toString() + "." + this.extension
```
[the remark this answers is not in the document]
FranAguilera: i though it would be cleaner for backend to process but i can keep the original file name if you prefer, lmk

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/CrashReporter.kt
```diff
@@ -0,0 +1,235 @@
...
+        fileExtension: String,
+    ): CrashReporterState {
+        val crashFile =
+            findCrashFile(sourceDirectory, fileExtension)
+                ?: let {
+                    return Initialized.WithoutPriorCrash("Crash file not found in the source directory")
+                }
+
+        val destinationFile = File(destinationDirectory, crashFile.toFilenameWithTimeStamp())
+        crashFile.copyTo(destinationFile, overwrite = true)
```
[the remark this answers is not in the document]
FranAguilera: oh yeah, this was before the timestamp. Will make sure to delete any prior files on that directory

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/CrashReporter.kt
FranAguilera: I will tackle that as a follow up task https://linear.app/bitdrift/issue/BIT-4837

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/CrashReporter.kt
FranAguilera: Added a check now. Lmk what you think

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/CrashReporter.kt
FranAguilera: good catch, updated to last modified

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/CrashReporter.kt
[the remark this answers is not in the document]
FranAguilera: added the timestamp to the begining of file again

platform/jvm/common/src/main/kotlin/io/bitdrift/capture/common/MainThreadHandler.kt
```diff
@@ -25,4 +26,35 @@ class MainThreadHandler {
     fun run(run: () -> Unit) {
         mainHandler.post { run() }
     }
+
+    /**
+     * Runs the specific action on main thread and return its results.
+     */
+    fun <T> runAndReturnResult(action: () -> T): T {
```
FranAguilera, line 33: Not sure happy with this but we can discuss

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/CrashReporter.kt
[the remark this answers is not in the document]
FranAguilera: forgot this, will update later on

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/CrashReporter.kt
murki: let's do it in this one, i want to avoid the case where we release it and if we ever were to want to change the sdk_root in the future we would need to remember that there is the on sdk version where it owuld break crash reporting

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/CrashReporter.kt
murki: disregard this if we choose to remove the `message` in the status

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/CrashReporter.kt
```diff
@@ -0,0 +1,170 @@
...
+    private val appContext = APP_CONTEXT
+
+    /**
+     * Process existing crash report files.
+     *
+     * @return The status of the crash reporting processing
+     */
+    fun processCrashReportFile(): CrashReporterStatus =
+        runCatching {
+            if (MainThreadHandler.isOnMainThread()) {
```
murki, line 39: nit: shouldn't this method be inside the instance of `mainThreadHandler` that way it could be easily mocked

platform/jvm/common/src/main/kotlin/io/bitdrift/capture/common/MainThreadHandler.kt
murki: I see what you mean, but seems like a useful addition (as long as it works as expected)

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/CrashReporter.kt
```diff
@@ -0,0 +1,170 @@
...
+            }
+
+        val sourceDirectory =
+            File(appContext.cacheDir.absolutePath, crashConfigDetails.rootPath)
+        if (!sourceDirectory.exists() || !sourceDirectory.isDirectory) {
+            return CrashReporterState.Initialized.WithoutPriorCrash
+        }
+
+        val destinationDirectory =
+            File("$filesDir$DESTINATION_FILE_PATH").apply { if (!exists()) mkdirs() }
```
murki, line 82: manual string concat alert

murki (approved):
lgtm barring the hardcoding of `CONFIGURATION_FILE_PATH`

also would love if @kattrali can take a quick pass before merging

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/LoggerImpl.kt
```diff
@@ -529,6 +528,20 @@ internal class LoggerImpl(
             errorHandler.handleError("Couldn't start JankStatsMonitor", IllegalArgumentException("Invalid application provided"))
         }
     }
+
+    private fun writeSDKStartLog(duration: Duration) {
+        val fields: Map<String, FieldValue> =
+            if (runtime.isEnabled(RuntimeFeature.APPEND_INIT_CRASH_REPORTING_INFO)) {
```
snowp, line 534: did you make sure that this runtime flag correctly get loaded early enough? I think this might be racing with the async runtime loading that happens in response to startLogger

FranAguilera: good point, I was testing with default value but will check with staging update

murki: also I personally don't think the adding of data to be a risky operation that requires to be gated

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/Capture.kt
```diff
@@ -92,6 +97,22 @@ object Capture {
         // This is a lazy property to avoid the need to initialize the main thread handler unless needed here
         private val mainThreadHandler by lazy { MainThreadHandler() }
 
+        /**
+         * Initializes crash reporting mechanism.
+         *
+         * This should be called priorly to Capture.Logger.start()
```
murki, line 103:
```suggestion
         * This should be called prior to Capture.Logger.start()
```

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/LoggerImpl.kt
FranAguilera: that's true, it's a safe change though I thought we wanted to have control over this data for different customers. I can remove the gating, please lmk

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/CrashReporter.kt
FranAguilera: sounds good, will update before landing

kattrali (approved):

PR #256, opened 2025-03-13
Fix capture pom

FranAguilera:
See related PR https://github.com/google/bazel-common/pull/217

Big kudos to @quanturium for the investigation and root cause 🙏

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -78,7 +78,7 @@ new_git_repository(
 
 git_override(
     module_name = "google_bazel_common",
-    commit = "a348dcc487d9bea3d527be3b591a4170338257e6",
+    commit = "05f53f68ecec6aec44b54751306af0e7b6552682",
     remote = "https://github.com/google/bazel-common",
 )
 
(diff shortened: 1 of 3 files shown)


bazel/pom_template.xml
```diff
@@ -6,7 +6,7 @@
     <groupId>io.bitdrift</groupId>
     <artifactId>capture</artifactId>
     <version>{pom_version}</version>
-    <packaging>aar</packaging>
+    <type>aar</type>
```
murki, line 9: I don't think this is the right change. IIUC at the project level packaging is correct, it's inside the dependencies collection where it should be type.

FranAguilera: oh that's right, will continue exploring

murki (approved):
did the second change get in by mistake?

FranAguilera:
> did the second change get in by mistake?

It was to ensure build was triggered and that everything gets green

MODULE.bazel
```diff
@@ -78,7 +78,7 @@ new_git_repository(
 
 git_override(
     module_name = "google_bazel_common",
-    commit = "a348dcc487d9bea3d527be3b591a4170338257e6",
+    commit = "05f53f68ecec6aec44b54751306af0e7b6552682",
```
FranAguilera, line 81: https://github.com/google/bazel-common/pull/217/commits/05f53f68ecec6aec44b54751306af0e7b6552682

PR #258, opened 2025-03-14
[Android] [JankStats] Add error reporting for cases where JankStats API returns unexpected duration

FranAguilera:
Looks like there are reports with an overflown ms duration (more details on BIT-4813), on this PR the goal is to check for any potential incorrect duration on JankStats and log an error if that's the case


platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/events/performance/JankStatsMonitor.kt
```diff
@@ -278,7 +288,7 @@ internal class JankStatsMonitor(
     )
 
     private companion object {
-        private const val TO_MILLI = 1_000_000L
+        private const val ANR_DURATION_MAX_THRESHOLD_MS = 1_000_000L
```
FranAguilera, line 291: This is a very unlikely high number for an ANR so would help us determine if there was an overflow on incorrect reporting on JankStats

PR #261, opened 2025-03-17
Bump shared core

snowp:
This brings in a number of changes relating to crash handling and some changes to the log fields interface that required a few changes 

This seems to increase the binary size by a bit so we should decide whether we want to block this on figuring out how to claw back some of the binary size loss


mattklein123 (approved):
LGTM but we should probably do a bunch of extra manual testing on this diff before we release it. Will def to others on the question about binary size.

PR #263, opened 2025-03-18
Send AppExitInfo as fatal information

FranAguilera:
Relying on valid stack trace info from ApplicationExitInfo to send fatal information.

On this PR, we are doing basic pre processing for REASON_ANR, but a separate PR will take care of the processing for REASON_CRASH_NATIVE


platform/jvm/capture/src/test/kotlin/io/bitdrift/capture/AppExitLoggerTest.kt
```diff
@@ -299,6 +268,58 @@ class AppExitLoggerTest {
...
+            }.toFields()
+        verify(logger).log(
+            eq(LogType.LIFECYCLE),
+            eq(LogLevel.ERROR),
+            eq(expectedFields),
+            eq(null),
+            eq(LogAttributesOverrides(sessionId, timestamp)),
+            eq(false),
+            argThat { i: () -> String -> i.invoke() == "AppExit" },
+        )
```
snowp, line 313: One thing we need to figure out with this handling is how we can communicate between the platform and rust side to determine who is responsible for logging the log line - right now if we detect a file in the report inbox we treat it as a crash and log a crash line. In this case this would lead to double logging, so we probably need to make that better

FranAguilera: I agree, let's discuss offline

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/FatalIssueReporter.kt
```diff
@@ -7,46 +7,139 @@
...
+    @RequiresApi(Build.VERSION_CODES.R)
+    override fun processAndStoreAppExitInfoTrace(
+        errorHandler: ErrorHandler,
+        applicationExitReason: ApplicationExitInfo,
+    ) {
+        val traceInputStream = applicationExitReason.traceInputStream ?: return
+        val processedResult =
+            when (applicationExitReason.reason) {
+                ApplicationExitInfo.REASON_ANR -> AnrStackTraceProcessor.process(traceInputStream)
+                ApplicationExitInfo.REASON_CRASH_NATIVE -> NativeCrashStackTraceProcessor.process(traceInputStream)
```
murki, line 70: Do you need to check API >= 31 for this branch?
>Beginning with API 31, tombstone traces will be returned for [REASON_CRASH_NATIVE](https://developer.android.com/reference/android/app/ApplicationExitInfo#REASON_CRASH_NATIVE), with an InputStream containing a protobuf with [this schema](https://android.googlesource.com/platform/system/core/+/refs/heads/master/debuggerd/proto/tombstone.proto).

FranAguilera:
Will open a new PR

PR #264, opened 2025-03-18
[ios] upload report files from configured directory

kattrali:
iOS companion to #254

* reads configuration for source directory, copies matching report to destination, prepending the modification time to the filename
* [sample session](https://timeline.bitdrift.dev/session/3254afb3-af55-4511-bc8c-2b041f1f76c2)


kattrali:

platform/swift/source/Capture.swift
```diff
@@ -88,6 +88,17 @@ extension Logger {
...
     }
 
+    /// Initializes the crash reporting mechanism. Must be called prior to `Logger.start()`
+    /// This API is experimental and subject to change
+    public static func initFatalIssueReporting() {
+        if issueReporterInitResult.0 == .notInitialized {
+            issueReporterInitResult = (.initializing, 0)
+            issueReporterInitResult = FatalIssueReporter.processFiles()
+        } else {
+            log(level: .warning, message: "Crash reporting already being initialized")
```
FranAguilera, line 98: optional nit: Replace message with "Fatal Issue reporting"

platform/swift/source/Logger.swift
```diff
@@ -194,7 +196,11 @@ public final class Logger {
 
         defer {
             let duration = timeProvider.timeIntervalSince(start)
-            self.underlyingLogger.logSDKStart(fields: [:], duration: duration)
+            let fields: Fields = [
+                "_crash_reporting_state": "\(Logger.issueReporterInitResult.0)",
```
FranAguilera, line 200: nit: To rename this to _fatal_issue_reporting_state

FranAguilera (approved):

platform/swift/source/FatalIssueReporter.swift
```diff
@@ -0,0 +1,180 @@
...
+    case processingFailure(String)
+}
+
+typealias IssueReporterInitResult = (IssueReporterInitState, TimeInterval)
+
+func measureTime<T>(operation: () -> T) -> (T, TimeInterval) {
+    let start = DispatchTime.now()
+    let result = operation()
+    let end = DispatchTime.now()
+    let duration = Double(end.uptimeNanoseconds - start.uptimeNanoseconds) / Double(NSEC_PER_SEC)
```
murki, line 32: shouldn't this be a ms value? since it's reported in a field called `_fatal_issue_reporting_duration_ms`

platform/swift/source/FatalIssueReporter.swift
```diff
@@ -0,0 +1,180 @@
...
+
+    /// Create upload filename for a given URL
+    ///
+    /// - parameter entry: a file URL
+    ///
+    /// - returns: a new file name for `entry`, prepending the modification date as a Unix timestamp
+    private static func getUploadFilename(_ entry: URL) -> String {
+        if let properties = try? entry.resourceValues(forKeys: [.contentModificationDateKey]),
+           let modDate = properties.contentModificationDate {
+            let modDateMillis = UInt64(modDate.timeIntervalSince1970 * 1000)
```
murki, line 135: does this mean we actually don't have the precision of ms and we're just adding 0s?

murki: @snowp just clarified that this is actually returning fractions originally so we should be fine

platform/swift/source/FatalIssueReporter.swift
kattrali: It should, nice catch! The change'll be at the other end where its serialized, since `TimeInterval` is a measurement in (fractional) seconds

murki (approved):
maybe in this same pr we should male the android change to make it private and merge both at the same time so it can go out in the next release with the other fixes (i assume this is ready?)

PR #265, opened 2025-03-19
[Android] Extracting last exit reason logic

FranAguilera:
Extract the logic for getting the last ApplicationExitReason so it can be reused

### DIFF
--- a/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/exitinfo/LatestAppExitInfoProviderProvider.kt
+++ b/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/exitinfo/LatestAppExitInfoProviderProvider.kt
@@ -4,7 +4,7 @@
 // Use of this source code is governed by a source available license that can be found in the
 // LICENSE file or at:
 // https://polyformproject.org/wp-content/uploads/2020/06/PolyForm-Shield-1.0.0.txt
-package io.bitdrift.capture.reports.lastexitinfo
+package io.bitdrift.capture.reports.exitinfo
 
 import android.annotation.TargetApi
 import android.app.ActivityManager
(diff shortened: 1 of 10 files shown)


platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/lastexitinfo/ILastExitInfo.kt
```diff
@@ -0,0 +1,23 @@
...
+package io.bitdrift.capture.reports.lastexitinfo
+
+import android.annotation.TargetApi
+import android.app.ApplicationExitInfo
+import android.os.Build
+
+/**
+ * Retrieves the latest [ApplicationExitInfo] if available
+ */
+fun interface ILastExitInfo {
```
murki, line 17: maybe rename to something that implies this _retrieves_ or _manages_ the exit info, the way it is it sounds like this is a wrapper for the exit info itself

FranAguilera: good point, I think I had the "Retrieve" keyword at some point but will add it back

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/lastexitinfo/LastExitInfo.kt
```diff
@@ -0,0 +1,35 @@
...
+import android.annotation.TargetApi
+import android.app.ActivityManager
+import android.app.ApplicationExitInfo
+import android.os.Build
+import io.bitdrift.capture.common.ErrorHandler
+
+/**
+ * Concrete impl of [ILastExitInfo]
+ */
+internal class LastExitInfo(
```
murki, line 18: nit: seems like this class would should also  be responsible of things like `ApplicationExitInfo.getNativeCrashByteArray()` `buildAppExitInternalFieldsMap()` etc? for better encapsulation

murki (approved):

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/lastexitinfo/LastExitInfo.kt
FranAguilera: yup, will tackle in separate PR

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/exitinfo/LatestAppExitInfoProviderProvider.kt
```diff
@@ -13,12 +13,12 @@ import android.os.Build
 import io.bitdrift.capture.common.ErrorHandler
 
 /**
- * Concrete impl of [ILastExitInfo]
+ * Concrete impl of [ILatestAppExitInfoProvider]
  */
-internal class LastExitInfo(
+internal class LatestAppExitInfoProviderProvider(
```
murki, line 18: ProviderProvider? even Java people wouldn't approve 😆

PR #270, opened 2025-03-20
add script to point shared-core deps to local repo

snowp:
This should make it easier to iterate on shared-core changes when using the gradle test app which works with the path overrides


murki (approved):
yes pls

PR #272, opened 2025-03-20
Add start/end + parent span ID to spans interface

Reflejo:


platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/events/span/Span.kt
```diff
@@ -77,16 +84,24 @@ class Span internal constructor(
     fun end(
         result: SpanResult,
         fields: Map<String, String>? = null,
+        endTimeInMs: Long? = null,
     ) {
         logger?.apply {
+            val durationMs: Long =
+                if (endTimeInMs != null && startTimeInMs != null) {
+                    endTimeInMs - startTimeInMs
```
murki, line 92: so this means that for the single-shot api `logSpan()` both start and stop logs would appear back to back in the timeline even though technically the start should've happened at `startTimeInMs`, are we ok with that?

murki: also since they are passing `endTimeInMs` for the end this would mean that the timestamp won't be accurate either

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/Capture.kt
```diff
@@ -419,15 +420,50 @@ object Capture {
          *
          * @param name the name of the operation.
          * @param level the severity of the log.
-         * @fields additional fields to include in the log.
+         * @param fields additional fields to include in the log.
+         * @param startTimeInMs an optional custom start time to use in combination with an `endTimeInMs`
```
murki, line 424: /nit: is not common convention to include the word "In" (at least in our codebase) we usually just append the unit at the end of the variable, e.g. `startTimeMs` (similar to [c++ conventions](https://accu.org/journals/overload/24/136/mertz_2318/))

murki: also it would be good to clarify if this is means unix epoch time or something else

platform/swift/source/logging/Span.swift
```diff
@@ -27,16 +27,21 @@ public final class Span {
         case end = "end"
     }
 
-    private let timeProvider: TimeProvider
+    /// This is the autogenerated unique identifier of the span
+    public let id = UUID()
```
murki, line 31: One of the asks was to be able to end a span using solely this ID (I supposed that's why we exposed this?) since they didn't want to rely on having to pass full span objects around but I don't see this functionality added anywhere. I think we should consider adding a `endSpan()` method that just takes this ID (yes, that would mean we would need to persist some state around in-flight spans ourselves)

murki: also, this is not exposed on the Android side (not sure if on purpose)

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/events/span/Span.kt
```diff
@@ -55,14 +59,17 @@ class Span internal constructor(
             put(SpanField.Key.ID, id.toString())
             put(SpanField.Key.NAME, name)
             put(SpanField.Key.TYPE, SpanField.Value.TYPE_START)
+            put(SpanField.Key.PARENT, parentSpanId.toString())
```
murki, line 62: should we only add the field if not null?

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/ILogger.kt
```diff
@@ -115,14 +117,51 @@ interface ILogger {
...
+     * @param parentSpanId an optional ID of the parent span, used to build span hierarchies. A span
+     *                     without a parentSpanID is considered a root span.
+     */
+    fun logSpan(
+        name: String,
+        level: LogLevel,
+        fields: Map<String, String>? = null,
+        result: SpanResult,
+        startTimeInMs: Long,
+        endTimeInMs: Long,
```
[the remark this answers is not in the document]
Reflejo: I wouldn't boil the ocean, you can always pass 0 on start and duration in end :)

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/events/span/Span.kt
Reflejo: Sorry I didn't understand the first point? In the one-shot we don't send start

murki: Ah I guess I missed the `emitStartLog` part

PR #273, opened 2025-03-21
[Android] Preparing FatalIssueReporting for release

FranAguilera:


platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/FatalIssueReporter.kt
```diff
@@ -120,16 +120,17 @@ internal class FatalIssueReporter(
...
     private fun findCrashFile(
         sourceFile: File,
         fileExtension: String,
     ): File? =
-        sourceFile.walk().firstOrNull {
-            it.isFile && it.extension == fileExtension
-        }
+        sourceFile.walk()
+            .filter { it.isFile && it.extension == fileExtension }
+            .maxByOrNull { it.lastModified() }
```
murki, line 134: 👍 
is there no way to write a test for this?

murki (approved):

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/FatalIssueReporter.kt
FranAguilera: yup, it's wip but was putting it out for snow to test first

PR #286, opened 2025-03-28
Adding protocol to HTTPResponseInfo

FranAguilera:
Resolves BIT-4952

### Android
image

### iOS
image

### DIFF
--- a/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/network/okhttp/CaptureOkHttpEventListener.kt
+++ b/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/network/okhttp/CaptureOkHttpEventListener.kt
@@ -448,6 +448,7 @@ internal class CaptureOkHttpEventListener internal constructor(
             tcpDurationMs = tcpDurationMs,
             fetchInitializationMs = fetchInitializationMs,
             responseLatencyMs = responseLatencyMs,
+            protocolName = lastResponse?.protocol.toString(),
         )
 
     private fun isInterruptedException(e: Throwable): Boolean {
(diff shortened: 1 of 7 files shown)


platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/network/okhttp/CaptureOkHttpEventListener.kt
```diff
@@ -448,6 +448,7 @@ internal class CaptureOkHttpEventListener internal constructor(
             tcpDurationMs = tcpDurationMs,
             fetchInitializationMs = fetchInitializationMs,
             responseLatencyMs = responseLatencyMs,
+            protocolName = lastResponse?.protocol.toString(),
```
FranAguilera, line 451: https://square.github.io/okhttp/5.x/okhttp/okhttp3/-protocol/index.html#-1745682694%2FFunctions%2F1481842020

platform/swift/source/logging/HTTPRequestMetrics.swift
```diff
@@ -97,7 +103,8 @@ extension HTTPRequestMetrics {
                 tlsDuration: Self.getTLSDuration(metrics: metrics),
                 tcpDuration: Self.getTCPDuration(metrics: metrics),
                 fetchInitializationDuration: Self.getInitializationDuration(metrics: metrics),
-                responseLatency: Self.getResponseLatency(metrics: metrics)
+                responseLatency: Self.getResponseLatency(metrics: metrics),
+                protocolName: Self.getProtocolName(metrics: metrics)
```
Reflejo, line 107: I'd just inline metrics.transactionMetrics.last?.protocolName here compact map will map the whole array which you don't really need if you are taking the last one

FranAguilera: 👍 will update

PR #290, opened 2025-04-03
[Android] Remove send_crash_artifact flag

FranAguilera:
This was used for internal testing and can be removed now


FranAguilera:
> I wonder if we should only remove this once we have an alternative way of sending this info, in case we need data for testing in the interim

discussed offline, if we need this again we can patch locally for debug builds

PR #295, opened 2025-04-07
Add python library for the public API

Reflejo:
For now we'll only support the ingest API


mattklein123 (approved):
Should we put this directly in the api repo maybe? Or some new repo? Not sure here is the best place.

PR #296, opened 2025-04-08
[Android] Initial prototype for adding built in reporting

FranAguilera:
## What

Allows to configure the fatal issue reporting mechanism, user should be able to initialize it like below:

- `Capture.Logger.initFatalIssueReporting(FatalIssueMechanism.BUILT_IN)` // Relies on detecting issues without any 3rd party tools

- `Capture.Logger.initFatalIssueReporting(FatalIssueMechanism.INTEGRATION)` // Relies on custom configuration for easier integration with existing reporting mechanism

_**NOTE: For now this will be merging on feature branch named `feat-init-built-in-reporting` as will keep iterating on this**_

## Verification

- Open gradle demo app, navigate to configuration and select "BUILT_IN" as fatal issue reporter mechanism
- Verified on [session](https://timeline.bitdrift.dev/session/f5e18027-4e38-47bb-bb6e-80d3ac6dde8c?utm_source=sdk&pages=1&utilization=0&expanded=-7266192926758369634) , which has a simulated crash between  `initFatalIssueReporting` and `Capture.Logger.start()` calls. Capture.start logs will indicate the reporting mechanism in this case as BUILT_IN, see [timeline entry](https://timeline.bitdrift.dev/session/f5e18027-4e38-47bb-bb6e-80d3ac6dde8c?utm_source=sdk&pages=1&utilization=0&expanded=-1006156035084377033)
- Verified that when `initFatalIssueReporting` is never called AppExitLogger events for JVM crash/AppExit ANR/Native remain intact 
- See generated files [here](https://docs.google.com/document/d/1rcegAuIY30gJiCaUgVEJLWuJ2ektPsD-FihohJ50KiU/edit?tab=t.t615fz5acx4u#heading=h.vtknwiymf72o)

Resolves BIT-5069


platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/FatalIssueMechanism.kt
```diff
@@ -0,0 +1,35 @@
...
+    /**
+     * The readable name that can be send via [InternalFieldsLog]
+     */
+    val displayName: String,
+) {
+    /**
+     * Use this option, to integrate with existing [FatalIssue] mechanism. This will scan for specific
+     * fatal issues on the configured directory
+     */
+    CUSTOM_CONFIG("CUSTOM_CONFIG"),
```
FranAguilera, line 24: def open to naming suggestions here

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/FatalIssueType.kt
```diff
@@ -0,0 +1,38 @@
...
+package io.bitdrift.capture.reports
+
+/**
+ * Specifies the different types of fatal issues that can occur within the application.
+ *
+ * Each type represents a specific kind of fatal issue that may trigger the reporting mechanism.
+ *
+ * @param displayName The readable name of the fatal issue type used for logging or reporting purposes.
+ */
+enum class FatalIssueType(
```
FranAguilera, line 17: maybe this could live at some point on a proto or something

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/jvmcrash/CaptureUncaughtExceptionHandler.kt
```diff
@@ -0,0 +1,89 @@
...
+        try {
+            crashing = true
+            if (FatalIssueMechanism.BUILT_IN == fatalIssueReporter?.fetchStatus()?.mechanism) {
+                fatalIssueReporter?.persistJvmCrash(
+                    timestamp = System.currentTimeMillis(),
+                    callerThread = thread,
+                    throwable = throwable,
+                )
+            } else {
+                crashLogger?.logCrash(thread, throwable)
```
FranAguilera, line 79: the idea here is that we won't duplicate multiple crash info on timeline, though this could break for existing customers relying on that app exit log crash

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/persistence/FatalIssueReporterStorage.kt
```diff
@@ -0,0 +1,34 @@
...
+    // TODO(FranAguilera): BIT-5083 Replace storing via Gson with native call
+    private val gson by lazy { Gson() }
+
+    override fun persistFatalIssue(
+        terminationTimeStampInMilli: Long,
+        fatalIssueType: FatalIssueType,
+        fatalIssueReport: FatalIssueReport,
+    ) {
+        // TODO(FranAguilera): BIT-5083 Replace storing via Gson with native call
+        val fileName = "${terminationTimeStampInMilli}_${fatalIssueType.displayName}.json"
```
FranAguilera, line 30: @kattrali Here we can update to native call when ready. Probably we can pass also the path from here (given that from platform we have easy access to cached_dir, files_dir, etc

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/FatalIssueReporterState.kt
```diff
@@ -24,7 +27,13 @@ internal sealed class FatalIssueReporterState(
     data object NotInitialized : FatalIssueReporterState("NOT_INITIALIZED")
 
     /**
-     * Sealed class representing all initialized states
+     * Represents the initialized state when [FatalIssueMechanism.BUILT_IN] is configured
```
murki, line 30: Is the `Initializing` state above used for anything?

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/FatalIssueMechanism.kt
murki: I would just call it "INTEGRATION" since we would only advise to use it of there are already using an existing crash handling library /namingishard

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/FatalIssueMechanism.kt
```diff
@@ -0,0 +1,35 @@
...
+
+package io.bitdrift.capture.reports
+
+/**
+ * Specifies the [FatalIssue] mechanism.
+ *
+ */
+enum class FatalIssueMechanism(
+    /**
+     * The readable name that can be send via [InternalFieldsLog]
```
murki, line 16: fyi all these kdocs symbols don't resolve in the current build 
Screenshot 2025-04-09 at 9 46 56 AM

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/events/lifecycle/AppExitLogger.kt
```diff
@@ -98,7 +98,7 @@ internal class AppExitLogger(
             return
         }
 
-        when (val lastExitInfoResult = latestAppExitInfoProvider.get(activityManager)) {
+        when (val lastExitInfoResult = fatalIssueReporter.getLatestAppExitInfoProvider().get(activityManager)) {
```
murki, line 101: nested classes like that are usually a code smell https://wiki.c2.com/?TrainWreck

murki: also it sounds like Activity Manager should be a dependency of one of the classes

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/FatalIssueReporterState.kt
FranAguilera: Good catch, it's not. Will remove

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/FatalIssueMechanism.kt
FranAguilera: ah that's good to know, will check

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/FatalIssueReporter.kt
```diff
@@ -28,31 +41,155 @@ import kotlin.time.measureTime
...
+                fatalIssueReporterStatus = processReportFilesFromCustomConfig()
+            } else if (fatalIssueMechanism == FatalIssueMechanism.BUILT_IN) {
+                fatalIssueReporterStatus = initializeBuiltInReporting()
+            }
+        } else {
+            Log.w("capture", "Fatal issue reporting already being initialized")
+        }
+    }
+
+    override fun fetchStatus(): FatalIssueReporterStatus = fatalIssueReporterStatus
```
murki, line 72: the use of the `fetch` verb here makes me think this is an I/O operation or something of the sort

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/events/lifecycle/AppExitLogger.kt
```diff
@@ -114,17 +114,23 @@ internal class AppExitLogger(
                     lastExitInfo.processStateSummary?.toString(StandardCharsets.UTF_8) ?: return
                 val timestampMs = lastExitInfo.timestamp
 
-                logger.log(
-                    LogType.LIFECYCLE,
-                    lastExitInfo.reason.toLogLevel(),
-                    buildAppExitInternalFieldsMap(lastExitInfo),
-                    attributesOverrides = LogAttributesOverrides.SessionID(sessionId, timestampMs),
-                ) { APP_EXIT_EVENT_NAME }
+                if (FatalIssueMechanism.BUILT_IN != fatalIssueReporter.fetchStatus().mechanism) {
```
murki, line 117: trainwreck here again, I would just expose a method like isBuiltIn() or something

FranAguilera: agree, this went through a serious of million refactors so will be cleaning this up soon

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/FatalIssueReport.kt
```diff
@@ -0,0 +1,26 @@
...
+ *
+ * TODO(FranAguilera): BIT-5070 Update to include full FatalIssueReport
+ */
+internal data class FatalIssueReport(
+    @SerializedName("issue_type")
+    val issueType: String,
+    val errors: List<Exception>,
+)
+
+internal data class Exception(
```
murki, line 23: let's make sure this is not stripped out with proguard (see our gson config section
:

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/FatalIssueReporter.kt
```diff
@@ -28,31 +41,155 @@ import kotlin.time.measureTime
...
+     *
+     * @return The status of the fatal issue reporting processing
+     */
+    private fun initializeBuiltInReporting(): FatalIssueReporterStatus {
+        val mechanism = FatalIssueMechanism.BUILT_IN
+        return runCatching {
+            mainThreadHandler.runAndReturnResult {
+                captureUncaughtExceptionHandler.installForBuiltInReporting(this)
+                val lastReasonResult = latestAppExitInfoProvider.get(activityManager)
+                if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.R &&
```
murki, line 163: don't you need to distinguish between `Build.VERSION_CODES.R` and `Build.VERSION_CODES.S` for ANR and Native crashes respectively?

murki: fyi there exists a `BuildVersionChecker` abstraction class in you ever want to unit test the different code paths

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/jvmcrash/CaptureUncaughtExceptionHandler.kt
```diff
@@ -0,0 +1,88 @@
...
+        this.crashLogger = crashLogger
+    }
+
+    /**
+     * Installs for built in reporting.
+     *
+     * NOTE: This will be only applicable when [Capture.Logger.initFatalIssueReporting(FatalIssueMechanism.BUILT_IN)]
+     * is previously called
+     */
+    override fun installForBuiltInReporting(fatalIssueReporter: IFatalIssueReporter) {
```
murki, line 40: The fact that you can initialize equivalent functionality of the same class on very different ways at very different times makes the code hard to follow and reason about (hence why you need a ifneeded check). I would try to break this apart via polymorphism, maybe use a single generic Interface<T> that has only one install and one uninstall method and they get instantiated differently based on whatever logic

murki: you can imagine both IFatalIssueReporter and ICrashLogger deps adhere to a single interface that just has a "report crash" method

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/jvmcrash/CaptureUncaughtExceptionHandler.kt
murki: yeah that's not ideal, specially since we rely on these logs ourselves for our instant insights dashboard, we should discuss this

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/events/lifecycle/AppExitLogger.kt
```diff
@@ -38,12 +39,11 @@ internal class AppExitLogger(
...
     private val runtime: Runtime,
     private val errorHandler: ErrorHandler,
-    private val crashHandler: CaptureUncaughtExceptionHandler = CaptureUncaughtExceptionHandler(),
+    private val fatalIssueReporter: IFatalIssueReporter,
     private val versionChecker: BuildVersionChecker = BuildVersionChecker(),
     private val memoryMetricsProvider: IMemoryMetricsProvider,
     private val backgroundThreadHandler: IBackgroundThreadHandler = CaptureDispatchers.CommonBackground,
-    private val latestAppExitInfoProvider: ILatestAppExitInfoProvider = LatestAppExitInfoProvider(),
-) {
+) : ICrashLogger {
```
murki, line 46: Seems like this class is no longer just an "AppExitLogger" and now more of a god crash handler that now sometimes logs the appexit log, or sometimes installs the falta issues reporter, or both, so maybe you should rename it. I'm trying to understand the entire dependency graph of the relevant classes and it's quite hard to follow

murki (changes requested):

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/jvmcrash/CaptureUncaughtExceptionHandler.kt
FranAguilera: Discussed offline, will keep existing logic to double emit (like in JVM crash + App Exit REASON_CRASH)

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/FatalIssueMechanism.kt
FranAguilera: Will polish in a bit

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/events/lifecycle/AppExitLogger.kt
FranAguilera: Agree, this doesn't apply anymore

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/FatalIssueReport.kt
FranAguilera: Good call, @kattrali we may need to tweak the model to avoid issues with this naming

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/FatalIssueReporter.kt
FranAguilera: There's no need, we use traceInputStream also being null to differentiate

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/jvmcrash/CaptureUncaughtExceptionHandler.kt
FranAguilera: This no longer applies here, but lmw what you think of recent approach

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/events/lifecycle/AppExitLogger.kt
FranAguilera: Will create a ticket to follow up

murki (approved):
looks good, remaining comments are suggestions

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/FatalIssueReport.kt
FranAguilera: tracking in here BIT-5094

FranAguilera: my bad, the model specifies Error not Exception

PR #298, opened 2025-04-10
[android] make report type use an enum value

kattrali:
there are a limited number of types, and this avoids string parsing during binary serialization


platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/persistence/FatalIssueReporterStorage.kt
```diff
@@ -27,7 +27,7 @@ internal class FatalIssueReporterStorage(
         fatalIssueReport: FatalIssueReport,
     ) {
         // TODO(FranAguilera): BIT-5083 Replace storing via Gson with native call
-        val fileName = "${terminationTimeStampInMilli}_${fatalIssueType.displayName}.json"
+        val fileName = "${terminationTimeStampInMilli}_${fatalIssueType.name}.json"
```
FranAguilera, line 30: I think name is by default not obfuscated but there could be customers they are explicitly obfuscating those, though I don't think we are going to use this file name type here for anything

FranAguilera (approved):

PR #301, opened 2025-04-10
[android] write reports to disk in bin format

kattrali:
* generates .kt models for the report structures
* sets a random ID for the report file name – in the future we should move this call higher up to ensure the same report isn't sent more than once
* writes the report to disk in the new binary format, with a simple test for verification
* updated related tests in the report processor classes

The vast majority of the changes are generated files, which can be avoided by looking at each commit.

### DIFF
--- a/Cargo.Bazel.lock
+++ b/Cargo.Bazel.lock
@@ -1,5 +1,5 @@
 {
-  "checksum": "0a20d776a8c80165bce30c20d9c3b1dfe0b879cf30cd7fff0dbe1f41d66906c2",
+  "checksum": "2e0e725d6de2d08a67d435f223cc5c3d0af0cb664ffb128492bd45848568566e",
   "crates": {
     "addr2line 0.24.2": {
       "name": "addr2line",
@@ -527,14 +527,14 @@
       ],
       "license_file": "LICENSE-APACHE"
     },
-    "anyhow 1.0.97": {
+    "anyhow 1.0.98": {
       "name": "anyhow",
-      "version": "1.0.97",
+      "version": "1.0.98",
       "package_url": "https://github.com/dtolnay/anyhow",
       "repository": {
         "Http": {
-          "url": "https://static.crates.io/crates/anyhow/1.0.97/download",
-          "sha256": "dcfed56ad506cb2c684a14971b8861fdc3baaaae314b9e5f9bb532cbe3ba7a4f"
+          "url": "https://static.crates.io/crates/anyhow/1.0.98/download",
+          "sha256": "e16d2d3311acee920a9eb8d33b8cbc1787ce4a264e85f964c2404b969bdcd487"
         }
       },
       "targets": [
@@ -578,14 +578,14 @@
         "deps": {
           "common": [
             {
-              "id": "anyhow 1.0.97",
+              "id": "anyhow 1.0.98",
               "target": "build_script_build"
             }
           ],
           "selects": {}
         },
         "edition": "2018",
-        "version": "1.0.97"
+        "version": "1.0.98"
       },
       "build_script_attrs": {
         "compile_data_glob": [
@@ -736,7 +736,7 @@
               "target": "pin_project_lite"
             },
             {
-              "id": "tokio 1.44.2",
+              "id": "tokio 1.45.0",
               "target": "tokio"
             }
           ],
@@ -886,14 +886,14 @@
       ],
       "license_file": "LICENSE-APACHE"
     },
-    "axum 0.8.3": {
+    "axum 0.8.4": {
       "name": "axum",
-      "version": "0.8.3",
+      "version": "0.8.4",
       "package_url": "https://github.com/tokio-rs/axum",
       "repository": {
         "Http": {
-          "url": "https://static.crates.io/crates/axum/0.8.3/download",
-          "sha256": "de45108900e1f9b9242f7f2e254aa3e2c029c921c258fe9e6b4217eeebd54288"
+          "url": "https://static.crates.io/crates/axum/0.8.4/download",
+          "sha256": "021e862c184ae977658b36c4500f7feac3221ca5da43e3f25bd04ab6c79a29b5"
         }
       },
       "targets": [
@@ -1015,7 +1015,7 @@
               "target": "sync_wrapper"
             },
             {
-              "id": "tokio 1.44.2",
+              "id": "tokio 1.45.0",
               "target": "tokio"
             },
             {
@@ -1051,7 +1051,7 @@
           ],
           "selects": {}
         },
-        "version": "0.8.3"
+        "version": "0.8.4"
       },
       "license": "MIT",
       "license_ids": [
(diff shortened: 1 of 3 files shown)


platform/jvm/capture/detekt.yml
```diff
@@ -1,4 +1,5 @@
 comments:
+  excludes: '**/binformat/**'
```
FranAguilera, line 2: What was the detekt issue on the files here? Can we fix those files instead of excluding whole dir?

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/binformat/v1/Report.kt
```diff
@@ -0,0 +1,1371 @@
...
+import com.google.flatbuffers.StringVector
+import com.google.flatbuffers.Struct
+import com.google.flatbuffers.Table
+import com.google.flatbuffers.UnionVector
+import java.nio.ByteBuffer
+import java.nio.ByteOrder
+import kotlin.math.sign
+
+@Suppress("unused")
+class ReportType private constructor() {
```
FranAguilera, line 29: Would using object instead of class break here? That would be a more kotlin way of doing this

platform/jvm/capture/detekt.yml
FranAguilera: actually no worries, but can you add a ticket to remove this once we move the fbs gen to shared-core?

platform/jvm/capture/src/test/kotlin/io/bitdrift/capture/FatalIssueReporterStorageTest.kt
```diff
@@ -0,0 +1,97 @@
...
+                0,
+            )
+        builder.finish(report)
+        val timestamp = 1744287332021
+        storage.persistFatalIssue(timestamp, builder.sizedByteArray())
+
+        val files = dir.toFile().listFiles()
+        assertThat(files?.size).isEqualTo(1)
+
+        val datafile = files!![0]
```
FranAguilera, line 82: minor nit: you could add assertThat(files).isNotNull() prior to force unwrap

bazel/android/build.bzl
```diff
@@ -15,7 +15,11 @@ def bitdrift_kt_android_library(name, srcs, require_javadocs = True, **args):
         **args
     )
 
-    _jvm_lint_support(name, srcs, require_javadocs)
+    _jvm_lint_support(
+        name,
+        native.glob(srcs, exclude = ["**/binformat/**"]),
```
FranAguilera, line 20: same comment as below: to add a ticket to remove ignore once moved to shared-core

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/binformat/v1/Report.kt
```diff
@@ -0,0 +1,1371 @@
...
+        const val Unknown: Byte = 0
+        const val AppNotResponding: Byte = 1
+        const val HandledError: Byte = 2
+        const val JVMCrash: Byte = 3
+        const val MemoryTermination: Byte = 4
+        const val NativeCrash: Byte = 5
+        const val StrictModeViolation: Byte = 6
+    }
+}
+@Suppress("unused")
```
FranAguilera, line 40: nit: Probably you can remove these suppress now given the whole dir is being excluded

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/binformat/v1/Report.kt
```diff
@@ -0,0 +1,1371 @@
...
+        }
+    companion object {
+        fun validateVersion() = Constants.FLATBUFFERS_25_2_10()
+        fun getRootAsDisplay(_bb: ByteBuffer): Display = getRootAsDisplay(_bb, Display())
+        fun getRootAsDisplay(_bb: ByteBuffer, obj: Display): Display {
+            _bb.order(ByteOrder.LITTLE_ENDIAN)
+            return (obj.__assign(_bb.getInt(_bb.position()) + _bb.position(), _bb))
+        }
+        fun createDisplay(builder: FlatBufferBuilder, height: UInt, width: UInt, densityDpi: UInt) : Int {
+            builder.startTable(3)
```
FranAguilera, line 441: out of curiosity, how did you figured out all these hardcoded indexes?

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/processor/FatalIssueReporterProcessor.kt
```diff
@@ -66,56 +61,98 @@ internal class FatalIssueReporterProcessor(
...
-                    FatalIssueReport(
+                    Report.createReport(
+                        builder,
                         sdk,
+                        ReportType.NativeCrash,
                         appMetrics,
                         deviceMetrics,
-                        errors = emptyList(),
-                        threadsDetails = ThreadDetails(),
+                        0,
```
FranAguilera, line 72: Nit: Same comment as above regarding defaults

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/processor/AppExitAnrTraceProcessor.kt
```diff
@@ -81,11 +80,18 @@ internal object AppExitAnrTraceProcessor {
         }
     }
 
+    // get description before any parenthesized content
+    private fun summarizeDescription(description: String?): String? {
+        if (description != null && description.indexOf(" (") > 0) {
```
FranAguilera, line 85: I think this will only work for Input dispatching time out. There are other common descriptions like

- Broadcast of Intent { act=android.intent.action.BOOT_COMPLETED flg=0x10 } timed out after 5000ms
- Executing service io.example.app/.MyService timed out after 10000ms

More info here https://developer.android.com/topic/performance/anrs/diagnose-and-fix-anrs?hl=en#broadcast-receiver-anr

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/processor/AppExitAnrTraceProcessor.kt
```diff
@@ -81,24 +83,38 @@ internal object AppExitAnrTraceProcessor {
...
-                    type = FrameType.JVM.ordinal,
-                    className = className,
-                    symbolName = "$className.$symbolName",
-                    sourceFile = sourceFile,
+                Frame.createFrame(
+                    builder,
+                    FrameType.JVM,
+                    builder.createString(className),
+                    builder.createString(symbolName),
+                    sourceFile,
```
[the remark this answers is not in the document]
FranAguilera: i think this was resolved accidentally. wdyt of having a defaults here so this gets called just like this?

```
Frame.createFrame(
                    builder,
                    FrameType.JVM,
                    builder.createString(className),
                    builder.createString(symbolName),
                    sourceFile)
```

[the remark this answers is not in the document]
FranAguilera: ah that's right. missed that from before

FranAguilera (approved):

PR #305, opened 2025-04-11
[Android] FatalIssueReporting app context fix

FranAguilera:
- This will prevent any runtime crashing while trying to get a valid context. Prior PR for context https://github.com/bitdriftlabs/capture-sdk/pull/254 where APP_CONTEXT was initially added as lazy
- Related to BIT-5113

### DIFF
--- a/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/FatalIssueReporter.kt
+++ b/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/FatalIssueReporter.kt
@@ -8,6 +8,7 @@
 package io.bitdrift.capture.reports
 
 import android.os.Build
+import android.util.Log
 import androidx.annotation.UiThread
 import androidx.annotation.VisibleForTesting
 import io.bitdrift.capture.ContextHolder.Companion.APP_CONTEXT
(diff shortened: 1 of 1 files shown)


platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/FatalIssueReporter.kt
```diff
@@ -47,7 +48,9 @@ internal class FatalIssueReporter(
                 FatalIssueReporterStatus(fatalIssueReporterState, duration)
             }
         }.getOrElse {
-            FatalIssueReporterStatus(ProcessingFailure("Error while processCrashReportFile. ${it.message}"))
+            val errorMessage = "Error while initializing fatal issue reporter. $it"
```
murki, line 51: should put the word bitdrift somewhere since i noticed that's what people filter on in logcat

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/FatalIssueReporter.kt
```diff
@@ -29,7 +30,7 @@ import kotlin.time.measureTime
 internal class FatalIssueReporter(
     private val mainThreadHandler: MainThreadHandler = MainThreadHandler(),
 ) {
-    private val appContext = APP_CONTEXT
+    private val appContext by lazy { APP_CONTEXT }
```
murki, line 33: thanks for proving me wrong in that this worked!

murki (approved):

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/FatalIssueReporter.kt
FranAguilera: good idea, adding

PR #307, opened 2025-04-16
[Android] Pass application context into initFatalIssueReporting

FranAguilera:
Given that there could be cases where customer init the library via content providers there is still a chance that ContextHolder initializer is not yet called prior to `Capture.Logger.initFatalIssueReporting` call.

In this case making it explicit that will need an application context in order to initialize fatal issue reporting

### DIFF
--- a/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/Capture.kt
+++ b/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/Capture.kt
@@ -7,6 +7,7 @@
 
 package io.bitdrift.capture
 
+import android.content.Context
 import android.util.Log
 import com.github.michaelbull.result.Err
 import io.bitdrift.capture.common.MainThreadHandler
@@ -106,8 +107,8 @@ object Capture {
         @Suppress("UnusedPrivateMember")
         @ExperimentalBitdriftApi
         @JvmStatic
-        fun initFatalIssueReporting() {
-            val fatalIssueReporter = FatalIssueReporter()
+        fun initFatalIssueReporting(applicationContext: Context) {
+            val fatalIssueReporter = FatalIssueReporter(applicationContext)
             if (fatalIssueReporterStatus.state is FatalIssueReporterState.NotInitialized) {
                 fatalIssueReporterStatus = fatalIssueReporter.processPriorReportFiles()
             } else {
--- a/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/FatalIssueReporter.kt
+++ b/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/FatalIssueReporter.kt
@@ -7,11 +7,12 @@
 
 package io.bitdrift.capture.reports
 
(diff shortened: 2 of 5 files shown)


platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/FatalIssueReporter.kt
```diff
@@ -28,17 +29,21 @@ import kotlin.time.measureTime
...
     /**
      * Process existing crash report files.
      *
      * @return The status of the crash reporting processing
      */
-    fun processPriorReportFiles(): FatalIssueReporterStatus =
-        runCatching {
+    fun processPriorReportFiles(): FatalIssueReporterStatus {
+        if (applicationContext !is Application) {
+            val errorMessage = "Non-Application context passed into initFatalIssueReporting"
```
murki, line 42: why do we need it to be the application instance ? In particular for content providers I don't think that's always the case

FranAguilera: I'd rather be explicit here, we are mainly dealing with files operations and app context is better suited for this plus minimizes the risk of leaks that activity/service context have

murki: do you have references on why app context is better suited for file operations? afaik all the methods you're using come from the context itself. also have you tested this even works if called from a Content Provider?

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/Capture.kt
```diff
@@ -106,8 +107,8 @@ object Capture {
         @Suppress("UnusedPrivateMember")
         @ExperimentalBitdriftApi
         @JvmStatic
-        fun initFatalIssueReporting() {
-            val fatalIssueReporter = FatalIssueReporter()
+        fun initFatalIssueReporting(applicationContext: Context) {
```
murki, line 110: it's weird to call this `applicationContext` since there's no guarantees that that's what the user will be passing (as it will be very well the case when called from content providers)

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/FatalIssueReporter.kt
FranAguilera: @murki Yes, I’ve tested this with ContentProviders, and it works, but you’re right there could be cases where a ContentProvider doesn’t provide a valid application context. Regarding using the applicationContext is generally safer, since it lives for the entire app lifecycle, unlike an Activity or Service which can be destroyed/recreated. On the other hand, relying on start to release an activity context reference could be risky for cases that start is never called (e.g. via feature flag off). Either way, I’m not a fan of passing context here and will explore other approaches

@kattrali Yup that would be the ideal case, but looks like a potential case where customers can initialize via ContentProvider or AppStartup Initializers

PR #308, opened 2025-04-16
[andr] Allow to explicitly pass context to init() method

murki:
Allow sdk consumers to explicitly pass a context instant when initializing crash reporting. This is particularly useful when they want to call our `initFatalIssueReporting()` method from a content provider since they can do:
```kotlin
    override fun onCreate(): Boolean {
        @OptIn(ExperimentalBitdriftApi::class)
        Capture.Logger.initFatalIssueReporting(getContext())
        Bugsnag.start(getContext()!!, "bugsnag_apikey")
        return true
    }
```

This is an alternative proposal to #307

### DIFF
--- a/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/Capture.kt
+++ b/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/Capture.kt
@@ -56,7 +56,7 @@ internal sealed class LoggerState {
  * Top level namespace Capture SDK.
  */
 object Capture {
-    private const val LOG_TAG = "bitdrift capture"
+    private const val LOG_TAG = "BitdriftCapture"
     private val default: AtomicReference<LoggerState> = AtomicReference(LoggerState.NotStarted)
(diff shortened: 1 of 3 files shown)


platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/Capture.kt
```diff
@@ -101,18 +103,26 @@ object Capture {
         /**
          * Initializes fatal issue (ANR, JVM Crash, Native crash) reporting mechanism.
          *
+         * @param context an optional context reference. You should provide the context if called from a [android.content.ContentProvider]
+         *
          * This should be called prior to Capture.Logger.start()
          */
         @Suppress("UnusedPrivateMember")
         @ExperimentalBitdriftApi
         @JvmStatic
```
FranAguilera, line 112: in order to make Java to not specify null if initialized on a non content provider you can add @JvmOverloads here

FranAguilera (approved):

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/Capture.kt
murki: ah great call

PR #310, opened 2025-04-18
[andr] Proposal to phase out AppMemTrim

murki:
Trim Levels have been deprecated [as of Android API 34](https://developer.android.com/reference/android/content/ComponentCallbacks2#summary), and thus it's [not recommended to use this callback at all](https://developer.android.com/games/optimize/vitals/lmk#low-memory):
Screenshot 2025-04-18 at 5 11 34 PM

The proposal is to stop emitting "AppMemTrim" logs altogether (which has been historically noisy and unreliable) and instead substitute it by a new "AppMemPressure" log (name chosen by mirroring existing log on iOS). We will leverage the memory snapshots we take periodically for Resource Utilization to determine whether the memory is "low" and emit the new log if that's the case.  

Link to a session testing this via the new OOM Exception crasher: https://timeline.bitdrift.dev/session/4ac023a0-6cfc-4067-b043-391e631f40ef?pages=1&utilization=1&debug=true&type=2&type=3&h=-2896361944634902660&expanded=-2896361944634902660

frontend rendering after changes in https://github.com/bitdriftlabs/bitdrift-frontend/pull/2273:
Screenshot 2025-05-16 at 5 17 07 PM


platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/events/performance/MemoryMetricsProvider.kt
```diff
@@ -8,32 +8,44 @@
...
+            "_jvm_total_kb" to totalJvmMemoryBytes().bToKb(),
+            "_native_used_kb" to allocatedNativeHeapSizeBytes().bToKb(),
+            "_native_total_kb" to totalNativeHeapSizeBytes().bToKb(),
+            "_threshold_mem_kb" to memoryThresholdBytes.bToKb(),
+            "_is_memory_low" to isMemoryLow().toString(),
+            "_memory_class" to memoryClassMB().toString(),
         )
 
-    private fun totalJvmMemory(): String = (Runtime.getRuntime().totalMemory() / KB).toString()
+    override fun isMemoryLow(): Boolean = usedJvmMemoryBytes() > memoryThresholdBytes
```
murki, line 38: the is the new in-house logic

murki: another option would be to just calculate it as a percentage between used memory and `Runtime.getRuntime().maxMemory()`

FranAguilera: Given that Runtime.getRuntime().maxMemory() only captures JVM we may be missing important info for native, the only downside of threshold is that it is system wide and not app specific but still good to know this info

murki: jvm should be good enough as a start. trim levels only concerned with jvm levels as well

FranAguilera (approved):

FranAguilera (approved):

PR #314, opened 2025-04-22
[Android] [Gradle Test App] Adding strict mode reporting to gradle demo app

FranAguilera:
Demo session https://timeline.bitdrift.dev/session/5ac10aff-278f-4bb0-b075-f86bfc02a825?utm_source=sdk&pages=1


murki (approved):
nice, eventually you can think of a world where our sdk captures all these automatically

PR #315, opened 2025-04-25
[Android] [BuiltIn-FatalIssueReporter] Ensure that appexit processing runs off main thread

FranAguilera:
This was a leftover from https://github.com/bitdriftlabs/capture-sdk/pull/296

TBF some samples on real device with prior ANR

### DIFF
--- a/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/FatalIssueReporter.kt
+++ b/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/FatalIssueReporter.kt
@@ -14,6 +14,7 @@ import android.util.Log
 import androidx.annotation.UiThread
 import androidx.annotation.VisibleForTesting
 import io.bitdrift.capture.Capture.LOG_TAG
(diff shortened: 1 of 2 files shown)


FranAguilera:
@kattrali @murki Actually this won't work without shared-core changes given the logic there is relying on main thread. Will start refactoring to move all this logic into shared-core

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/FatalIssueReporter.kt
```diff
@@ -150,7 +153,9 @@ internal class FatalIssueReporter(
                                 FatalIssueReporterStorage(destinationDirectory.destinationDirectory),
                             )
                         captureUncaughtExceptionHandler.install(this)
-                        persistLastExitReasonIfNeeded(appContext)
+                        backgroundThreadHandler.runAsync {
```
FranAguilera, line 156: This happens on next app launch, which will be too late for shared-core logic to detect the file if we won't do this on main thread. Will be following up on this refactor here BIT-5191

PR #316, opened 2025-04-28
Capture threads for fatal issues

kattrali:


platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/FatalIssueReporter.kt
```diff
@@ -77,12 +77,14 @@ internal class FatalIssueReporter(
     override fun onJvmCrash(
         thread: Thread,
         throwable: Throwable,
+        allThreads: Map<Thread, Array<StackTraceElement>>?,
```
FranAguilera, line 80: Discussed offline, can we defer getting allThread within onJvmCrash instead?. We we merge back to main this will mean that we are calling `Thread.getAllStackTraces()` even for customers not using built in crash reporting

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/jvmcrash/CaptureUncaughtExceptionHandler.kt
```diff
@@ -54,14 +50,14 @@ internal object CaptureUncaughtExceptionHandler : ICaptureUncaughtExceptionHandl
     override fun install(jvmCrashListener: JvmCrashListener) {
         crashListeners.add(jvmCrashListener)
         if (!installed) {
+            prevExceptionHandler = Thread.getDefaultUncaughtExceptionHandler()
```
FranAguilera, line 53: This is also being handled on this PR https://github.com/bitdriftlabs/capture-sdk/pull/317, we could rebase when this gets merged

FranAguilera (approved):

PR #317, opened 2025-04-29
fix(jvmcrash): grab prev handler at install time

kattrali:
avoids an edge case where another handler is set as default between when class is created and installation

### DIFF
--- a/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/jvmcrash/CaptureUncaughtExceptionHandler.kt
+++ b/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/jvmcrash/CaptureUncaughtExceptionHandler.kt
@@ -16,14 +16,11 @@ import java.util.concurrent.CopyOnWriteArrayList
  * that will notify the specified listener when a JVM crash has occurred
  */
 internal object CaptureUncaughtExceptionHandler : ICaptureUncaughtExceptionHandler {
-    private var crashing = false
+    @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+    var crashing = false
     private var installed: Boolean = false
-    private val prevExceptionHandler: UncaughtExceptionHandler? =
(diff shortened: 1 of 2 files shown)


platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/jvmcrash/CaptureUncaughtExceptionHandler.kt
```diff
@@ -16,15 +16,12 @@ import java.util.concurrent.CopyOnWriteArrayList
  * that will notify the specified listener when a JVM crash has occurred
  */
 internal object CaptureUncaughtExceptionHandler : ICaptureUncaughtExceptionHandler {
-    private var crashing = false
+    @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+    public var crashing = false
```
FranAguilera, line 20: Nit: We could limit viz a bit more and make this internal instead

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/jvmcrash/CaptureUncaughtExceptionHandler.kt
```diff
@@ -16,15 +16,12 @@ import java.util.concurrent.CopyOnWriteArrayList
...
     private var installed: Boolean = false
-    private val prevExceptionHandler: UncaughtExceptionHandler? =
-        Thread.getDefaultUncaughtExceptionHandler()
+    private var prevExceptionHandler: UncaughtExceptionHandler? = null
```
FranAguilera, line 22: Wondering if this is an issue on main too given this is created now on AppExitLogger instance

https://github.com/search?q=repo%3Abitdriftlabs%2Fcapture-sdk%20CaptureUncaughtExceptionHandler&type=code

FranAguilera (approved):
thanks for fixing!

PR #320, opened 2025-04-29
Fixing main build errors

FranAguilera:
This needed now given we have [multiple fbs](https://github.com/search?q=repo%3Abitdriftlabs%2Fapi+fbs&type=code) in api now

```
ERROR: /home/runner/work/capture-sdk/capture-sdk/proto/BUILD:4:26: declared output 'proto/buffer_log.schema.json' was not created by genrule. This is probably because the genrule actually didn't create this output, or because the output was a directory and the genrule was run remotely (note that only the contents of declared file outputs are copied from genrules run remotely)
ERROR: /home/runner/work/capture-sdk/capture-sdk/proto/BUILD:4:26: Generating flatbuffer files for descriptor: //proto:descriptor failed: not all outputs were created or valid

```
Discussed here https://bitdrift.slack.com/archives/C058ZD2M2CQ/p1745949413605059

### DIFF
--- a/bazel/BUILD.bitdriftlabs_api
+++ b/bazel/BUILD.bitdriftlabs_api
@@ -9,3 +9,9 @@ filegroup(
   srcs = glob(["**/*.fbs"]),
   visibility = ["//visibility:public"],
 )
+
+filegroup(
+  name = "buffer_log",
+  srcs = ["src/bitdrift_public/fbs/logging/v1/buffer_log.fbs"],
+  visibility = ["//visibility:public"],
+)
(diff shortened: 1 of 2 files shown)


murki (approved):
hoping it works

FranAguilera:
Depending on this approach for now https://github.com/bitdriftlabs/api/pull/50

bazel/BUILD.bitdriftlabs_api
```diff
@@ -9,3 +9,9 @@ filegroup(
   srcs = glob(["**/*.fbs"]),
   visibility = ["//visibility:public"],
 )
+
```
FranAguilera, line 12: tbh not sure why we need this here and in API

PR #345, opened 2025-05-02
[WIP] [Android] [FatalIssueReporting] Adding signal process method

FranAguilera:
In order to be able to process app exit reports off main thread, will need to send a signal to shared-core to start that processing

Will attach here also related shared-core PR when ready


platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/FatalIssueReporter.kt
```diff
@@ -150,7 +154,9 @@ internal class FatalIssueReporter(
                                 FatalIssueReporterStorage(destinationDirectory.destinationDirectory),
                             )
                         captureUncaughtExceptionHandler.install(this)
-                        persistLastExitReasonIfNeeded(appContext)
+                        backgroundThreadHandler.runAsync {
```
murki, line 157: any concerns about this causing a race condition for the appexitinfo case between `initFatalIssueReporting()` and `start()` (which would read the file report) and now  `processFatalIssue()`?

FranAguilera:
Discussed offline, abandoning this approach for now

PR #347, opened 2025-05-02
[Android] Verify apk size check on latest main

FranAguilera:
TBF


FranAguilera:
Getting closer, unfortunately the baseline logic is off 

image

It shoudl compare agains't the one above^

image

PR #348, opened 2025-05-02
[Android] APK size check - further fix

FranAguilera:
Further tweaks to https://github.com/bitdriftlabs/capture-sdk/pull/346 

We need to upload apk size on latest main as baseline

Will be using this to validate https://github.com/bitdriftlabs/capture-sdk/pull/347

### DIFF
--- a/.github/workflows/android.yaml
+++ b/.github/workflows/android.yaml
@@ -72,74 +72,80 @@ jobs:
         if: env.run_tests == 'true'
         run: ./ci/run_tests.sh
 
-  build_apk:
+  build_and_compare_apk:
     runs-on: ubuntu-latest
     if: needs.pre_check.outputs.should_run == 'true'
     needs: pre_check
     steps:
-      # Checkout repo to Github Actions runner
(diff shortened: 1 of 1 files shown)


.github/workflows/android.yaml
```diff
@@ -77,69 +77,66 @@ jobs:
...
+      - name: Upload baseline APK if on main
+        if: github.ref == 'refs/heads/main'
+        uses: actions/upload-artifact@v4
         with:
-          path: baseline_apk_size.txt
-          key: apk-baseline
+          name: apk-main-baseline
+          path: ./bazel-bin/examples/android/android_app.apk
+
+      - name: Download baseline APK from main
```
murki, line 102: should the download only happen if in a PR? otherwise for builds on main it would be redundant

murki (approved):
trial and error is the name of the game

.github/workflows/android.yaml
FranAguilera: good catch, will update

PR #362, opened 2025-05-09
[Android] [BIT-4820] Fix frame duration workflow

FranAguilera:
Resolves BIT-4820

The theory here is that _duration_ms is a frequently used field that can get triggered on other events relying on this, so the proposal here is to update the event with a more explicit name

This depends on https://github.com/bitdriftlabs/bitdrift-frontend/pull/2236 being deployed

Verified with client changes emitting `_frame_duration_ms` and existing `_duration_ms`

image

image


jacksonhardaker (approved):
Not a mobile engineer, but lgtm

PR #371, opened 2025-05-12
[android] Avoid crashing when modern pixel copy fails

murki:
Caught this when a customer enabled the experimental feature for a day (~54 incidences)


platform/jvm/replay/src/main/kotlin/io/bitdrift/capture/replay/internal/ScreenshotCaptureEngine.kt
```diff
@@ -71,16 +71,17 @@ internal class ScreenshotCaptureEngine(
                 .ofWindow(topView)
                 .build()
         PixelCopy.request(screenshotRequest, executor) { screenshotResult ->
-            val resultBitmap = screenshotResult.bitmap
```
murki, line 74: there's a `validateStatus()` inside the getter that throws an `IllegalStateException` when the `status != SUCCESS`

FranAguilera (approved):

PR #376, opened 2025-05-13
ios: ingest mx reports

kattrali:
creates a logs from MXCrashDiagnostic payloads, adding a new `.crashReporting()` Integration option

a couple caveats:

* the ffi header is copied in manually — preferably there would be a good way to copy it over during the compilation of the `bd-report-writer` crate to keep them in sync
* this runs immediately on the same thread as when `.crashReporting()` is called

### DIFF
--- a/platform/swift/source/reports/DiagnosticEventReporter.m
+++ b/platform/swift/source/reports/DiagnosticEventReporter.m
@@ -100,6 +100,15 @@ static inline const char * cstring_from(NSString *str) {
   return [str cStringUsingEncoding:NSUTF8StringEncoding];
 }
 
+static uint32_t crashed_thread_index(NSArray *stacks) {
+  for (uint32_t index = 0; index < stacks.count; index++) {
+    if ([number_for_key(stacks[index], @"threadAttributed") boolValue]) {
+      return index;
+    }
+  }
+  return 0; // first thread is crashed thread if none attributed
+}
+
 static uint64_t count_frames(NSDictionary *rootFrame) {
   uint64_t count = 0;
   NSDictionary *frame = rootFrame;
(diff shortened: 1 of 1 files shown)


platform/swift/source/Capture.swift
```diff
@@ -14,6 +14,11 @@ public typealias Fields = [String: FieldValue]
 /// The value of a log field.
 public typealias FieldValue = Encodable & Sendable
 
+public enum IssueReporterType {
+    case builtIn
+    case customConfig
```
FranAguilera, line 19: nit: this was updated to `Integration` https://github.com/bitdriftlabs/capture-sdk/blob/067df1c6aeb9ec200a71026d9906e7302f505bbc/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/FatalIssueMechanism.kt#L24

platform/swift/source/reports/DiagnosticEventReporter.m
```diff
@@ -0,0 +1,295 @@
...
+  return self;
+}
+
+- (void)didReceiveDiagnosticPayloads:(NSArray<MXDiagnosticPayload *> * _Nonnull)payloads API_AVAILABLE(ios(14.0), macos(12.0)) {
+  NSFileManager *fileManager = [NSFileManager defaultManager];
+  if (![fileManager createDirectoryAtPath:[self.dir path] withIntermediateDirectories:YES attributes:0 error:nil]) {
+    return;
+  }
+
+  const void *handle = 0;
```
Reflejo, line 63: Should this be `BDProcessorHandle` instead of `const void *`?

platform/swift/source/reports/DiagnosticEventReporter.m
```diff
@@ -0,0 +1,295 @@
...
+  return [arch containsString:@"arm64"] ? /* Architecture.arm64 */ 2 : /* Architecture.x86_64 */ 4;
+}
+
+static void serialize_device_metrics(BDProcessorHandle handle, NSDictionary *metadata, NSTimeInterval timestamp) {
+  NSRegularExpression *version_matcher = [NSRegularExpression regularExpressionWithPattern:OS_VERSION_MATCHER options:0 error:nil];
+  NSString *version_info = string_for_key(metadata, @"osVersion");
+  NSTextCheckingResult *match = [[version_matcher matchesInString:version_info options:0 range:NSMakeRange(0, version_info.length)] firstObject];
+  NSString *os_name = match ? [version_info substringWithRange:[match rangeWithName:@"osName"]] : nil;
+  NSString *os_version = match ? [version_info substringWithRange:[match rangeWithName:@"osVersion"]] : nil;
+  NSString *os_kernversion = match ? [version_info substringWithRange:[match rangeWithName:@"buildNumber"]] : nil;
```
Reflejo, line 171: If one of these `rangeWithName` are not present in the regex grouping this will crash I believe

platform/swift/source/Capture.swift
```diff
@@ -90,12 +95,29 @@ extension Logger {
...
-            issueReporterInitResult = FatalIssueReporter.processFiles()
-        } else {
-            log(level: .warning, message: "Fatal issue reporting already being initialized")
+    ///
+    /// - parameter type: mechanism for crash detection
+    public static func initFatalIssueReporting(_ type: IssueReporterType = .customConfig) {
+        switch type {
+        case .builtIn:
+            if let outputDir = Logger.reportCollectionDirectory() {
+                if #available(iOS 14, *) {
```
Reflejo, line 104: What's the plan for < 14.0?

Reflejo (commented):
overall looks good! One question: MetricsKit only delivers the diagnostics immediately in iOS 15+ so what's our plan for the gap?

murki (approved):

PR #382, opened 2025-05-15
[Android] [FatalIssueReporting][BuiltIn] Append reason for ANR case

FranAguilera:
Resolves BIT-5378

Adding ANR types as the reason for AppExit ANR

- Currently on main

https://timeline.bitdrift.dev/session/1b479b2d-bbe8-4984-a98f-93affb76f72f?utm_source=sdk&pages=1&utilization=0&expanded=3223018819215494859

- On PR
https://timeline.bitdrift.dev/s/b5ad76f1-de87-4b43-bde5-4b0bfc43db81?utm_source=sdk

### DIFF
--- a/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/processor/AppExitAnrTraceProcessor.kt
+++ b/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/processor/AppExitAnrTraceProcessor.kt
@@ -15,6 +15,10 @@ import io.bitdrift.capture.reports.binformat.v1.FrameType
 import io.bitdrift.capture.reports.binformat.v1.Report
 import io.bitdrift.capture.reports.binformat.v1.ReportType
 import io.bitdrift.capture.reports.binformat.v1.SourceFile
+import io.bitdrift.capture.reports.processor.AppExitAnrTraceProcessor.AnrReason.BroadcastReceiver
+import io.bitdrift.capture.reports.processor.AppExitAnrTraceProcessor.AnrReason.ExecutingService
+import io.bitdrift.capture.reports.processor.AppExitAnrTraceProcessor.AnrReason.Generic
+import io.bitdrift.capture.reports.processor.AppExitAnrTraceProcessor.AnrReason.StartForegroundNotCalled
 import java.io.BufferedReader
 import java.io.InputStream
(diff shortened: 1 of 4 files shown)


platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/processor/AppExitAnrTraceProcessor.kt
```diff
@@ -118,4 +123,53 @@ internal object AppExitAnrTraceProcessor {
...
+        }
+    }
+
+    /**
+     * Based on android source (see frameworks/base/core/java/com/android/internal/os/TimeoutRecord.java)
+     */
+    private sealed class AnrReason(
+        val message: String,
+    ) {
+        // Note: Combining all Input Dispatching Timed Out as User Perceived ANR as per public definition
```
murki, line 152: maybe should add a link or the quote from here https://developer.android.com/topic/performance/vitals/anr#android-vitals
>Currently only ANRs of type `Input dispatching timed out` are considered user-perceived.

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/processor/AppExitAnrTraceProcessor.kt
```diff
@@ -118,4 +123,53 @@ internal object AppExitAnrTraceProcessor {
...
+
+        data object ShortFgsTimeout : AnrReason("Short Foreground Service Timeout ANR")
+
+        data object JobService : AnrReason("Job Service ANR")
+
+        data object AppStart : AnrReason("App Start ANR")
+
+        data object ServiceStart : AnrReason("Service Start ANR")
+
+        data object Generic : AnrReason("Generic ANR")
```
murki, line 173: >"Generic ANR"
sounds a little weird without a definition, how about something like "Undetermined ANR" or something like that?

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/processor/AppExitAnrTraceProcessor.kt
FranAguilera: Good idea

murki (approved):
v nice

PR #383, opened 2025-05-15
fix(anr): handle missing fields in frame matcher

kattrali:
## perf improvements
* added early exits for method and file parsing by excluding separator characters ('.' and ':')
* skip string trim in favor of what should be a fairly quick and performant match on exactly whitespace + 'at'

## fixes
* handle lines where one or more of the stack fields (method, file, line number) are missing + additional tests

added extra docs for the regex itself for future travelers


FranAguilera (approved):
lgtm, for completeness could you add a timeline session link in PR description with ANR stacktrace?

PR #397, opened 2025-05-22
feat(jvm): add sdk version to reports

kattrali:
there's already a nice test for ensuring the value set is sent (via `FatalIssueReporterStorageTest`. This change propagates the value from core to reports, while stubbing the value for the gradle build (and confirming the correct value is set via unit tests)

[sample](https://timeline.bitdrift.dev/session/6A4A8A2A-31B4-48DB-9F4A-C57B53F2164D?utm_source=explorations.bitdrift.dev&exploration_id=7914&workflow_id=rfz9&pages=1&utilization=0&expanded=-6865857055975819944) (you'd need to look at the actual crash artifact to confirm, I think)


FranAguilera (changes requested):
Actually let's wait for a moment, I wasn't able to get ANR/crash reports after this

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/processor/FatalIssueReporterProcessor.kt
```diff
@@ -166,3 +170,11 @@ internal class FatalIssueReporterProcessor(
         return DeviceMetrics.endDeviceMetrics(builder)
     }
 }
+
+internal interface ISDKVersionProvider {
+    fun getSDKVersion(): String?
+}
+
+private class CoreSDKVersionProvider : ISDKVersionProvider {
+    override fun getSDKVersion(): String? = CaptureJniLibrary.getSdkVersion()
```
FranAguilera, line 179: looks like this is the issue for ANR/AppExit case given we still haven't loaded the native library upon start call

getting 2025-05-23 22:25:35.050 24571-24571 BitdriftCapture         io.bitdrift.gradletestapp            E  Error while initializing reporter for BuiltIn mode. No implementation found for java.lang.String io.bitdrift.capture.CaptureJniLibrary.getSdkVersion() (tried Java_io_bitdrift_capture_CaptureJniLibrary_getSdkVersion and Java_io_bitdrift_capture_CaptureJniLibrary_getSdkVersion__) - is the library loaded, e.g. System.loadLibrary?

platform/shared/BUILD
```diff
@@ -20,3 +26,12 @@ filegroup(
     srcs = [".sdk_version"],
     visibility = ["//visibility:public"],
 )
+
+genrule(
+    name = "version_kt_source",
+    outs = ["BuildConstants.kt"],
```
murki, line 32: this might be for a future pr but just be aware that this file won't be picked up by the gradle system automatically we might need to configure them via gradle sourceSets https://docs.gradle.org/current/dsl/org.gradle.api.tasks.SourceSet.html
this might be more relevant later when we want to actually expose this value via our API

PR #408, opened 2025-05-30
[Android] Force start to run off main thread

FranAguilera:
- Forces internals of Logger.start() to always happen of the main thread
- Add temporary in memory cache for any Logger.log(*) calls that happen during Logger.start() gets initialized

## Verification

[Session](https://timeline.bitdrift.dev/session/4d4871db-296e-4e31-9d77-860b83ac8648?utm_source=sdk&pages=1&utilization=0&expanded=5334525886215530043) with a simulated expensive LoggerImpl creation and logs being emitted while start is in process 

image

image

image

### DIFF
--- a/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/Capture.kt
+++ b/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/Capture.kt
@@ -10,6 +10,7 @@ package io.bitdrift.capture
 import android.content.Context
 import android.util.Log
 import com.github.michaelbull.result.Err
+import io.bitdrift.capture.common.IBackgroundThreadHandler
 import io.bitdrift.capture.common.MainThreadHandler
 import io.bitdrift.capture.events.span.Span
 import io.bitdrift.capture.events.span.SpanResult
@@ -22,6 +23,7 @@ import io.bitdrift.capture.providers.SystemDateProvider
 import io.bitdrift.capture.providers.session.SessionStrategy
 import io.bitdrift.capture.reports.FatalIssueMechanism
 import io.bitdrift.capture.reports.FatalIssueReporter
+import io.bitdrift.capture.threading.CaptureDispatchers
 import okhttp3.HttpUrl
 import java.util.UUID
 import java.util.concurrent.atomic.AtomicReference
@@ -35,14 +37,17 @@ internal sealed class LoggerState {
 
     /**
      * The logger is in the process of being started. Subsequent attempts to start the logger will be ignored.
+     * Any calls to Logger.log() meanwhile Logger.start() is in process will be cached in memory
      */
-    data object Starting : LoggerState()
+    class Starting(
+        val preInitInMemoryLogger: PreInitInMemoryLogger,
+    ) : LoggerState()
 
     /**
      * The logger has been successfully started and is ready for use. Subsequent attempts to start the logger will be ignored.
      */
     class Started(
-        val logger: LoggerImpl,
+        val loggerImpl: LoggerImpl,
     ) : LoggerState()
 
     /**
@@ -58,6 +63,7 @@ object Capture {
     internal const val LOG_TAG = "BitdriftCapture"
     private val default: AtomicReference<LoggerState> = AtomicReference(LoggerState.NotStarted)
     private val fatalIssueReporter = FatalIssueReporter()
+    private val preInitInMemoryLogger by lazy { PreInitInMemoryLogger() }
 
     /**
      * Returns a handle to the underlying logger instance, if Capture has been started.
@@ -67,8 +73,8 @@ object Capture {
     fun logger(): ILogger? =
         when (val state = default.get()) {
             is LoggerState.NotStarted -> null
-            is LoggerState.Starting -> null
-            is LoggerState.Started -> state.logger
+            is LoggerState.Starting -> state.preInitInMemoryLogger
+            is LoggerState.Started -> state.loggerImpl
             is LoggerState.StartFailure -> null
         }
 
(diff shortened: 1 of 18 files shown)


platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/Capture.kt
```diff
@@ -183,23 +190,28 @@ object Capture {
...
+                                fieldProviders = fieldProviders,
+                                dateProvider = dateProvider ?: SystemDateProvider(),
+                                configuration = configuration,
+                                sessionStrategy = sessionStrategy,
+                                bridge = bridge,
+                                fatalIssueReporter = fatalIssueReporter,
+                            )
+                        inMemoryLogger.flushToNative(loggerImpl)
+                        loggerImpl.startLogger()
+                        default.set(LoggerState.Started(loggerImpl))
```
FranAguilera, line 209: Still need to verify this a bit more

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/PreInitInMemoryLogger.kt
```diff
@@ -0,0 +1,117 @@
...
+    private fun addLoggerCall(logCall: (ILogger) -> Unit) {
+        if (_bufferedLoggerCalls.size >= MAX_LOG_CALL_SIZE) {
+            _bufferedLoggerCalls.poll()
+        }
+        _bufferedLoggerCalls.add(logCall)
+    }
+
+    private companion object {
+        private const val DEFAULT_NOT_SETUP_MESSAGE = "SDK starting"
+        private const val MAX_LOG_CALL_SIZE = 1024
```
murki, line 115: what does this size represent? is it the number of function types stored in the heap?

and how was this size determined? does it match the number in the pre-config buffer?

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/IPreInitLogFlusher.kt
```diff
@@ -0,0 +1,15 @@
...
+// LICENSE file or at:
+// https://polyformproject.org/wp-content/uploads/2020/06/PolyForm-Shield-1.0.0.txt
+package io.bitdrift.capture
+
+/**
+ * Flushes the internal memory logging into the loggerImpl
+ */
+fun interface IPreInitLogFlusher {
+    /** Flush all in memory Logger calls into the Native `LoggerImpl` **/
+    fun flushToNative(nativeLogger: ILogger)
```
murki, line 14: the use of "native" is odd here since technically you're flushing to JVM LoggerImpl

murki (approved):

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/PreInitInMemoryLogger.kt
```diff
@@ -0,0 +1,117 @@
...
+    override fun startSpan(
+        name: String,
+        level: LogLevel,
+        fields: Map<String, String>?,
+        startTimeMs: Long?,
+        parentSpanId: UUID?,
+    ): Span {
+        val span = Span(null, name, level, fields, startTimeMs, parentSpanId)
+        addLoggerCall { it.startSpan(name, level, fields, startTimeMs, parentSpanId) }
+        return span
```
snowp, line 90: did you verify that this correctly closes the span? seems like we might be returning a span with a null logger so it won't ever be closed with a logger?

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/PreInitInMemoryLogger.kt
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+    override fun log(
+        level: LogLevel,
+        fields: Map<String, String>?,
+        throwable: Throwable?,
+        message: () -> String,
+    ) {
+        addLoggerCall { it.log(level, fields, throwable, message) }
+    }
```
snowp, line 71: One potential issue with this is that the timestamp of the logs will be at the time that they were replayed, not the actual time they were logged? Maybe fine for now but something to keep in mind

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/Capture.kt
```diff
@@ -184,23 +191,27 @@ object Capture {
...
+                                apiUrl = apiUrl,
+                                fieldProviders = fieldProviders,
+                                dateProvider = dateProvider ?: SystemDateProvider(),
+                                configuration = configuration,
+                                sessionStrategy = sessionStrategy,
+                                bridge = bridge,
+                                fatalIssueReporter = fatalIssueReporter,
+                                preInitLogFlusher = preInitInMemoryLogger,
+                            )
+                        default.set(LoggerState.Started(loggerImpl))
```
snowp, line 209: From a concurrency perspective it seems like there might be a chance that logs are written to the in memory queue after it gets cleared but before we update the LoggerState, which seems like it could result in dropped logs? 

I don't have a good solution for how to avoid this without making it significantly more complicated so maybe it's fine?

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/PreInitInMemoryLogger.kt
FranAguilera: that's right, is the max number of logger function calls stored (though will remove oldest entries if value is reached to append new ones at the end). The size was a bit arbitrary but big enough to capture most recent calls. I can do a follow up to investigate an ideal size but i think is good threshold to start testing on the field

murki: for reference, if you want to use the preconfigbuffer as inspiration: it's bounded not only by count but also by memory size (in bytes), and the current limits are [512, and 1 MB](https://github.com/bitdriftlabs/shared-core/blob/e12fd292f88ca2901a87ccf0ad1ca4c9ff0402ea/bd-logger/src/builder.rs#L190-L191)

FranAguilera: awesome thanks, will match current size for now

PR #409, opened 2025-06-02
[Android] Gradle test app - Minor visual tweaks

FranAguilera:
Only a few minor UI tweak to ease Fatal Issue Reporting testing (before always need to scroll down to app exit button). Also fixed dropdown selector for `Log Message` type

### DIFF
--- a/platform/jvm/gradle-test-app/src/main/java/io/bitdrift/gradletestapp/FirstFragment.kt
+++ b/platform/jvm/gradle-test-app/src/main/java/io/bitdrift/gradletestapp/FirstFragment.kt
@@ -16,6 +16,7 @@ import android.view.LayoutInflater
 import android.view.View
 import android.view.ViewGroup
 import android.widget.ArrayAdapter
+import android.widget.Spinner
 import android.widget.Toast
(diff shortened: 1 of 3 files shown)


platform/jvm/gradle-test-app/src/main/java/io/bitdrift/gradletestapp/FirstFragment.kt
```diff
@@ -131,15 +132,8 @@ class FirstFragment : Fragment() {
 
         binding.textviewFirst.text = Logger.sessionId
 
-        val items = LogLevel.values().map { it.name }.toTypedArray()
-        (binding.logLevelItems as? MaterialAutoCompleteTextView)?.setSimpleItems(items)
-        binding.logLevelItems.setText(LogLevel.INFO.name, false)
```
FranAguilera, line 136: This wasn't rendering properly on my android 15 device

image

murki: I think this would happen if you navigate to another screen and then back to the main one since state wasn't persisted properly anywhere

PR #415, opened 2025-06-03
[Android] Harden AppExitInfo logic

FranAguilera:
## Prior context

- BIT-5474 
- [Conversation](https://bitdrift.slack.com/archives/C0891AS6H9Q/p1748902136838469)
- [Investigation doc](https://docs.google.com/document/d/1hxhVkwt8LB5YP-bpDz1RT9_DUqwOnUq3zkf60-rBPuY/edit?tab=t.0#heading=h.i35k1euojnjk) 

## Summary

- Hardens logic at `LatestAppExitInfoProvider` in order to only get the latest exit info related to the app process
- Add new error logs for Empty and early return when process state summary cannot be retreived
- Add WebView with multiprocess config to Gradle demo app

## Verification 

1. Navigate to a WebView and force kill the app from this view, next app launch should report ApplicationExitInfo info on next timeline event. [See session](https://timeline.bitdrift.dev/session/5bf6c062-1e5d-430b-8377-f133e5b68b22?utm_source=sdk&pages=1&utilization=0&expanded=-411963397190064195)

2. Click on OOM crash, Navigate to WebView, wait for the app to crash. Next app launch should show the OOM crash on ApplicationExit Info. [See session](https://timeline.bitdrift.dev/session/5bf6c062-1e5d-430b-8377-f133e5b68b22?utm_source=sdk&pages=1&utilization=0&expanded=8631404110426693078)

### DIFF
--- a/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/exitinfo/LatestAppExitInfoProvider.kt
+++ b/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/exitinfo/LatestAppExitInfoProvider.kt
@@ -8,6 +8,7 @@ package io.bitdrift.capture.reports.exitinfo
 
 import android.annotation.TargetApi
 import android.app.ActivityManager
+import android.app.Application
 import android.app.ApplicationExitInfo
 import android.os.Build
 import io.bitdrift.capture.reports.binformat.v1.ReportType
@@ -21,12 +22,21 @@ internal object LatestAppExitInfoProvider : ILatestAppExitInfoProvider {
         try {
             // a null packageName means match all packages belonging to the caller's process (UID)
             // pid should be 0, a value of 0 means to ignore this parameter and return all matching records
-            // maxNum should be 1, The maximum number of results to be returned, as we need only the last one
-            val latestExitReasons = activityManager.getHistoricalProcessExitReasons(null, 0, 1)
-            return if (latestExitReasons.isEmpty()) {
+            // maxNum should be 0, this will return the list of all last exists at the time
+            val latestKnownExitReasons =
+                activityManager
+                    .getHistoricalProcessExitReasons(null, 0, 0)
+            val matchingProcessReason =
+                latestKnownExitReasons
+                    .firstOrNull {
+                        it.processName == Application.getProcessName()
+                    }
+            return if (latestKnownExitReasons.isEmpty()) {
                 LatestAppExitReasonResult.Empty
+            } else if (matchingProcessReason == null) {
+                LatestAppExitReasonResult.ProcessNameNotFound
             } else {
-                LatestAppExitReasonResult.Valid(latestExitReasons.first())
+                LatestAppExitReasonResult.Valid(matchingProcessReason)
             }
         } catch (error: Throwable) {
             return LatestAppExitReasonResult.Error(
--- a/platform/jvm/capture/src/test/kotlin/io/bitdrift/capture/AppExitLoggerTest.kt
+++ b/platform/jvm/capture/src/test/kotlin/io/bitdrift/capture/AppExitLoggerTest.kt
@@ -22,6 +22,8 @@ import io.bitdrift.capture.common.RuntimeFeature
 import io.bitdrift.capture.events.lifecycle.AppExitLogger
 import io.bitdrift.capture.fakes.FakeBackgroundThreadHandler
 import io.bitdrift.capture.fakes.FakeLatestAppExitInfoProvider
+import io.bitdrift.capture.fakes.FakeLatestAppExitInfoProvider.Companion.DEFAULT_ERROR
+import io.bitdrift.capture.fakes.FakeLatestAppExitInfoProvider.Companion.FAKE_EXCEPTION
 import io.bitdrift.capture.fakes.FakeLatestAppExitInfoProvider.Companion.SESSION_ID
 import io.bitdrift.capture.fakes.FakeLatestAppExitInfoProvider.Companion.TIME_STAMP
 import io.bitdrift.capture.fakes.FakeMemoryMetricsProvider
@@ -33,10 +35,15 @@ import io.bitdrift.capture.reports.jvmcrash.ICaptureUncaughtExceptionHandler
 import io.bitdrift.capture.utils.BuildVersionChecker
 import org.junit.Before
 import org.junit.Test
+import org.junit.runner.RunWith
 import org.mockito.ArgumentMatchers.anyInt
+import org.robolectric.RobolectricTestRunner
+import org.robolectric.annotation.Config
 import java.io.IOException
 import java.nio.charset.StandardCharsets
 
+@RunWith(RobolectricTestRunner::class)
+@Config(sdk = [30])
 class AppExitLoggerTest {
     private val logger: LoggerImpl = mock()
     private val activityManager: ActivityManager = mock()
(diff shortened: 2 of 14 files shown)


platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/FatalIssueReporter.kt
```diff
@@ -170,7 +170,7 @@ internal class FatalIssueReporter(
     private fun persistLastExitReasonIfNeeded(appContext: Context) {
         val activityManager: ActivityManager =
             appContext.getSystemService(Context.ACTIVITY_SERVICE) as ActivityManager
-        val lastReasonResult = latestAppExitInfoProvider.get(activityManager)
+        val lastReasonResult = latestAppExitInfoProvider.get(activityManager, appContext)
```
murki, line 173: I don't love that we're passing the entire context around when all we need is just the packageName string

murki: but by looking at this code I think that ship has sailed

platform/jvm/capture/src/test/kotlin/io/bitdrift/capture/AppExitLoggerTest.kt
```diff
@@ -227,7 +274,14 @@ class AppExitLoggerTest {
...
             put("_app_exit_description", "test-description")
-            put("_app_exit_description", "test-description")
+            put("_memory_class", "1")
+        }.toFields()
+
+    private fun buildEmptyReasonFieldMap(): Map<String, FieldValue> =
+        buildMap {
+            put("_app_exit_source", "ApplicationExitInfo")
+            put("_app_exit_reason", "EMPTY_REASON")
+            put("_app_exit_description", "ApplicationExitInfo returned an empty list of reasons")
```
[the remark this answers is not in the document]
FranAguilera: yup, already fixed

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/exitinfo/LatestAppExitInfoProvider.kt
```diff
@@ -17,16 +18,24 @@ import io.bitdrift.capture.reports.binformat.v1.ReportType
...
-            // a null packageName means match all packages belonging to the caller's process (UID)
-            // pid should be 0, a value of 0 means to ignore this parameter and return all matching records
-            // maxNum should be 1, The maximum number of results to be returned, as we need only the last one
-            val latestExitReasons = activityManager.getHistoricalProcessExitReasons(null, 0, 1)
-            return if (latestExitReasons.isEmpty()) {
+            // Passing the current app package name (`packageName`) is critical to filter only app exits for the integrated app
+            // maxNum defined as 0 to get the latest list of issues.
+            val latestExitReason =
+                activityManager
+                    .getHistoricalProcessExitReasons(null, 0, 0)
```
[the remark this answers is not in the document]
FranAguilera: yup, that and only match of the current app process name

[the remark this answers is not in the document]
murki: quickly skimming through the code I'm not certain that's true, `mExitInfos` seems to be the "real" source of truth, and `mTmpInfoList2` just a local copy, my guess to avoid concurrent modification exceptions etc

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/exitinfo/LatestAppExitInfoProvider.kt
```diff
@@ -17,16 +18,24 @@ import io.bitdrift.capture.reports.binformat.v1.ReportType
...
+                    .sortedByDescending { it.timestamp }
+                    .firstOrNull {
+                        it.processName == appContext.packageName
```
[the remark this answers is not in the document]
murki: I don't think that will work since I'm pretty sure what's happening here is that the process with package name X is forking a new process which means the relationship between package:process is 1:N
for example these are some relevant logs printed after the lyft rider app launches
```
I  Loading com.google.android.webview version 136.0.7103.125 (code 710312533)
D  Forked child process 11835
I  Start proc 11835:com.google.android.webview:sandboxed_process0:org.chromium.content.app.SandboxedProcessService0:0/u0i26 for  {me.lyft.android.dev/org.chromium.content.app.SandboxedProcessService0:0}
```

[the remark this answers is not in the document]
murki: Also remember @FranAguilera that we tested at lyft using both calls:
```kotlin
val latestExitReasons = activityManager.getHistoricalProcessExitReasons(null, 0, 0)
val latestExitReasonsV2 = activityManager.getHistoricalProcessExitReasons(application.applicationContext.packageName, 0, 0)
```

and both lists returned the same values (with `com.google.android.webview:sandboxed_process0` at location 0)

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/FatalIssueReporter.kt
FranAguilera: Will refactor in here BIT-5478

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/exitinfo/LatestAppExitInfoProvider.kt
```diff
@@ -17,16 +18,24 @@ import io.bitdrift.capture.reports.binformat.v1.ReportType
...
+            // packageName: Must be the one from the current application
+            // pid: Should be 0, a value of 0 means to ignore this parameter and return all matching records
+            // maxNum: Should be 0, this will return the list of all last exists at the time
+            val latestExitReason =
+                activityManager
+                    .getHistoricalProcessExitReasons(appContext.packageName, 0, 0)
+                    .firstOrNull {
+                        it.processName == appContext.packageName
+                    }
+            return if (latestExitReason == null) {
```
murki, line 35: if might be good to log something if the method return values but no process matches the packagename

FranAguilera: we are logging at AppExitLogger here. Also added test for it

image

murki: but that's misleading since we don't know if the list returned empty, or if we just couldn't find a matching packageName == processName. I think the distinction is important since we don't have guarantees that they will always be >0 matches

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/FatalIssueReporter.kt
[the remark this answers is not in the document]
FranAguilera: yeah,though i see this as the "latest" known reason for app termination. I can clarify a bit on the naming

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/exitinfo/LatestAppExitInfoProvider.kt
FranAguilera: I can rename Empty to Unknown given that the new logic tries to find the first matching item

PR #417, opened 2025-06-03
Pick right Xcode from mac_ci_setup

Reflejo:


ci/mac_ci_setup.sh
```diff
@@ -3,7 +3,7 @@
 set -e
 
 # https://github.com/actions/runner-images/blob/main/images/macos/macos-14-Readme.md#xcode
-sudo xcode-select --switch /Applications/Xcode_15.4.app
+sudo xcode-select --switch /Applications/Xcode_16.2.app
```
murki, line 6: we should update the link above since we're running `macos-15`  https://github.com/actions/runner-images/blob/main/images/macos/macos-15-Readme.md#xcode ?

murki (approved):

PR #419, opened 2025-06-04
[Android] AppExit polish follow up

FranAguilera:
## What

Follow up from https://github.com/bitdriftlabs/capture-sdk/pull/415 where we simplify logic at LatestAppExitInfoProvider and error message generation

## Verification

Verified on multiprocess demo app. See session here


platform/jvm/capture/src/test/kotlin/io/bitdrift/capture/LatestAppExitInfoProviderTest.kt
```diff
@@ -59,7 +70,10 @@ class LatestAppExitInfoProviderTest {
...
         val exitReason = latestAppExitInfoProvider.get(activityManager)
 
-        assertThat(exitReason is LatestAppExitReasonResult.Empty).isTrue()
+        assertResult<LatestAppExitReasonResult.Error>(
+            exitReason,
+            EXIT_REASON_EMPTY_LIST_MESSAGE,
+        )
     }
 
     @Ignore("TODO(FranAguilera): BIT-5484. This works on gradle with Roboelectric. Fix on bazel")
```
FranAguilera, line 79: Verified on gradle. Will do a follow up PR to fix BIT-5484

murki (approved):

PR #423, opened 2025-06-05
Add a mechanism for "publishing" internal APIs for use across our libraries

kstenerud:
A client library loads the API class "BD_InternalAPI_Capture" and then calls "exposeAPI" on it to link up at runtime. If the class fails to load, or the requested API doesn't exist, it downgrades gracefully to a no-op rather than crashing.

See `InternalAPI.h` for a more thorough description + example client.

The unit tests also include `SampleClient.[hm]`, which is a full example of a client library.

### DIFF
--- a/bazel/bitdrift_swift_test.bzl
+++ b/bazel/bitdrift_swift_test.bzl
@@ -58,3 +58,47 @@ def bitdrift_mobile_objc_test(name, srcs, data = [], deps = [], tags = [], visib
         tags = tags,
         visibility = visibility,
     )
+
+def bitdrift_mobile_swift_objc_test(name, srcs_swift, srcs_objc, hdrs_objc, data = [], deps = [], tags = [], use_test_host = False, repository = "", visibility = []):
+    test_swift_lib_name = name + "_swift_lib"
+    test_objc_lib_name = name + "_objc_lib"
+    swift_library(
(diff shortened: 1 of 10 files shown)


kattrali (commented):
This makes sense overall. I think a good next step is probably to pull the branch into the RN library and try it out, to test both the ergonomics and the functionality work as expected. wdyk, @snowp? (Since I think you were looking into having internal APIs for RN?)

Another set of eyes on the build system changes also would help, maybe @Reflejo?

snowp:
Yeah the way this would be hooked up for the RN use case would be to expose some variant of https://github.com/bitdriftlabs/capture-sdk/blob/main/platform/swift/source/Logger.swift#L248 via this new interface such that the RN implementation at https://github.com/bitdriftlabs/capture-es/blob/main/packages/react-native/ios/BdReactNative.mm can expose a RN function to log a span log. The capture-es RN library consumes capture-sdk via Cocapods but there there is some way to pull in a branch? 

The code on the RN to create a span was never merged since we ran into this issue, but I have some old code that implemented it here https://github.com/bitdriftlabs/capture-es/pull/88/commits/cb61ea2e12cd110ae9a41e13f0ce9600239ef177 that we could bring back to call the RN function whenever we have time to put all this together

PR #428, opened 2025-06-09
feat: add public interface to the sdk version

kattrali:
Adds `Capture.Logger.sdkVersion()` for both Apple and Android platforms (using `+[CAPLogger sdkVersion]` for Objective-C)

Picked `sdkVersion` over `version` to avoid shadowing [`NSObject.version()`](https://developer.apple.com/documentation/objectivec/nsobject-swift.class/version())


murki (approved):
neat

PR #430, opened 2025-06-09
fix(ios): attribute first non-empty stack as crash diagnostic cause

kattrali:
avoid case where no stack is attributed as the first thread has zero frames while others do

fixed BIT-5531


murki (approved):
sure, let's try it

PR #441, opened 2025-06-16
Align FatalIssueReporter mechanism on both platforms

FranAguilera:
Just a minor rename to bring parity with Android [FatalIssueMechanism definition](https://github.com/bitdriftlabs/capture-sdk/blob/95a288a4471996869fa8271fccb386d2eac3bfbc/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/FatalIssueMechanism.kt#L24)

### DIFF
--- a/platform/swift/source/Capture.swift
+++ b/platform/swift/source/Capture.swift
@@ -124,7 +124,7 @@ extension Logger {
                 MXMetricManager.shared.add(reporter)
                 return .initialized(.monitoring)
                 #endif
-            case .customConfig:
+            case .integration:
                 return CustomConfigIssueReporter.processFiles()
             }
         }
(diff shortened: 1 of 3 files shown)


FranAguilera:
> I'm not sure calling it just `integration` works when there is already an `Integration` interface in the library. For example, this makes enabling the reporting integration something like:
> 
> ```swift
> logger.enableIntegrations([.crashReporting(.integration)])
> ```

uops missed this one, fixing soon

FranAguilera:
actually nvm, missed this comment here too. This will be a bit confusing with iOS "integrations". Probably we should rename Android at some point to keep it consistent (maybe something to imply vendor more directly)

PR #454, opened 2025-06-21
[ANDR] Copy jank framedata before using it

murki:
While prepping for my droidcon talk I realized we never copy the volatile data like [the docs say](https://developer.android.com/topic/performance/jankstats#reporting) (and like I suggested here: https://github.com/bitdriftlabs/capture-sdk/pull/222#discussion_r1962292003) and then we use it for potentially async operations. This could explain some of the invalid values we sometimes get.

Screenshot 2025-06-23 at 10 05 44 AM


FranAguilera:
> ### 📦 APK Size Report
> Metric	Size (KB)
> Baseline	3602
> Current	3610
> Difference	8
> > ❌ APK size increased by 8 KB.

there is seems to be an issue with the caching on github

PR #473, opened 2025-06-30
[Android] Remove com.github.michaelbull

FranAguilera:
## What

Removes dependency on `com.github.michaelbull ` for 'CaptureResult'

**_NOTE: These updated CaptureResult will be used on https://github.com/bitdriftlabs/capture-sdk/pull/474_**

## Verification

Verified on this [session](https://timeline.bitdrift.dev/session/96eb0c60-11a3-4c93-9dec-fa87a79e8d52?utm_source=sdk&pages=1&utilization=0&expanded=2904188329063557441)

### DIFF
--- a/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/Capture.kt
+++ b/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/Capture.kt
@@ -283,22 +283,11 @@ object Capture {
         @JvmStatic
         fun createTemporaryDeviceCode(completion: (CaptureResult<String>) -> Unit) {
             logger()?.also {
-                it.createTemporaryDeviceCode { result ->
-                    mainThreadHandler.run {
-                        when (result) {
-                            is CaptureResult.Success -> {
-                                completion(CaptureResult.Success(result.value))
-                            }
-                            is CaptureResult.Failure -> {
-                                completion(CaptureResult.Failure(result.error))
-                            }
-                        }
-                    }
+                it.createTemporaryDeviceCode {
+                    mainThreadHandler.run { completion(it) }
                 }
             } ?: run {
-                mainThreadHandler.run {
-                    completion(CaptureResult.Failure(SdkNotStartedError))
-                }
+                mainThreadHandler.run { completion(CaptureResult.Failure(SdkNotStartedError)) }
             }
         }
 


platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/Models.kt
```diff
@@ -10,15 +10,34 @@ package io.bitdrift.capture
...
+    ) : CaptureResult<V>()
+
+    /**
+     * Represents a failed operation result within the Capture SDK.
+     *
+     * @property error the [Error] describing the reason for the failure.
+     */
+    data class Failure(
+        val error: Error,
+    ) : CaptureResult<Nothing>()
```
murki, line 31: why `Nothing` ?

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/Models.kt
```diff
@@ -57,4 +76,4 @@ sealed class ApiError(
 /**
  * Represents a failed operation due to the SDK not being started
  */
-data object SdkNotStartedError : Error("SDK not started")
+class SdkNotStartedError : Error("SDK not started")
```
murki, line 79: shouldn't be class

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/Models.kt
```diff
@@ -10,15 +10,34 @@ package io.bitdrift.capture
 /**
  * A monad for modeling success or failure operations in the Capture SDK.
  */
-typealias CaptureResult<V> = com.github.michaelbull.result.Result<V, Error>
+sealed class CaptureResult<out V> {
```
murki, line 13: you should take a look at the library we used, it can be helpful as a reference implementation for things like this:
> The Result type is modelled as an inline value class. This achieves zero object allocations on the happy path.

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/Models.kt
murki: oh yeah I guess that way you force them to use our Errors and not define them themselves?

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/Models.kt
murki: I'm not saying we have to do it that way but it's a neat property of the library

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/Models.kt
FranAguilera: yup that's right

murki (approved):
I'm overall supportive of this change.

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/Capture.kt
```diff
@@ -284,11 +283,22 @@ object Capture {
...
+                                completion(CaptureResult.Failure(result.error))
+                            }
+                        }
+                    }
                 }
             } ?: run {
-                mainThreadHandler.run { completion(Err(SdkNotStartedError)) }
+                mainThreadHandler.run {
+                    completion(CaptureResult.Failure(SdkNotStartedError))
+                }
```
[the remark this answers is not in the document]
FranAguilera: uops my bad, this went through a series of refactors, will revert back to this

PR #474, opened 2025-07-01
[Android] Add startAsync + deprecate start

FranAguilera:
## What

Depends on https://github.com/bitdriftlabs/capture-sdk/pull/473

Resolves BIT-5721

1) Deprecate `Capture.Logger.start` for `Capture.Logger.startAsync`
2) Capture.startAsync has a mandatory parameter `completion: ((CaptureResult<Unit>) -> Unit)` to get notified when startAsync has completed

With this customers can get notified in cases there is an error while starting the SDK or when the SDK is fully initialized for async start

## Verification

[Session](https://timeline.bitdrift.dev/s/96eb0c60-11a3-4c93-9dec-fa87a79e8d52?utm_source=sdk)

Below demo with a simulated delay for creating LoggerImpl instance

https://github.com/user-attachments/assets/b4d3cf20-9e8b-45d9-92af-b2d996d4e277


platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/Capture.kt
```diff
@@ -262,7 +288,7 @@ object Capture {
          */
         @JvmStatic
         val deviceId: String?
-            get() = logger()?.deviceId
+            get() = (logger() as? LoggerImpl)?.deviceId
```
FranAguilera, line 291: this keeps behavior as before and only returning device id from LoggerImpl instance only

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/Capture.kt
```diff
@@ -177,35 +187,82 @@ object Capture {
...
+         * a [android.content.ContentProvider]
+         */
         @Synchronized
         @JvmStatic
         @JvmOverloads
-        internal fun start(
+        fun startAsync(
             apiKey: String,
             sessionStrategy: SessionStrategy,
+            completion: ((CaptureResult<Unit>) -> Unit),
```
murki, line 217: per kotlin conventions we should make this a trailing lambda https://kotlinlang.org/docs/lambdas.html#passing-trailing-lambdas

FranAguilera:
will reopen when ready again

PR #482, opened 2025-07-03
[Android] Adding startAsync and deprecate start

FranAguilera:
## What

Resolves BIT-5748, BIT-5721

Following feedback on https://github.com/bitdriftlabs/capture-sdk/pull/474

Add `Capture.Logger.startAsync` (while deprecating `Capture.Logger.start`) which will initialize the sdk in a dedicated background thread to prevent any negative impact on app launch if this method is called from main thread

Below an example init call:

```
Capture.Logger.startAsync(
      apiKey,
      apiUrl,
      sessionStrategy,
      configuration
  ) {
      if (it is CaptureResult.Success) {
          Timber.i("Bitdrift Logger initialized with session_url=${it.value.sessionId}")
      }
  }
```

_**NOTE: BIT-5747 Will do a follow up on capture-es to update react with async call**_ 

 ## Verification

### `Logger.startAsync` - GradleTestApp

1. Verified without additional delay and timber log after initial async call https://timeline.bitdrift.dev/session/9eab3521-535c-4b59-bac7-2f1123c622d6?utm_source=sdk&pages=1&utilization=0&expanded=-4628434477939081327

2. Verified on this session with a simulated delay while creating LoggerImpl

https://timeline.bitdrift.dev/s/c0ec3e49-b3bc-4c31-b9b8-10b0b8666e87?utm_source=sdk

https://github.com/user-attachments/assets/d9e8b218-aecf-459b-8772-ee49c97df570

image

### `Logger.startAsync` - External app via local maven

Verified on this session

https://timeline.bitdrift.dev/session/c5756b42-c2f9-4dba-a7d6-b358e37fc929?utm_source=sdk&pages=1&utilization=0&expanded=-2904482817931087679

### `Logger.start()` - GradleTestApp

Verified on this session

https://timeline.bitdrift.dev/s/79073999-7cd3-44df-baa0-446d6c04bd54?utm_source=sdk

image


platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/Capture.kt
```diff
@@ -220,21 +241,155 @@ object Capture {
...
+            } else {
+                Log.w(LOG_TAG, "Multiple attempts to start Capture")
+            }
+        }
+
         /**
          * The Id for the current ongoing session.
          * It's equal to `null` prior to the start of Capture SDK.
          */
+        @Deprecated(
```
FranAguilera, line 364: Deprecating these nullable fields in favor of completionResult callback passed into startAsync

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/Capture.kt
```diff
@@ -220,21 +241,155 @@ object Capture {
...
+                dateProvider,
+                apiUrl,
+                CaptureJniLibrary,
+            )
+        }
+
+        @Synchronized
+        @JvmStatic
+        @JvmOverloads
+        internal fun startAsync(
```
FranAguilera, line 292: Pending to add similar tests for internal start

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/Models.kt
```diff
@@ -77,3 +77,16 @@ sealed class ApiError(
...
 data object SdkNotStartedError : Error("SDK not started")
+
+/**
+ * Represents a Sdk Start due to having an invalid context
+ */
+data object StartAsyncWithNullContext :
+    Error("SDK not started due to null context upon startAsync call")
+
+/**
+ * Represents a Sdk Start due to having an exception thrown while calling startAsync
```
FranAguilera, line 88: Same

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/Capture.kt
```diff
@@ -220,21 +241,155 @@ object Capture {
...
+                                apiKey = apiKey,
+                                apiUrl = apiUrl,
+                                fieldProviders = fieldProviders,
+                                dateProvider = dateProvider ?: SystemDateProvider(),
+                                configuration = configuration,
+                                sessionStrategy = sessionStrategy,
+                                bridge = bridge,
+                                fatalIssueReporter = fatalIssueReporter,
+                                preInitLogFlusher = preInitInMemoryLoggerRef.get(),
+                            )
```
kstenerud, line 339: Inside LoggerImpl's init method, it calls the flusher to flush the pre-init logger, but we don't clear `preInitInMemoryLoggerRef` until the finally clause. Wouldn't that leave a hole for another thread to get the pre-init log after the flush, but before switching the instance to the regular logger (effectively losing that log)?

What if we didn't pass the flusher into the logger, but instead built the logger fully out here, did an atomic switch of the reference, and then finally did a flush of the now outside-inaccessible pre-init logger?

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/Capture.kt
```diff
@@ -512,14 +704,17 @@ object Capture {
...
 
         /**
          * Used for testing purposes.
          */
+        @VisibleForTesting
         internal fun resetShared() {
             default.set(LoggerState.NotStarted)
         }
+
+        private fun getInternalLogger(): ILogger? = preInitInMemoryLoggerRef.get() ?: logger()
```
kstenerud, line 718: This feels like it should just be a single reference that we change over once the real logger is ready.

kattrali (approved):

FranAguilera:
Pending to validate react native

FranAguilera:
Will update CaptureTree to avoid using Capture.logger() upon construction

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/Capture.kt
```diff
@@ -517,16 +704,40 @@ object Capture {
...
+
+        /**
+         * Ease testing of Capture.Logger calls
+         *
+         * @param logger The ILogger instance to be verified
+         *
+         */
+        @VisibleForTesting
+        @JvmStatic
+        fun setLoggerForTesting(logger: ILogger) {
```
FranAguilera, line 718: Not really sold into this, still thinking best approach to ease testing without relying on Capture.logger()

PR #195, opened 2025-01-23
[ANDR] Set kotlin to 2.0.0

FranAguilera:
Kotlin 2.1.0 is causing build errors on React apps. Set kotlin to previous [2.0.0](https://kotlinlang.org/docs/whatsnew20.html#current-k2-compiler-limitations)

Related internal thread https://bitdrift.slack.com/archives/C058ZD2M2CQ/p1737665327205719

NOTE:  To get KOTLIN_COMPILER_SHA. 

1) Download [kotlin-compiler-2.0.0.zip](https://github.com/JetBrains/kotlin/releases/download/v2.0.0/kotlin-compiler-2.0.0.zip)
2) shasum -a 256 kotlin-compiler-2.0.0.zip

PR #483, opened 2025-07-04
Remove some overrides and update rules

Reflejo:


MODULE.bazel
```diff
@@ -12,9 +12,9 @@ LIFECYCLE_VERSION = "2.8.7"
 
 COMPOSE_VERSION = "1.7.0"
 
-KOTLIN_COMPILE_VERSION = "2.0.0"
+KOTLIN_COMPILE_VERSION = "2.1.21"
```
FranAguilera, line 15: just recalled this old thread https://bitdrift.slack.com/archives/C058ZD2M2CQ/p1737665327205719 and past revert https://github.com/bitdriftlabs/capture-sdk/pull/195

Will validate if this still breaks react

FranAguilera: Prior error on sample repo app with 2.1.0

image

FranAguilera: Same issue with this PR

FranAguilera (changes requested):
Regarding kotlin version comments from above

.bazelrc
```diff
@@ -21,6 +21,8 @@ build --features=swift.opt_uses_osize
 build --features=swift.opt_uses_wmo
 build --features=swift.remap_xcode_path
 build --features=swift.use_global_module_cache
+build --features=swift.emit_swiftinterface
```
murki, line 24: This option is currently making the flatbuffer dep fail, not sure what we need this for

FranAguilera (approved):

PR #488, opened 2025-07-08
Add recordAppOpen platform API

kattrali:


platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/LoggerImpl.kt
```diff
@@ -370,6 +370,10 @@ internal class LoggerImpl(
         CaptureJniLibrary.setSleepModeEnabled(this.loggerId, sleepMode == SleepMode.ACTIVE)
     }
 
+    override fun recordAppOpen() {
```
FranAguilera, line 373: Let's sync offline, I thought this should happen automatically and not something we expose to customers

FranAguilera: - Prototype discussed before https://github.com/bitdriftlabs/capture-sdk/pull/490/files

- [Related doc](https://docs.google.com/document/d/13OchgAihTDbQZLjI3s4Gk0WA_KMPARI2KDKOdaXVxos/edit?tab=t.oe08nmtigaaw#heading=h.wrwzgo3mr4rp)

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/events/lifecycle/AppLifecycleListenerLogger.kt
```diff
@@ -104,6 +106,12 @@ internal class AppLifecycleListenerLogger(
         }
     }
 
+    private fun emitMetricsForEvent(event: Lifecycle.Event) {
+        if (event == Lifecycle.Event.ON_START || event == Lifecycle.Event.ON_RESUME) {
```
murki, line 110: we should confirm these are indeed called for all scenarios we care about. I seem to remember these 2 in particular to be tied to having an active Activity in the app

PR #499, opened 2025-07-10
Initial KSCrash integration

kstenerud:
This PR integrates parts of KSCrash to capture fatal exceptions (Mach, Signal, NSException, C++ exception) along with the extra metadata that KSCrash can record, in order to augment any MetricKit crash reports we receive.

To start, this will only include thread names.

When a `MXCrashDiagnostic` report is received, we attempt to enhance the report like so:

```objc
NSDictionary *enhancedReport = [BitdriftKSCrashWrapper enhancedMetricKitReport:event.dictionaryRepresentation];
```

We use `BitdriftKSCrashWrapper`, which is a thin wrapper around `BitdriftKSCrashHandler` that is gated by the preprocessor directive `BITDRIFT_OMIT_KSCRASH` (which causes the wrapper to no-op for everything). This gives us an easy way to produce an xcframework that doesn't include this functionality.

`BitdriftKSCrashWrapper` must be configured and started in order for report enhancement to work (we do this in `Capture.swift`):

```objc
+ (bool)configureWithBasePath:(NSURL *)basePath;
+ (bool)startCrashReporter;
```

Once the crash reporter is started, it will record the next fatal exception to disk on crash, and reload the data on the next launch, keeping it ready for comparison against any MetricKit report passed in to `[BitdriftKSCrashWrapper enhancedMetricKitReport:]`. Any threads in the MetricKit report that it can match to a corresponding thread in the KSCrash report that has an associated name will have an extra key "name" added to it in the enhanced report.

The KSCrash report is stored in a crash-resilient format that can be (partially) reloaded even if we crash while writing it.

Rather than forking our own KSCrash, we simply grab a subset of the code from the KSCrash git repo directly:

```bazel
new_git_repository(
    name = "kscrash",
    branch = "66eb948d9c2f8d2c46e818f62b6a4e016035aef8",
    build_file = "//bazel:BUILD.kscrash",
    remote = "https://github.com/kstenerud/KSCrash.git",
)
```

And then apply custom namespacing to avoid linker conflicts with other KSCrash-based reporters:

```bazel
    defines = [
        "KSCRASH_NAMESPACE=_bitdrift",
        ],
```

See also [bazel/BUILD.kscrash](https://github.com/bitdriftlabs/capture-sdk/pull/499/files#diff-5b61a24cb9fda91eeab2396a8d6beb2fd9fc316f8aa9f2d0be262c8066239bbeR1)


Reflejo (approved):
I did a very high level pass, there are some null checks here and there I would double check but approving for now to unblock

murki (changes requested):
Let's discuss with the team what it would take to write as much as possible of this code in rust instead

kstenerud:
Holding off on any further changes until we have a rust discussion.

kattrali (commented):
still working through some of it but here is a few small initial comments

PR #500, opened 2025-07-16
[Android] Adding Architecture, supported ABIs to reports

FranAguilera:
## What

Resolves BIT-5777, BIT-5246

Adding architecture, cpuAbis to Reports

## Verification

[timeline session](https://timeline.bitdrift.dev/session/87c72347-ff73-4bb3-be27-b7c6052f0b57?utm_source=sdk&pages=1&utilization=0&expanded=2241004306869461929 )

image

### DIFF
--- a/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/attributes/ClientAttributes.kt
+++ b/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/attributes/ClientAttributes.kt
@@ -39,6 +39,14 @@ internal class ClientAttributes(
             }
         }
 
+    val supportedAbis: List<String>
+        get() = Build.SUPPORTED_ABIS.toList()
+
+    val architecture: String
+        get() {
+            return supportedAbis.firstOrNull() ?: "unknown"
+        }
+
     @Suppress("SwallowedException")
     private val packageInfo: PackageInfo? =
         try {
@@ -63,6 +71,8 @@ internal class ClientAttributes(
             // A positive integer used as an internal version number.
             // This number helps determine whether one version is more recent than another.
             "_app_version_code" to appVersionCode.toString(),
+            // The current architecture e.g. (arm64-v8a)
+            "_architecture" to architecture,
         )
 
     private fun isForeground(): String {
--- a/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/processor/FatalIssueReporterProcessor.kt
+++ b/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/processor/FatalIssueReporterProcessor.kt
@@ -13,7 +13,9 @@ import com.google.flatbuffers.FlatBufferBuilder
 import io.bitdrift.capture.BuildConstants
 import io.bitdrift.capture.attributes.ClientAttributes
 import io.bitdrift.capture.reports.binformat.v1.AppBuildNumber
+import io.bitdrift.capture.reports.binformat.v1.Architecture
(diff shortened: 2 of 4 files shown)


platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/processor/FatalIssueReporterProcessor.kt
```diff
@@ -150,9 +152,18 @@ internal class FatalIssueReporterProcessor(
         timestampMillis: Long,
     ): Int {
         val duration = timestampMillis.toDuration(DurationUnit.MILLISECONDS)
-        // TODO: BIT-5246 Add device info
+        val cpuAbis =
+            createCpuAbisVector(
+                builder,
+                clientAttributes.supportedAbis.map { builder.createString(it) }.toIntArray(),
```
FranAguilera, line 158: @kattrali I'm not sure if this is ok to return the raw value (e.g. `armeabi`) or if needs the specific fbs value (`arm32`)

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/processor/FatalIssueReporterProcessor.kt
```diff
@@ -150,9 +152,18 @@ internal class FatalIssueReporterProcessor(
...
+            )
         DeviceMetrics.startDeviceMetrics(builder)
         DeviceMetrics.addPlatform(builder, Platform.Android)
+        DeviceMetrics.addArch(builder, clientAttributes.architectureAsFbs())
```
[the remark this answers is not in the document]
FranAguilera: but why not reuse the one provided on clientAttributes? it seems more fragmentation just to get to the same place. happy to sync offline later

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/attributes/ClientAttributes.kt
```diff
@@ -39,6 +39,14 @@ internal class ClientAttributes(
             }
         }
 
+    val supportedAbis: List<String>
+        get() = Build.SUPPORTED_ABIS.toList()
```
murki, line 43: can this be cached or is it expected to change?

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/attributes/ClientAttributes.kt
```diff
@@ -39,6 +39,14 @@ internal class ClientAttributes(
...
+
+    val architecture: String
+        get() {
+            return supportedAbis.firstOrNull() ?: "unknown"
```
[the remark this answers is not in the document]
FranAguilera: This is what they mention on that API

```
 /**
     * An ordered list of ABIs supported by this device. The most preferred ABI is the first
     * element in the list.
     *
     * See {@link #SUPPORTED_32_BIT_ABIS} and {@link #SUPPORTED_64_BIT_ABIS}.
     */
       public static final String[] SUPPORTED_ABIS = getStringList("ro.product.cpu.abilist", ",");
```

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/attributes/ClientAttributes.kt
murki: actually if it's from the Build object should already be static

FranAguilera: ah that's right, will keep it

PR #506, opened 2025-07-17
enable relr for android build

snowp:
Per https://android.googlesource.com/platform/ndk/+/master/docs/BuildSystemMaintainers.md#relr-and-relocation-packing this is safe now that we are targeting min version 24


FranAguilera (approved):

FranAguilera (changes requested):
Actually let me do some test on some devices before merging

FranAguilera (approved):

PR #508, opened 2025-07-17
[andr] Move expensive operations on start to a bg thread

murki:
Tackling mainly:
- [x] `monitorNetworkType()` (went from 9.70% to 0.79%)
- [x] `writeSdkStartLog()` (went from 2.27% to 0%)

This should make `start()` perform 11.18% faster

|Before|After|
|---|---|
|Screenshot 2025-07-17 at 9 59 14 AM (1)|Screenshot 2025-07-17 at 12 36 30 PM|

Session example testing the `SDKConfigured` log with the right `network_type` value: https://timeline.bitdrift.dev/session/fda02916-f33f-4365-8262-94d6132b2266?utm_source=sdk&pages=1&utilization=0&h=3450357454420969661
Screenshot 2025-07-17 at 1 17 22 PM

After the change this is the threading structure:
```
connectivityManager operations called in thread=io.bitdrift.capture.background-thread-worker
updateNetworkType() callback called in thread=io.bitdrift.capture.background-thread-worker
updateNetworkType() callback called in thread=ConnectivityThread
writeSdkStartLog() called in thread=io.bitdrift.capture.background-thread-worker
```

### DIFF
--- a/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/attributes/NetworkAttributes.kt
+++ b/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/attributes/NetworkAttributes.kt
@@ -41,11 +41,13 @@ import android.telephony.TelephonyManager.NETWORK_TYPE_UNKNOWN
 import androidx.core.content.ContextCompat
 import io.bitdrift.capture.providers.FieldProvider
 import io.bitdrift.capture.providers.Fields
+import java.util.concurrent.ExecutorService
 import java.util.concurrent.atomic.AtomicReference
 
 @SuppressLint("MissingPermission")
 internal class NetworkAttributes(
     private val context: Context,
+    executor: ExecutorService,
 ) : ConnectivityManager.NetworkCallback(),
     FieldProvider {
     @SuppressLint("InlinedApi")
@@ -76,7 +78,9 @@ internal class NetworkAttributes(
     private val connectivityManager = context.getSystemService(Context.CONNECTIVITY_SERVICE) as ConnectivityManager
 
     init {
-        monitorNetworkType()
+        executor.execute {
+            monitorNetworkType()
+        }
     }
 
     override fun invoke(): Fields =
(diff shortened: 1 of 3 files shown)


platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/attributes/NetworkAttributes.kt
```diff
@@ -41,11 +41,13 @@ import android.telephony.TelephonyManager.NETWORK_TYPE_UNKNOWN
 import androidx.core.content.ContextCompat
 import io.bitdrift.capture.providers.FieldProvider
 import io.bitdrift.capture.providers.Fields
+import java.util.concurrent.ExecutorService
 import java.util.concurrent.atomic.AtomicReference
 
 @SuppressLint("MissingPermission")
 internal class NetworkAttributes(
     private val context: Context,
+    executor: ExecutorService,
```
FranAguilera, line 50: nit: Can you pass directly CaptureDispatchers.CommonBackground so we are explicit about it?

murki: this is done so on tests we can pass a more appropriate one

FranAguilera: There is CaptureDispatchers.setTestExecutorService for tests as well

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/attributes/NetworkAttributes.kt
```diff
@@ -76,7 +79,9 @@ internal class NetworkAttributes(
     private val connectivityManager = context.getSystemService(Context.CONNECTIVITY_SERVICE) as ConnectivityManager
```
FranAguilera, line 79: Maybe this connectivityManager be lazy now given this will be accessed from background thread

FranAguilera (approved):

FranAguilera:
Oh actually missed that we are changing the reported start thread via writeSdkStartLog. We should store the caller thread name and pass it into `writeSdkStartLog` otherwise will report incorrectly

image

FranAguilera (changes requested):
Regarding thread start name

PR #509, opened 2025-07-17
optimize gradle .so build for size

snowp:
This is now comparable to the Bazel output @ 1.2MB when stripped + zipped


murki:
@snowp do you want to give this another try now that you migrated the gradle plugin?

.cargo/config.toml
```diff
@@ -8,6 +8,17 @@ rustflags = ["-Csplit-debuginfo=unpacked"]
...
 
+# Android target configuration
+[target.'cfg(target_os="android")']
+rustflags = [
+  "-C",
+  "link-arg=-Wl,--retain-symbols-file=platform/jvm/jni_symbols.lds",
+  "-C",
+  "link-arg=-Wl,--build-id",
+  "-C",
+  "link-arg=-Wl,--pack-dyn-relocs=android",
```
snowp, line 19: not sure if enabling this in dev matters?

.cargo/config.toml
```diff
@@ -8,6 +8,17 @@ rustflags = ["-Csplit-debuginfo=unpacked"]
 [target.x86_64-unknown-linux-gnu]
 rustflags = ["-C", "link-arg=-fuse-ld=lld"]
 
+# Android target configuration
+[target.'cfg(target_os="android")']
+rustflags = [
+  "-C",
+  "link-arg=-Wl,--retain-symbols-file=platform/jvm/jni_symbols.lds",
```
snowp, line 15: doing this in dev is good because it will let us catch issues with the symbols file early

snowp:
I gave this another shot and it seems to do the right thing, now debug vs release uses the different configurations.

I also fixed stripping in the gradle-test-app, seems like we weren't doing this since the default ndk version wasn't installed so I set it to match the NDK version used by the capture lib

platform/jvm/capture/build.gradle.kts
```diff
@@ -88,12 +88,31 @@ android {
...
+    }
+
     module = "../.."
     targetDirectory = "./target"
     // Default set for local dev on ARM-based macos
     targets = arrayListOf("arm64")
     // enable 16 KB ELF alignment on Android to support API 35+
-    extraCargoEnv = mapOf("RUSTFLAGS" to "-C link-args=-Wl,-z,max-page-size=16384")
+    extraCargoEnv = mapOf(
+      "RUSTFLAGS" to "-C link-args=-Wl,-z,max-page-size=16384",
```
murki, line 113: we should be able to remove this once we update NDK version to 28 https://developer.android.com/guide/practices/page-sizes#compile-r28

PR #518, opened 2025-07-24
[Android] Add putAllNonNull extensions

FranAguilera:
## Problem

Resolves BIT-5914

Prevents potential runtime crashes given a map could be passed from Java call site with null values.

For example this call will produce a NPE on java site

```
IllegalStateException exception = new IllegalStateException();
HashMap<String, String> fields = new HashMap<>();
fields.put("exception_message", exception.getMessage());
Capture.Logger.logDebug(
        fields,
        exception,
        () -> "Sample message"
        );
```

## Solution

1. Add extension method that internally will check for key/value being null for the case listed above
2. Added a custom detekt rule to prevent further non-safe usages of mutableMap.putAll(fields) in the future

## Verification
1. Verify there are no runtime crashes with the snippet below added into the GradleTestApp. 

```
IllegalStateException exception = new IllegalStateException();
HashMap<String, String> fields = new HashMap<>();
fields.put("exception_message", exception.getMessage());
Capture.Logger.logDebug(
        fields,
        exception,
        () -> "Sample message"
        );
```
        
2. Verify that the custom rules only detekt issues on mutableMap.putAll calls

Will paste screenshot

### DIFF
--- a/platform/jvm/capture/build.gradle.kts
+++ b/platform/jvm/capture/build.gradle.kts
@@ -42,6 +42,8 @@ dependencies {
     testImplementation(libs.androidx.test.core)
     testImplementation(libs.robolectric)
     testImplementation(libs.mockwebserver)
+
+    detektPlugins(project(":detekt-custom-rules"))
 }
 
 protobuf {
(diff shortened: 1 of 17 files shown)


FranAguilera:
Will repurpose for custom detekt rules which can be useful later on

PR #519, opened 2025-07-24
[Android] Prevent interopt crash for toFields

FranAguilera:
## What

Resolves BIT-5914

Prevents potential NPE if a map with null key/value is constructed via Java and passed into Capture.Logger calls

## Verification

Verified with JUnit test with existing call and updated one. See [initial commit for demo](https://github.com/bitdriftlabs/capture-sdk/pull/519/commits/6302e3cb4ee1d6012f3d3e5ac9baa0d49323c947)

Temporary add this method call to GradleTestApp. App should not crash

image

### DIFF
--- a/platform/jvm/capture/BUILD.bazel
+++ b/platform/jvm/capture/BUILD.bazel
@@ -1,5 +1,4 @@
 load("@rules_java//java:defs.bzl", "java_lite_proto_library")
-load("@rules_java//java:java_library.bzl", "java_library")
 load("@rules_jvm_external//:defs.bzl", "artifact")
 load("@rules_kotlin//kotlin:jvm.bzl", "kt_jvm_library")
 load("@rules_proto//proto:defs.bzl", "proto_library")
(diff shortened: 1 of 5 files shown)


FranAguilera:
Re triggering emulator

platform/jvm/capture/BUILD.bazel
```diff
@@ -15,6 +16,14 @@ java_lite_proto_library(
     deps = [":tombstone_proto"],
 )
 
+java_library(
+    name = "java_test_utils",
+    srcs = glob(
+        include = ["src/test/**/*.java"],
```
murki, line 22: shouldn't we just instead exand the existing `bitdrift_kt_android_local_test` rule to include java files?

murki (approved):

platform/jvm/capture/BUILD.bazel
FranAguilera: Ended updating the bitdrift_kt_android_local_test macro itself

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/providers/FieldProvider.kt
```diff
@@ -118,8 +118,19 @@ internal fun Pair<String, String>.toField(): Pair<String, FieldValue> = Pair(thi
 
 /**
  * Converts a Map<String, String> into a List<Field>.
+ *
```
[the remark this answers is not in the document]
FranAguilera: Adding it here https://github.com/bitdriftlabs/capture-sdk/pull/520/files

PR #516, opened 2025-07-22
[Android] Add native load duration to Capture SDK start log

FranAguilera:
## What

Resolves BIT-5887 (See [related context](https://bitdrift.slack.com/archives/C058ZD2M2CQ/p1753180710509879?thread_ts=1752696297.259089&cid=C058ZD2M2CQ) )

Adds another duration field for native load to the `SDKConfigured` log. This will help breaking down on production the overall `Capture.Logger.start` time

## Verification

Verified on these sessions:

- [gradle test app session](https://timeline.bitdrift.dev/session/e04fbe48-5922-4f28-b4f3-f207b2f34c18?utm_source=sdk&pages=1&utilization=0&expanded=-7944603258184882928) with the GradleTest sample app

image

image

- [bazel release test app](https://timeline.bitdrift.dev/session/5f2bd465-a012-4fe0-b7fb-b374df283eb2?utm_source=sdk&pages=1&utilization=0&expanded=5496365138251104009)

image

image

PR #526, opened 2025-07-25
[Android] Improvements on SDKConfigured duration

FranAguilera:
## What

1. For SDKConfigured log we are only measuring duration_ms within internal init constructor block. In this case will be missing duration of other init logic within Capture.Logger.start call (so for example the overall duration_ms doesn't include fatal issue reporting if configured, evaluation and instantiation of constructor arguments or setting the atomic internal variables)

2. We are now loading the native library as part of start call to allow easier measurement and avoid duplicating init calls at LoggerImpl and ErrorHanlder

3. We were incorrectly reporting n/a when crash reporting wasn't configured. This aligns with [suggestion made](https://github.com/bitdriftlabs/capture-sdk/pull/516#discussion_r2222305956) by @kattrali  to not report that duration key at all if not valid

4. There is no need to report initialization thread for fatal issue given we are now initializing inside start call

So in summary SDKConfigured will have these fields related to duration/init

| Status | Field | Context
| -------| -----| --------|
| Existing | _duration_ms | The total duration of Capture.Logger.start internals. This now measures whole instantiation of LoggerImpl and start internals
| Existing | _capture_start_thread | This is the thread name where Capture.Logger.start took place 
| Removed | _fatal_reporter_init_thread | This is redundant given is included as part of _capture_start_thread
| Changed | _fatal_issue_reporting_duration_ms | Only applicable now when customer has configured crash reporting 
| New | _logger_build_duration_ms | The total duration of native logger instance (e.g. LoggerImpl instantiation)
| Existing | _logger_build_duration_ms | The total duration of native so load (e.g.  System.loadLibrary call)

## Verification

Compare the duration now accounts for whole start internals. 

| | main | PR |
| ----------| ------|------|
| Without previous ANR report | image | image |
| With previous ANR report | image | image|

### DIFF
--- a/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/ErrorHandler.kt
+++ b/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/ErrorHandler.kt
@@ -15,12 +15,6 @@ import io.bitdrift.capture.common.ErrorHandler
  * unexpected failures.
  */
 internal class ErrorHandler : ErrorHandler {
-    init {
-        // Make sure this is loaded before we call into CaptureJniLibrary. Generally this should
-        // already have been loaded.
-        CaptureJniLibrary.load()
-    }
-
     /**
      * Handles a single error.
      * @param detail a string identifying the action that resulted in an error.
--- a/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/LoggerImpl.kt
+++ b/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/LoggerImpl.kt
@@ -55,7 +55,6 @@ import okhttp3.HttpUrl
 import java.util.UUID
 import kotlin.time.Duration
 import kotlin.time.DurationUnit
-import kotlin.time.measureTime
 
 typealias LoggerId = Long
 internal typealias InternalFieldsList = List<Field>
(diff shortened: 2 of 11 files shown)


platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/ErrorHandler.kt
```diff
@@ -15,12 +15,6 @@ import io.bitdrift.capture.common.ErrorHandler
  * unexpected failures.
  */
 internal class ErrorHandler : ErrorHandler {
-    init {
-        // Make sure this is loaded before we call into CaptureJniLibrary. Generally this should
-        // already have been loaded.
-        CaptureJniLibrary.load()
```
murki, line 21: this race condition doesn't cause issues anymore?

FranAguilera:

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/LoggerImpl.kt
```diff
@@ -107,166 +106,156 @@ internal class LoggerImpl(
...
-                                clientAttributes,
-                                networkAttributes,
-                                deviceAttributes,
-                            ),
-                        errorHandler = errorHandler,
-                        customFieldProviders = fieldProviders,
-                    )
+        setUpInternalLogging(context)
+
+        this.sessionUrlBase =
```
FranAguilera, line 111: sorry about the noise, main changes here due to removal of measureTime duration

murki (approved):

PR #528, opened 2025-07-25
Make CaptureTree configurable

murki:
Allow consumer to extend and therefore modify the behavior of the Tree, specifically around filtering, e.g.
```kotlin
private class MyTree : CaptureTree() {
    override fun isLoggable(tag: String?, priority: Int): Boolean {
        return priority > Log.VERBOSE
    }
}
```


FranAguilera (commented):
Few suggestions:

1. If we make this open, we should probably lock down visibility of `override fun log` so no one unintentionally changes the current implementation that relies on capture

```
    final override fun log(
        priority: Int,
        tag: String?,
        message: String,
        t: Throwable?,
    ) {
        logger?.log(toCaptureLogLevel(priority), extractFields(tag), t) { message }
    }
```

2. Another alternative to avoid opening this, could be to pass in the allowed log levels set as a constructor parameter. Though this will be less flexible in case consumer wants to do something more custom while overriding`fun isLoggable(tag: String?, priority: Int)`

murki:

PR #541, opened 2025-07-31
[Android] Update integration app to rely only on maven central

FranAguilera:
Update integration gradle test app since version 0.18.0 we are publishing to maven central

Timeline session https://timeline.bitdrift.dev/session/1a1f05c0-ff42-4856-ac9a-b873b9fb5f73?utm_source=sdk&pages=1&utilization=0&expanded=-2496588997681936027


gradle/app/build.gradle
```diff
@@ -1,17 +1,16 @@
 plugins {
     id 'com.android.application'
     id 'org.jetbrains.kotlin.android'
-    id 'io.bitdrift.capture-plugin'
```
kstenerud, line 4: I think we need this for the symbol upload gradle task?

FranAguilera: Good catch, added back and validated localy

kstenerud (approved):

PR #546, opened 2025-08-01
[Android] Adding back native crash processing

FranAguilera:
TBF


platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/persistence/FatalIssueReporterStorage.kt
```diff
@@ -22,6 +22,7 @@ internal class FatalIssueReporterStorage(
         val fileName = "${terminationTimeStampInMilli}_${mapToReadableType(reportType)}_${UUID.randomUUID()}.cap"
         val outputFile = File(destinationDirectory, fileName)
         outputFile.writeBytes(data)
+        // [BIT-5823] Pending to call new native call to indicate to shared-core layer that processing/storage is done
```
FranAguilera, line 25: fyi @kattrali one of the spots where we could call the new native function to signal processing is done

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/Configuration.kt
```diff
@@ -12,14 +12,40 @@ import io.bitdrift.capture.replay.SessionReplayConfiguration
 /**
  * A configuration object representing the feature set enabled for Capture.
  * @param sessionReplayConfiguration The resource reporting configuration to use. Passing `null` disables the feature.
- * @param enableFatalIssueReporting When set to true wil capture Fatal Issues automatically [JVM crash, ANR, etc] and without requiring
- * any external 3rd party library integration
+ * @param fatalIssueReportingMode When set to [FatalIssueReportingMode.JVM_AND_ANR] wil capture Fatal Issues
+ * automatically [JVM crash + ANR] and without requiring any external 3rd party library integration.
+ * If [FatalIssueReportingMode.JVM_ANR_AND_NATIVE] will additionally report native crashes
```
FranAguilera, line 17: need to polish these docs a bit more

FranAguilera:
> ### 📦 Bazel APK(x86_64) Size Report
> Metric	Size (KB)
> Baseline	3393
> Current	3365
> Difference	-28
> > ✅ Bazel APK(x86_64) size decreased by 28 KB.

heads ups: this is because this PR is on top of https://github.com/bitdriftlabs/capture-sdk/pull/551

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/FatalIssueReporter.kt
```diff
@@ -38,6 +39,7 @@ import kotlin.time.measureTime
  * Handles internal reporting of crashes
  */
 internal class FatalIssueReporter(
+    private val fatalIssueReportingMode: FatalIssueReportingMode = FatalIssueReportingMode.DISABLED,
```
FranAguilera, line 42: will think about this a bit more, probably don't need to pass it here and once this is constructed only JVM_ANR and JVM_ANR_AND_NATIVE should be applicable

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/FatalIssueReportingMode.kt
```diff
@@ -0,0 +1,44 @@
...
+         */
+        object JvmAndAnr : Enabled()
+
+        /**
+         * Enables reporting of:
+         * - JVM crashes (uncaught exceptions)
+         * - ANR (Application Not Responding) events
+         * - Native crashes (NDK)
+         */
+        object JvmAnrAndNative : Enabled()
```
murki, line 42: maybe we should consider calling it NDK? It's less ambiguos and keeps the format of 3-letter acronyms like the others

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/FatalIssueReportingMode.kt
```diff
@@ -0,0 +1,44 @@
...
+     */
+    sealed class Enabled : FatalIssueReportingMode() {
+        /**
+         * Enables reporting of:
+         * - JVM crashes (uncaught exceptions)
+         * - ANR (Application Not Responding) events
+         *
+         * Native (NDK) crash reporting is not enabled in this mode.
+         */
+        object JvmAndAnr : Enabled()
```
murki, line 34: nit/ also we should consider dropping the `And` here and there to make it more readable

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/Configuration.kt
```diff
@@ -8,18 +8,22 @@
...
+ * - [FatalIssueReportingMode.Enabled.JvmAnrAndNative]: Enables reporting of JVM crashes, ANRs,
+ * and native (NDK) crashes
  * @param sleepMode SleepMode.ACTIVE if Capture should initialize in minimal activity mode
  */
 data class Configuration
     @JvmOverloads
     constructor(
         val sessionReplayConfiguration: SessionReplayConfiguration = SessionReplayConfiguration(),
-        val enableFatalIssueReporting: Boolean = false,
+        val fatalIssueReportingMode: FatalIssueReportingMode = FatalIssueReportingMode.Disabled,
```
murki, line 27: thinking about this more, I actually think we should just add another boolean for native, if only because we expect it to be temporary, this way we don't break current customers, and eventually when we remove the temporary native flags, all customers that already have `enableFatalIssueReporting = true` would automatically be opted-in by default.

So basically, why go through the effort of creating a complex sealed class, and breaking current customers if we'd eventually delete that sealed class?

what do you think?

murki (approved):
lgtm - except perhaps for the boolean configuration comment

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/Configuration.kt
FranAguilera: I agree, that was my additional thought but wasn't sure if this was going to be a temporary option. Given that, will revert back to additional flag to avoid breaking changes

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/processor/FatalIssueReporterProcessor.kt
```diff
@@ -59,9 +61,15 @@ internal class FatalIssueReporterProcessor(
                     )
                 }
 
-                ReportType.NativeCrash -> {
-                    // TODO(FranAguilera): BIT-5823 use NativeCrashProcessor once async processing is ready
-                    null
+                fatalIssueType == ReportType.NativeCrash && fatalIssueReportingMode == FatalIssueReportingMode.Enabled.JvmAnrAndNative -> {
+                    NativeCrashProcessor.process(
```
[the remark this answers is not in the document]
FranAguilera: I have added these earlier, just realized naming is off. Will update after moving to single boolean

PR #547, opened 2025-08-01
use cargo-ndk to build .so in gradle project

snowp:


platform/jvm/capture/build.gradle.kts
```diff
@@ -1,8 +1,9 @@
 plugins {
+    alias(libs.plugins.android.library)
+
     // The rust android gradle plugin needs to go first
     //  see: https://github.com/mozilla/rust-android-gradle/issues/147
```
murki, line 5: comment is now obsolete

platform/jvm/capture/build.gradle.kts
```diff
@@ -97,26 +83,14 @@ android {
...
-}
-
-// workaround bug in rust-android-gradle plugin that causes .so to not be available on app launch
-// see: https://github.com/mozilla/rust-android-gradle/issues/118#issuecomment-1569407058
-tasks.whenTaskAdded {
-    if ((this.name == "mergeDebugJniLibFolders" || this.name == "mergeReleaseJniLibFolders")) {
-        this.dependsOn("cargoBuild")
-        this.inputs.dir(layout.buildDirectory.dir("rustJniLibs/android"))
-    }
+    targetDirectory = "./target"
```
murki, line 90: how is it able to find the target here? doesn't this run inside capture-sdk/platform/jvm ?

murki (approved):
I mean if it works, I'm happy to give it a try. Specially if we can use the build type conditionals for adding customer flags and building different archs

examples/README.md
```diff
@@ -72,7 +72,13 @@ To install without Homebrew you can follow these instructions: [https://docs.azu
 
 Setup the environment, open Android Studio.
 
-1. Select `Open...` and select folder `/capture-sdk/platform/jvm`. This should automatically kick off the Gradle sync process.
+1. Install `cargo-ndk` which is used to build the Rust shared library.
+
+  ```
+  cargo install cargo-ndk --version 3.4.0
```
murki, line 78: better update our ci jobs in this same pr as well

platform/jvm/capture/build.gradle.kts
snowp: `module` specifies ../.., so this becomes ../../target

.github/workflows/android.yaml
```diff
@@ -294,6 +304,9 @@ jobs:
     - name: Check Licenses for modules
       working-directory: ./platform/jvm
       run: ./gradlew replay:checkLicense common:checkLicense capture:checkLicense
+    - name: Set up local.properties, required for cargo-ndk
```
murki, line 307: weird, do they require any vars to be set there?

snowp: doesn't seem like it https://github.com/willir/cargo-ndk-android-gradle/issues/23

PR #551, opened 2025-08-04
use new async report monitor on iOS/Android

kattrali:
depends on https://github.com/bitdriftlabs/shared-core/pull/203

testing:

* ensure reports still sent
* match reports to correct session ID, even when changed next launch
* match reports to correc timestamp
* ensure fatal issue init stats still added to launch log message

Additional Android verification:
* Verify that only one "AppExit" event with crash data is recorded . [Session](https://timeline.bitdrift.dev/session/85c44671-0bba-40ee-a41f-1a2f0e3fa457?utm_source=sdk&pages=1&utilization=0&expanded=-1899214116432239192)

image

* Trigger ANR via gradle test app. Overall `_duration_ms` should be similar to cold launches without prior ANR report. [Session](https://timeline.bitdrift.dev/session/85c44671-0bba-40ee-a41f-1a2f0e3fa457?utm_source=sdk&pages=1&utilization=0&expanded=4918995890820301650)

image

* Trigger native crash with changes added in https://github.com/bitdriftlabs/capture-sdk/pull/546. Capture SDK start time is now the same as without prior crash reports. [Session](https://timeline.bitdrift.dev/session/06a586d1-2552-4a84-af6a-d671d092dd71?utm_source=sdk&pages=1&utilization=0&expanded=-5042773049862532567)

image

image

Resolves BIT-5823


FranAguilera (approved):
Native changes lgtm. Added Android changes on https://github.com/bitdriftlabs/capture-sdk/pull/551/commits/b4ceb18e26fd9763489f9847ccb4712294f72005

platform/jvm/capture/src/test/kotlin/io/bitdrift/capture/reports/FatalIssueReporterTest.kt
```diff
@@ -52,13 +57,14 @@ class FatalIssueReporterTest {
...
     fun initialize_whenBuiltInMechanism_shouldInitCrashHandlerAndFetchAppExitReason() {
-        fatalIssueReporter.initBuiltInMode(appContext, clientAttributes)
+        fatalIssueReporter.initBuiltInMode(appContext, clientAttributes, completedReportsProcessor)
 
         verify(captureUncaughtExceptionHandler).install(eq(fatalIssueReporter))
         verify(latestAppExitInfoProvider).get(any())
         fatalIssueReporter.fatalIssueReporterStatus.assert(
             FatalIssueReporterState.BuiltIn::class.java,
         )
+        verify(completedReportsProcessor).processCrashReports()
```
FranAguilera, line 67: @kattrali Added a quick verification here

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/FatalIssueReporter.kt
```diff
@@ -65,7 +70,10 @@ internal class FatalIssueReporter(
                                 clientAttributes,
                             )
                         captureUncaughtExceptionHandler.install(this)
-                        persistLastExitReasonIfNeeded(appContext)
+                        backgroundThreadHandler.runAsync {
+                            persistLastExitReasonIfNeeded(appContext)
+                            completedReportsProcessor.processCrashReports()
```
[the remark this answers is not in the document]
murki: Reading this comments makes me even more nervous: https://github.com/bitdriftlabs/shared-core/pull/203/files#r2254864269

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/FatalIssueReporter.kt
```diff
@@ -149,7 +157,7 @@ internal class FatalIssueReporter(
     private fun buildDefaultReporterStatus(): FatalIssueReporterStatus =
         FatalIssueReporterStatus(
             FatalIssueReporterState.NotInitialized,
-            mechanism = FatalIssueMechanism.None,
+            mechanism = FatalIssueMechanism.BuiltIn,
```
murki, line 160: why changing the default from None to BuiltIn?

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/Capture.kt
```diff
@@ -547,6 +550,8 @@ object Capture {
 
             default.set(LoggerState.Started(loggerImpl))
 
+            fatalIssueReporter?.initBuiltInMode(appContext, clientAttributes, loggerImpl)
```
murki, line 553: why are we handling all the construction and initialization of the fatalIssueReporter outside of LoogerImpl? Seems like it deviates from the current pattern. Seems like it could all be done inside the Logger instantiation like everything else

murki (approved):
android lgtm, just one question that might come from my misunderstanding of the rust code

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/FatalIssueReporter.kt
FranAguilera: Given call order have changed now, will need the passed IFatalIssueReporter instance into LoggerImpl which creates AppExitLogger to have this BuiltIn type to avoid double "AppExit" emission

This FatalIssueMechanism is still around in case there is a need to bring back vendor. I'm tracking BIT-6043 as a follow up if we decide that vendor won't be used in any foreseeable future

PR #552, opened 2025-08-05
[andr] Allow Android Studio to run tests locally by suppressing some warnings

murki:
We're using `@Suppress("INVISIBLE_REFERENCE")` in tests which are reported as warnings. We are setting `allWarningsAsErrors = true` in the kotlin compiler but we're opting to bypass these particular warnings via the `-Xdont-warn-on-error-suppression` flag.

Not ideal but we'll have to make due until kotlin has first-class support for "friend" modules to access `internal`, see: https://youtrack.jetbrains.com/issue/KT-67920/K2-Allow-suppression-of-INVISIBLEREFERENCE-errors

### DIFF
--- a/platform/jvm/capture/build.gradle.kts
+++ b/platform/jvm/capture/build.gradle.kts
@@ -1,3 +1,6 @@
+import org.jetbrains.kotlin.gradle.dsl.JvmTarget
+import org.jetbrains.kotlin.gradle.dsl.KotlinVersion
+
 plugins {
     alias(libs.plugins.android.library)
     alias(libs.plugins.rust.android)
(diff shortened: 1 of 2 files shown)


platform/jvm/capture/build.gradle.kts
```diff
@@ -57,12 +60,14 @@ android {
         targetCompatibility = JavaVersion.VERSION_1_8
     }
 
-    // TODO(murki): consider updating to using kotlin.compilerOptions {} block
```
murki, line 60: I took the opportunity to migrate these options as recommended https://kotlinlang.org/docs/gradle-compiler-options.html#migrate-from-kotlinoptions-to-compileroptions

PR #554, opened 2025-08-05
[DO NOT REVIEW YET] Exclude protobuf-javalite

FranAguilera:
Exclude protobuf-javalite from protobuf-kotlin-lite to prevent class duplication errors when consumers use protobuf-java.

This avoids transitive conflicts with apps relying on com.google.protobuf:protobuf-java and ensures compatibility without forcing consumers to exclude on their end

Kudos to @pyricau for the investigation and link to existing issue: https://github.com/protocolbuffers/protobuf/issues/8104


FranAguilera:
Will update our release notes to include the required exclusion for customers that have :protobuf-java

PR #559, opened 2025-08-07
Add applicatinIdSuffix to `Logger.start`

pyricau:
applicatinIdSuffix is a suffix that will be added to the representation of the application id (aka package name) in Bitdrift. This is useful if you have different build flavors with the same application id and you want a way to apply entire workflows only to a subset of those.


murki:
Closing in favor of #569

PR #562, opened 2025-08-07
[DO NOT REVIEW] Upload to maven central action

FranAguilera:
Testing for now


.github/workflows/android.yaml
```diff
@@ -361,11 +362,193 @@ jobs:
           emulator-options: -no-snapshot-save -no-window -accel on -gpu swiftshader_indirect -noaudio -no-boot-anim -camera-back none
           script: ./ci/verify_android.sh
 
+
+  test_maven_central_upload:
```
FranAguilera, line 366: Heads up: This won't be living here, only adding to trigger the action

.github/workflows/android.yaml
```diff
@@ -361,11 +362,63 @@ jobs:
...
+          sudo apt-get install -y gnupg
+
+      - name: Download release artifacts
+        env:
+          GH_TOKEN: ${{ github.token }}
+          VERSION: 0.17.28
+          RELEASE_TAG: v0.17.28
+        run: |
+          ZIP_NAME="Capture-${VERSION}.android.zip"
+          DOWNLOAD_URL="https://github.com/bitdriftlabs/capture-sdk/releases/download/${RELEASE_TAG}/${ZIP_NAME}"
```
FranAguilera, line 391: This is not final, it's best to download for s3 bucket with md5 and sha already available

PR #569, opened 2025-08-08
Add `appIdSuffix` to `Logger.start` via `Configuration`

murki:
A slight modification of https://github.com/bitdriftlabs/capture-sdk/pull/559

`appIdSuffix` is a suffix that will be appended to the representation of the app_id (aka package name) in Bitdrift. This is useful if you have different build flavors with the same application id and you want a way to apply entire workflows only to a subset of those.


platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/attributes/ClientAttributes.kt
```diff
@@ -47,8 +48,9 @@ internal class ClientAttributes(
     @Suppress("SwallowedException")
     private val packageInfo: PackageInfo? by lazy {
         try {
-            context.packageManager.getPackageInfoCompat(appId)
-        } catch (e: Exception) {
+            // we intentionally not use appId+suffix to retain the original's packageInfo
+            context.packageManager.getPackageInfoCompat(context.packageName.orEmpty())
```
murki, line 52: @pyricau I opted for keeping the `app_version` of the original `app_id` (sans suffix), let me know what you think

pyricau (approved):

PR #595, opened 2025-08-25
bump shared core to bring in panic removal

mattklein123:
Fixes BIT-6163


platform/swift/source/src/session.rs
```diff
@@ -79,7 +80,9 @@ impl ActivityBasedStrategyCallbacks for SessionStrategy {
     // Safety: Since we receive `ActivityBasedStrategyCallbacks` as a typed protocol, we know that
     // it responds to `sessionIdChanged:` and will take a NSString.
     objc::rc::autoreleasepool(|| unsafe {
-      let session_id = make_nsstring(session_id);
+      let Ok(session_id) = make_nsstring(session_id) else {
+        return;
+      };
```
mattklein123, line 85: For this one I could make this return an error but would need to change the shared_core interface. Not sure if it's worth it. Let me know.

snowp (approved):

PR #603, opened 2025-08-28
support disabling crash reporting via runtime

kattrali:
Companion to https://github.com/bitdriftlabs/shared-core/pull/221

Checks for the new runtime config cache values before enabling crash detection and reporting

## todo

- [x] testing on iOS ([sample session](https://timeline.bitdrift.dev/session/2788265F-E585-41F9-A567-60A8C54D2F05))
- [x] testing on Android ([sample session](https://timeline.bitdrift.dev/session/200FA522-B6B5-42D9-8D94-D8BF197BEFA8))
- [x] update shared-core pin after companion pr merged

### DIFF
--- a/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/utils/ConfigCache.kt
+++ b/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/utils/ConfigCache.kt
@@ -37,21 +37,22 @@ internal object ConfigCache {
      * @throws CacheFormattingError if the file format does not match
      */
     @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
-    fun readValues(text: String): Map<String, Any> {
-        val values = HashMap<String, Any>()
-        for (line in text.split("\n")) {
-            val pair = line.split(",", limit = 2)
-            if (pair.size == 2) {
-                values[pair[0]] =
-                    when (pair[1]) {
-                        "true" -> true
-                        "false" -> false
-                        else -> pair[1]
-                    }
-            } else {
-                throw CacheFormattingError()
-            }
-        }
-        return values
-    }
+    fun readValues(text: String): Map<String, Any> =
+        text
+            .lines()
+            .map { line ->
+                val pair = line.split(",", limit = 2)
+                if (pair.size == 2) {
+                    Pair(
+                        pair[0],
+                        when (pair[1]) {
+                            "true" -> true
+                            "false" -> false
+                            else -> pair[1]
+                        },
+                    )
+                } else {
+                    throw CacheFormattingError()
+                }
+            }.associate { it.first to it.second }
 }


FranAguilera (commented):
Thanks for making the check internal to FatalIssueReporter. 

Main feedback is around moving the new call at FatalIssueReporter within runCaching.

Also another thing to consider is that if the config cache flag is disable we won't get reports of the on JVM crash event as the internal check at `AppExitLogger.onJvmCrash` only checks for configuration but it should also check now for the reporter config cache (we could use something similar prior to this [PR](https://github.com/bitdriftlabs/capture-sdk/pull/572/)) where a nullable FatalIssueReporter instance was passed in, and check the init status of that one

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/FatalIssueReporter.kt
```diff
@@ -64,48 +65,60 @@ internal class FatalIssueReporter(
...
-                                FatalIssueMechanism.BuiltIn,
-                            )
-                    }
+        if (fatalIssueReporterStatus.state != NotInitialized) {
+            Log.e(LOG_TAG, "Fatal issue reporting already being initialized")
+            return
+        }
+
+        val duration = TimeSource.Monotonic.markNow()
+        val sdkDirectory: String = SdkDirectory.getPath(appContext)
```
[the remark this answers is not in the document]
FranAguilera: Added here https://github.com/bitdriftlabs/capture-sdk/pull/607

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/FatalIssueReporter.kt
```diff
@@ -65,39 +66,56 @@ internal class FatalIssueReporter(
...
-                }
-            }.getOrElse {
-                logError(completedReportsProcessor, it)
+        if (fatalIssueReporterState != NotInitialized) {
+            Log.e(LOG_TAG, "Fatal issue reporting already being initialized")
+            return
+        }
+        val duration = TimeSource.Monotonic.markNow()
+        try {
+            if (!isFatalIssueReportingRuntimeEnabled(sdkDirectory)) {
```
FranAguilera, line 75: If we change the internals of `ConfigCache` (see comments below) we can move the `isFatalIssueReportingRuntimeEnabled` call into the existing `runCatching` keeping same behavior to fail if exception occurs

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/utils/ConfigCache.kt
```diff
@@ -0,0 +1,56 @@
...
+ */
+internal object ConfigCache {
+    /**
+     * Read configuration values from a file
+     *
+     * @param file File to read
+     *
+     * @return The values or Exception which occurred during parsing
+     */
+    fun readValues(file: File): Result<Map<String, Any>> =
```
FranAguilera, line 25: We could store the cached map here per file to perform caching after initial readValues call

e.g   private val cachedValues = mutableMapOf<File, Map<String, String>>()

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/utils/ConfigCache.kt
```diff
@@ -0,0 +1,56 @@
...
+            val pair = line.split(",", limit = 2)
+            if (pair.size == 2) {
+                values[pair[0]] =
+                    when (pair[1]) {
+                        "true" -> true
+                        "false" -> false
+                        else -> pair[1]
+                    }
+            } else {
+                return Result.failure(CacheFormattingError())
```
FranAguilera, line 51: Instead of returning a Result.failure we could throw directly so we can reuse existing runCatching block at FatalIssueReporter

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/utils/ConfigCache.kt
```diff
@@ -0,0 +1,56 @@
...
+            }
+        }
+        return Result.success(values)
```
FranAguilera, line 54: Same as above, instead of Result return a boolean directly

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/FatalIssueReporter.kt
```diff
@@ -162,6 +180,12 @@ internal class FatalIssueReporter(
         return FatalIssueDirectories(sdkDirectory, destinationDirectory)
     }
 
+    private fun isFatalIssueReportingRuntimeEnabled(sdkDirectory: String): Boolean {
+        val configFile = File(sdkDirectory, "reports/config.csv")
```
FranAguilera, line 184: If we let ConfigCache to throw directly and do the caching internally (see comments below) this could be simplified as

```
    private fun isFatalIssueReportingRuntimeEnabled(sdkDirectory: String): Boolean {
        val configFile = File(sdkDirectory, "reports/config.csv")
        val configMap = ConfigCache.readValues(configFile)
        return configMap["crash_reporting.enabled"]?.equals("true", ignoreCase = true) ?: false
    }
```

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/utils/ConfigCache.kt
```diff
@@ -0,0 +1,56 @@
...
+        }
+
+    /**
+     * Read configuration values from text
+     *
+     * @param text to parse
+     *
+     * @return The values or Exception which occurred during parsing
+     */
+    fun readValues(text: String): Result<Map<String, Any>> {
```
FranAguilera, line 39: Then this will become private as it doesn't need to be exposed anymore

platform/swift/source/reports/ReportConfigCache.swift
```diff
@@ -0,0 +1,33 @@
...
+// Use of this source code is governed by a source available license that can be found in the
+// LICENSE file or at:
+// https://polyformproject.org/wp-content/uploads/2020/06/PolyForm-Shield-1.0.0.txt
+
+/// Parse text for comma-delimited key/value pairs
+///
+/// - parameter text: text to parse
+///
+/// - returns: A mapping between keys and values, upon success
+package func readCachedValues(_ text: String) -> [String: Any]? {
```
FranAguilera, line 13: For parity would be good to match the recent Android implementation

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/utils/ConfigCache.kt
FranAguilera: I guess I was getting confused by the name which implied caching. I agree, if we don't need caching of the read values here at the moment please feel free to ignore

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/utils/ConfigCache.kt
FranAguilera: I thought we wanted a fail safe approach and not assume it's ok to continue init crash reporting if this flag cannot be read as we won't have a way to turn it of if that's the case.

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/FatalIssueReporter.kt
```diff
@@ -65,9 +66,21 @@ internal class FatalIssueReporter(
...
+            return
+        }
+
+        val duration = TimeSource.Monotonic.markNow()
+        runCatching {
+            var fileCheckPassed = true
+            if (!isFatalIssueReportingRuntimeEnabled(sdkDirectory)) {
+                fileCheckPassed = false
+                fatalIssueReporterState =
+                    FatalIssueReporterState.RuntimeDisabled
```
FranAguilera, line 80: optional nit: We could remove the new `fileCheckPassed` here and early return after setting the state. As if there is an exception thrown within `isFatalIssueReportingRuntimeEnabled` we'll jump directly into `getOrElse`

platform/jvm/capture/src/test/kotlin/io/bitdrift/capture/reports/FatalIssueReporterTest.kt
```diff
@@ -51,13 +53,40 @@ class FatalIssueReporterTest {
...
+        )
+    }
+
+    @Test
+    fun initialize_whenConfigNotPresent_shouldNotInit() {
+        configFile.delete()
+        fatalIssueReporter.init(appContext, sdkDirectory, clientAttributes, completedReportsProcessor)
+
+        fatalIssueReporter.fatalIssueReporterState.assert(
+            FatalIssueReporterState.InitializationFailed::class.java,
```
FranAguilera, line 84: nit: we could also add a verification call of `completedReportsProcessor` with the expected message/exception (e.g. `verify(completedReportsProcessor).onReportProcessingError(...)`)

FranAguilera (approved):
Thanks for updating. Overall lgtm, just one nit about verifying call of `completedreportsProcessor` on config error case and other optional nit for the init logic

PR #624, opened 2025-09-09
fix(android): remove misleading error log

kattrali:
this message should not be considered as an error and is otherwise reported within the sdk initialization to indicate state


platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/FatalIssueReporter.kt
```diff
@@ -83,7 +83,6 @@ internal class FatalIssueReporter(
                     return
                 }
             }.onFailure {
-                logError(completedReportsProcessor, it)
```
FranAguilera, line 86: With this removal will miss all detailed errors reported during fatal issue reporter init.

FranAguilera (changes requested):

PR #629, opened 2025-09-12
Cache `os_version` and add `osApiLevel`

FranAguilera:
Related to BIT-6360


FranAguilera:
> byt this logic should we also use by lazy on other things? like supported abis?

definitely, will update before merging

PR #632, opened 2025-09-12
include all binary images from android tombstones

snowp:
Instead of only including binary images with a valid build ID, synthesize a binary image for each mapped region referred to by the stack trace. This ensures that we have all the information we need for all frames, even if they don't have a build ID

### DIFF
--- a/platform/jvm/capture/src/test/kotlin/io/bitdrift/capture/reports/processor/NativeCrashProcessorTest.kt
+++ b/platform/jvm/capture/src/test/kotlin/io/bitdrift/capture/reports/processor/NativeCrashProcessorTest.kt
@@ -17,7 +17,7 @@ import java.io.ByteArrayOutputStream
 
 class NativeCrashProcessorTest {
     @Test
-    fun `dedupes by build-id and picks the lowest address memory map for report`() {
+    fun `populates buildId for binary images`() {
         val tombstone =
             TombstoneProtos.Tombstone
                 .newBuilder()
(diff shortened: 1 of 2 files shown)


platform/jvm/capture/src/test/kotlin/io/bitdrift/capture/reports/processor/NativeCrashProcessorTest.kt
```diff
@@ -31,34 +31,112 @@ class NativeCrashProcessorTest {
...
+                            },
+                    )
+                    it.addMemoryMappings(
+                        TombstoneProtos.MemoryMapping
+                            .newBuilder()
+                            .also {
+                                it.mappingName = "file1"
+                                it.buildId = "build-id"
+                                it.beginAddress = 150
+                                it.beginAddress = 199
```
FranAguilera, line 58: This probably should be endAddress

platform/jvm/capture/src/test/kotlin/io/bitdrift/capture/reports/processor/NativeCrashProcessorTest.kt
```diff
@@ -31,34 +31,112 @@ class NativeCrashProcessorTest {
...
                                     .newBuilder()
-                                    .setBuildId("build-id"),
+                                    .also {
+                                        it.fileName = "file1"
+                                        it.buildId = "build-id"
+                                        it.pc = 110
+                                    },
                             ).build(),
                     )
+                    it.addMemoryMappings(
```
FranAguilera, line 41: Nit: We could extract private methods here to avoid these test preparation duplication

PR #649, opened 2025-09-25
perform full LTO during Android .so build using rustc

snowp:
Previously we were building a cc_binary target which seemed to have an issue with the version of LLVM used that prevented LTO from working. Based on some comments it meant that we were not enabling LTO for this reason (despite flags that might indicate this).

Realizing that cargo was able to do this, by moving the final link to rust_shared_library we can use rustc for everything which simplified this and allows us to run fat LTO as part of the .so build. 

To get this working I needed to update the std rlib to be built with LTO compile flags.

In addition, this brings in Rust 1.90.0 for both Android and iOS.


platform/jvm/capture/build.gradle.kts
```diff
@@ -113,7 +113,7 @@ cargoNdk {
     targets = arrayListOf("arm64")
     // enable 16 KB ELF alignment on Android to support API 35+
     extraCargoEnv = mapOf(
-      "RUSTFLAGS" to "-C link-args=-Wl,-z,max-page-size=16384,--build-id",
+      "RUSTFLAGS" to "-C codegen-units=1 -C embed-bitcode -C lto=fat -C opt-level=z -C link-args=-Wl,-z,max-page-size=16384,--build-id",
```
murki, line 116: is this still required?

.bazelrc
```diff
@@ -71,16 +71,13 @@ build:release-common --config=abort-panic
...
-# but the ndk's lld won't support the output
-build:release-common --@rules_rust//rust/settings:lto='manual'
+build:release-common --@rules_rust//rust/settings:extra_rustc_flag='-Copt-level=z'
+build:release-common --@rules_rust//rust/settings:lto='fat'
 
 build:release-android --define android_strip_symbols=true
 build:release-android --android_platforms=@rules_android//:armeabi-v7a,@rules_android//:arm64-v8a,@rules_android//:x86,@rules_android//:x86_64
 build:release-android --config=release-common
-build:release-android --copt=-flto=thin --linkopt=-flto=thin
+# build:release-android --copt=-flto=thin --linkopt=-flto=thin
```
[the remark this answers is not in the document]
snowp: oops, no reason so I'll remove

platform/jvm/capture/build.gradle.kts
snowp: This is just for the cargo build so I figured I'd leave this in so that there is parity between all cargo and bazel. All the other stuff just targets the bazel pipeline so this is still necessary for the gradle build

snowp:
Seems like this breaks iOS, looking into it

PR #640, opened 2025-09-17
Client attributes perf improvements

FranAguilera:
### What

Follow up to https://github.com/bitdriftlabs/capture-sdk/pull/558

In this case we avoid creating new map upon each `invoke` call and use the priorly cache map with foreground/background event

### Benchmark Comparison

| Main | PR |
| ------| ------|
| image | image |
| 10999 allocs | 0 allocs |

PR #650, opened 2025-09-29
[DO NOT REVIEW YET] Remove DeviceAttributes

FranAguilera:
Follow up to https://github.com/bitdriftlabs/capture-sdk/pull/640

Pending to add verification


platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/attributes/NetworkAttributes.kt
```diff
@@ -70,18 +70,39 @@ internal class NetworkAttributes(
         context.getSystemService(Context.CONNECTIVITY_SERVICE) as ConnectivityManager
     }
 
+    private var cachedCarrier: String? = null
+    private var cachedRadioType: String? = null
+    private var cachedNetworkType: String? = null
+    private val cachedFields: MutableMap<String, String> by lazy { mutableMapOf() }
```
FranAguilera, line 76: This is intentional so the cached map is created on the same thread where invoke call happens

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/attributes/NetworkAttributes.kt
```diff
@@ -107,6 +129,7 @@ internal class NetworkAttributes(
         networkCapabilities: NetworkCapabilities,
     ) {
         updateNetworkType(networkCapabilities)
+        updateTelephonyAttributes()
```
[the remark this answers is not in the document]
FranAguilera: so far looks like it's occurring on each change on my local testing

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/attributes/NetworkAttributes.kt
```diff
@@ -122,6 +145,19 @@ internal class NetworkAttributes(
         )
     }
 
+    private fun updateTelephonyAttributes() {
+        val newCarrier = telephonyManager.networkOperatorName ?: "unknown"
+        val newRadioType = permissiveOperation({ radioType() }, READ_PHONE_STATE)
+
+        if (currentCarrier.get() != newCarrier) {
```
murki, line 152: why is this check necessary? can we just assign the values as read and skip the extra check?

FranAguilera: It’s just a small optimization to avoid calling `set` if the value hasn’t actually changed (this avoid unnecessary atomic writes)

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/attributes/NetworkAttributes.kt
```diff
@@ -133,29 +105,38 @@ internal class NetworkAttributes(
...
+        networkType: String? = null,
+        radioType: String? = null,
+    ) {
+        val updated =
+            currentFields.toMutableMap().apply {
+                carrier?.let { this[KEY_CARRIER] = it }
+                networkType?.let { this[KEY_NETWORK_TYPE] = it }
+                radioType?.let { this[KEY_RADIO_TYPE] = it }
+            }
+        currentFields = updated.toMap()
```
pyricau, line 139: You should double check, but I believe both `toMutableMap()` and `toMap()` always create a copy. If I'm right, given that `NetworkAttributes` is internal, and assuming the consuming code doesn't try to make any change to the map returned by `fun invoke(): Fields`, you could remove the `toMap()` here, save yourself a copy. That map will never be updated again

murki (approved):
after `toMap()` removal suggestion

FranAguilera:
Doing another round of verification. Will merge after that

PR #656, opened 2025-09-30
[Android] Set minSdk to 23

FranAguilera:
Setting minSdk to 23. This is inline with [androidx minSdk level](https://developer.android.com/jetpack/androidx/versions/all-channel) 

The most relevant changes:

- `JankStatsMonitor` to run again `sendJankFrameData` outside main thread <24
- `NetworkAttributes` to use `registerNetworkCallback` below <24 

### Verification

Verified on API 23 device. [Timeline session](https://timeline.bitdrift.dev/session/4d53088c-a1b6-4160-9092-34aba9bd82e9?utm_source=sdk&utilization=0&expanded=-787543766163269708)

### DIFF
--- a/platform/jvm/gradle-test-app/src/androidTest/java/io/bitdrift/gradletestapp/ComposeReplayTest.kt
+++ b/platform/jvm/gradle-test-app/src/androidTest/java/io/bitdrift/gradletestapp/ComposeReplayTest.kt
@@ -227,7 +227,7 @@ class ComposeReplayTest {
 
         val capture = verifyReplayScreen(viewCount = 7)
         assertThat(capture).contains(ReplayRect(ReplayType.SwitchOff, 42, 126, 84, 84))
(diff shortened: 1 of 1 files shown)


platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/attributes/NetworkAttributes.kt
```diff
@@ -88,7 +90,12 @@ internal class NetworkAttributes(
                 connectivityManager.activeNetwork?.let { network ->
                     updateNetworkType(connectivityManager.getNetworkCapabilities(network))
                 }
-                connectivityManager.registerDefaultNetworkCallback(this)
+                if (Build.VERSION.SDK_INT < Build.VERSION_CODES.N) {
```
murki: does this mean <23 devices will not receive networking info at all? sounds like we should fallback to the more expensive polling logic in these cases

PR #659, opened 2025-09-30
Print device code on android if app is debuggable

murki:
Counterpart to https://github.com/bitdriftlabs/capture-sdk/pull/658

### DIFF
--- a/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/LoggerImpl.kt
+++ b/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/LoggerImpl.kt
@@ -12,6 +12,7 @@ import android.app.ActivityManager
 import android.app.Application
 import android.content.Context
 import android.system.Os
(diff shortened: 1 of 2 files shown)


platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/LoggerImpl.kt
```diff
@@ -606,6 +609,18 @@ internal class LoggerImpl(
...
     }
 
+    private fun startDebugOperationsAsNeeded(context: Context) {
+        if (!BuildTypeChecker.isDebuggable(context)) {
+            return
+        }
+
+        createTemporaryDeviceCode { result ->
+            if (result is CaptureResult.Success) {
+                Log.i("capture", "Temporary device code: ${result.value}")
```
FranAguilera, line 619: nit: Can we reuse `Capture.LOG_TAG`?

PR #670, opened 2025-10-03
[Android] Bump version for the gradle integration app

FranAguilera:
As a followup we should probably auto update this on release as well

### DIFF
--- a/examples/android/HelloWorldApp.kt
+++ b/examples/android/HelloWorldApp.kt
@@ -32,20 +32,20 @@ class HelloWorldApp : Application() {
 
         val userID = UUID.randomUUID().toString();
 
-        // Logger.start(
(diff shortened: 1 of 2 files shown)


examples/android/HelloWorldApp.kt
```diff
@@ -32,20 +32,20 @@ class HelloWorldApp : Application() {
...
-            fieldProviders = listOf(
-                FieldProvider {
-                    mapOf(
-                        "foo" to "bar",
-                        "user_id" to userID
-                    )
-                }
-            )
-        )
+        // Logger.start(
```
FranAguilera, line 35: Will revert. Just double checking ci should broke with this

murki (approved):

PR #704, opened 2025-10-17
Add xcframework Makefile target

kstenerud:
Convenience Makefile target for building `Capture.xcframework`


kstenerud:
Note: the docs for repin are here: https://bazelbuild.github.io/rules_rust/crate_universe_bzlmod.html

The instructions say to `CARGO_BAZEL_REPIN=1 bazel sync --only=crates`, but this doesn't actually work. Calling `test` does, so I'm calling test on a target that has no tests (so that no tests run, thus keeping the repin fast)

PR #708, opened 2025-10-21
Use mach_continuous_time() for calculating uptime

kstenerud:
`mach_continuous_time()` is as efficient as is possible on Apple systems.

### DIFF
--- a/platform/swift/source/shared/Uptime.swift
+++ b/platform/swift/source/shared/Uptime.swift
@@ -7,39 +7,22 @@
 
 import Foundation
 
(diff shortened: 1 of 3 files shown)


kstenerud:
Note: This introduces three instances of `Thread.sleep(forTimeInterval: 1.0)`.

Timing tests aren't ideal because CI could lag out and cause a spurious failure, so I've allowed a variance of 200ms when checking the results. Because of this, the span tests are now also testing that `Uptime` is actually giving correct elapsed time (within the allowed variance).

This also means that `Uptime` no longer contains hidden test hack code.

PR #713, opened 2025-10-24
[Android] Make installAppExit run on caller code

FranAguilera:
Resolves BIT-6647


platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/events/lifecycle/AppExitLogger.kt
```diff
@@ -69,11 +66,9 @@ internal class AppExitLogger(
...
             return
         }
-        backgroundThreadHandler.runAsync {
-            captureUncaughtExceptionHandler.install(this)
-            saveCurrentSessionId()
-            logPreviousExitReasonIfAny()
-        }
+        captureUncaughtExceptionHandler.install(this)
+        saveCurrentSessionId()
+        logPreviousExitReasonIfAny()
```
snowp: I haven't dug into the code but is `logPreviousExitReasonIfAny` the only bit of this that needs to run inline? Not sure if there is any benefit to running some of the other stuff on the bg thread

snowp (approved):

PR #715, opened 2025-10-24
[Android] Update session persistence

FranAguilera:
## What

Resolves BIT-6726

Instead of relying on `setProcessStateSummary` from app exit, persist the current session using the existing `IPreferences` and decouple it from AppExitLogger

### WARNING: Pending Verification

Verify with fatal issue sessions crashes (both [ActivityBased](https://timeline.bitdrift.dev/session/49f6dac8-d3c7-4876-9812-012f3cbd5277?utm_source=sdk&utilization=0&expanded=-2541339556489310041)/[Fixed](https://timeline.bitdrift.dev/session/ed6de2e9-c781-4846-af87-36bedebffd40?utm_source=sdk)) and with regular app kill with Activity Based session.

### DIFF
--- a/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/events/lifecycle/AppExitLogger.kt
+++ b/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/events/lifecycle/AppExitLogger.kt
@@ -20,10 +20,10 @@ import io.bitdrift.capture.LogLevel
 import io.bitdrift.capture.LogType
 import io.bitdrift.capture.LoggerImpl
 import io.bitdrift.capture.common.ErrorHandler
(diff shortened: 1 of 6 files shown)


platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/events/lifecycle/AppExitLogger.kt
```diff
@@ -71,31 +69,12 @@ internal class AppExitLogger(
         }
         captureUncaughtExceptionHandler.install(this)
         logPreviousExitReasonIfAny()
-        backgroundThreadHandler.runAsync {
```
FranAguilera: This now happens at LoggerImpl constructor

FranAguilera:
Will explore if we can handle all this into the rust layer

PR #721, opened 2025-10-29
Remove internal okhttp reference

EricKuck:
`okhttp3.internal.toHexString` is marked internal as of OkHttp 5.0. It's nothing but a proxy for `java.lang.Long.toHexString()` anyway, so no reason to not just use that directly.


FranAguilera (approved):
thanks!

PR #730, opened 2025-11-04
[Android] [Gradle-Test-App] Adding sleep mode card

FranAguilera:
### What

- Adding sleep mode card to ease testing
- Minor refactor to move testing logic out of FirstFragment

### Verification

Check bd logs while transitioning state

Screenshot 2025-11-04 at 12 27 54

Screenshot 2025-11-04 at 12 28 46

### DIFF
--- a/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/Capture.kt
+++ b/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/Capture.kt
@@ -104,7 +104,7 @@ object Capture {
         private val mainThreadHandler by lazy { MainThreadHandler() }
 
         /**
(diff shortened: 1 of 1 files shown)


platform/jvm/gradle-test-app/src/main/java/io/bitdrift/gradletestapp/ui/compose/MainScreen.kt
```diff
@@ -132,93 +122,96 @@ fun MainScreen(
 
             val listState = rememberLazyListState()
 
-            Box(modifier = Modifier.weight(1f)) {
```
FranAguilera: We don't need this Box

kattrali (approved):
neat!

PR #746, opened 2025-11-17
[Android] expose new js error processing method

FranAguilera:
## What

Related to BIT-6844

Provides new js persist method that will be use for react native integration. Full implementation on follow up PR

## Verification

Check current behavior is intact. 

- [JVM crash](https://explorations.bitdrift.dev/issues/801789271362821980/a6ad7c57-0de4-4044-921a-f2ca3c0b3db8?utm_source=sdk)
- [ANR](https://explorations.bitdrift.dev/issues/12479416272078624542/cf5d06d0-46b2-49e3-a727-da9bce0dde4a?utm_source=sdk)
- [Native crash ](https://explorations.bitdrift.dev/issues/327532788420334552/e8886d75-8db0-4983-8ba4-e2c92ae84c01?utm_source=sdk)

### DIFF
--- a/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/Capture.kt
+++ b/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/Capture.kt
@@ -537,7 +537,7 @@ object Capture {
         }
 
         /**
-         * Internal API that allows that persist JavaScript error report.
+         * Internal API which allows to persist JavaScript error reports.
          *
          * @param errorName The main readable error name
          * @param message The detailed JavaScript error message


platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/persistence/ReporterIssueStorage.kt
```diff
@@ -0,0 +1,84 @@
...
+// LICENSE file or at:
+// https://polyformproject.org/wp-content/uploads/2020/06/PolyForm-Shield-1.0.0.txt
+
+package io.bitdrift.capture.reports.persistence
+
+import io.bitdrift.capture.reports.binformat.v1.issue_reporting.ReportType
+import java.io.File
+import java.util.UUID
+
+internal class ReporterIssueStorage(
```
FranAguilera: Before this was named `FatalIssueReporterStorage.kt`

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/persistence/IReporterIssueStorage.kt
```diff
@@ -10,12 +10,18 @@ package io.bitdrift.capture.reports.persistence
 /**
  * Persists a fatal issue report to disk
  */
-internal interface IFatalIssueReporterStorage {
+internal interface IReporterIssueStorage {
```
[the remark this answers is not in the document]
FranAguilera: yeah but it looks so confusing to read with the `I` prefix so I ended leaving it like this. I can update it if you prefer though

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/processor/IssueReporterProcessor.kt
```diff
@@ -20,19 +22,67 @@ import io.bitdrift.capture.reports.binformat.v1.issue_reporting.Platform
...
 
 /**
  * Process reports into a packed format
  */
-internal class FatalIssueReporterProcessor(
-    private val fatalIssueReporterStorage: IFatalIssueReporterStorage,
-    private val clientAttributes: IClientAttributes,
-    private val streamingReportsProcessor: IStreamingReportProcessor,
-) {
+internal object IssueReporterProcessor {
```
murki: why is this an object (static)? we should really try to stray away from global objects: it makes it hard to test/mock, they can never be garbage collected, require manual lifecycle tracking (isInitialized), etc

murki: yeah let's discuss the alternatives

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/persistence/IReporterIssueStorage.kt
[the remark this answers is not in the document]
murki: as imperfect as the I nomenclature is, if we're sticking with it we might as well use it consistently

PR #747, opened 2025-11-17
[Android] Add JS error file generation report

FranAguilera:
## What

Related to BIT-6844

Depends on https://github.com/bitdriftlabs/capture-sdk/pull/746 and https://github.com/bitdriftlabs/shared-core/pull/318,

Uses shared-core report generation logic for js errors

## Verification

_**NOTE: To setup capture-es to point to local capture-sdk [use this guide here](https://docs.google.com/document/d/1UMIFfuVVH2juz9mzu0tt_--uZFVeERvWAHSCxR017TQ/edit?tab=t.yvrwsslkapmt)**_

- Android expo RN sample app. [See report here](https://explorations.bitdrift.dev/issues/8840162654540574342/19a7cb1b-dc27-4978-90b3-81bb4503f90c?utm_source=sdk)

- iOS expo RN sample app. [See report here](https://explorations.bitdrift.dev/issues/3017071801471540792/64fad9ea-47ae-42a6-9688-b4671cbdc9c0)

### DIFF
--- a/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/processor/IssueReporterProcessor.kt
+++ b/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/processor/IssueReporterProcessor.kt
@@ -13,6 +13,7 @@ import io.bitdrift.capture.BuildConstants
 import io.bitdrift.capture.Capture.LOG_TAG
 import io.bitdrift.capture.attributes.ClientAttributes
 import io.bitdrift.capture.attributes.IClientAttributes
+import io.bitdrift.capture.providers.DateProvider
 import io.bitdrift.capture.reports.binformat.v1.issue_reporting.AppBuildNumber
 import io.bitdrift.capture.reports.binformat.v1.issue_reporting.Architecture
 import io.bitdrift.capture.reports.binformat.v1.issue_reporting.DeviceMetrics
@@ -34,6 +35,7 @@ internal class IssueReporterProcessor(
     private val reporterIssueStorage: IIssueReporterStorage,
     private val clientAttributes: IClientAttributes,
     private val streamingReportsProcessor: IStreamingReportProcessor,
+    private val dateProvider: DateProvider,
 ) {
     companion object {
         // Initial size for file builder buffer
@@ -60,7 +62,7 @@ internal class IssueReporterProcessor(
         sdkVersion: String,
     ) {
         runCatching {
-            val timestamp = System.currentTimeMillis()
+            val timestamp = dateProvider.invoke().time
             val destinationPath =
                 if (isFatalIssue) {
                     reporterIssueStorage.generateFatalIssueFilePath()
--- a/platform/jvm/capture/src/test/kotlin/io/bitdrift/capture/fakes/FakeDateProvider.kt
+++ b/platform/jvm/capture/src/test/kotlin/io/bitdrift/capture/fakes/FakeDateProvider.kt
@@ -1,0 +1,21 @@
+// capture-sdk - bitdrift's client SDK
+// Copyright Bitdrift, Inc. All rights reserved.
+//
+// Use of this source code is governed by a source available license that can be found in the
+// LICENSE file or at:
+// https://polyformproject.org/wp-content/uploads/2020/06/PolyForm-Shield-1.0.0.txt
+
+package io.bitdrift.capture.fakes
+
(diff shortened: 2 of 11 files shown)


platform/swift/source/src/bridge.rs
```diff
@@ -567,6 +581,126 @@ extern "C" fn capture_process_issue_reports(
...
+  stack_trace: *const c_char,
+  is_fatal: bool,
+  engine: *const c_char,
+  debugger_id: *const c_char,
+  timestamp_millis: i64,
+  destination_path: *const c_char,
+) {
+  with_handle_unexpected(
+    || -> anyhow::Result<()> {
+      let error_name_str = unsafe { CStr::from_ptr(error_name) }.to_str()?;
```
snowp: in Rust you typically wouldn't put the typename at the end like this, shadowing is much more idiomatic

```
let error_name = unsafe { CStr::from_ptr(err_name)}.to_str()?;
```

platform/swift/source/src/bridge.rs
```diff
@@ -567,6 +581,126 @@ extern "C" fn capture_process_issue_reports(
...
+      let error_message_str = unsafe { CStr::from_ptr(error_message) }.to_str()?;
+      let stack_trace_str = unsafe { CStr::from_ptr(stack_trace) }.to_str()?;
+      let engine_str = unsafe { CStr::from_ptr(engine) }.to_str()?;
+      let destination_str = unsafe { CStr::from_ptr(destination_path) }.to_str()?;
+
+      let debugger_id_opt = if debugger_id.is_null() {
```
snowp: this is called debug_id everywhere else

platform/swift/source/src/bridge.rs
```diff
@@ -567,6 +581,126 @@ extern "C" fn capture_process_issue_reports(
...
+        } else {
+          Some(id_str)
+        }
+      };
+
+      let timestamp_seconds = u64::try_from(timestamp_millis / 1_000).unwrap_or_default();
+      let timestamp_nanos = u32::try_from((timestamp_millis % 1_000) * 1_000).unwrap_or_default();
+
+      let metadata_provider = logger_id.metadata_provider();
+      let (ootb_fields, _custom_fields) = metadata_provider.fields()?;
```
snowp: this call is a bit scary since it ends up calling back into all the user provided FieldProviders in addition the default ones. Using global_state::Reader might be safer until we can redo the interfaces to not have to rely on field providers for all this

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/processor/IssueReporterProcessor.kt
```diff
@@ -47,22 +47,42 @@ internal class IssueReporterProcessor(
...
+        sdkVersion: String,
     ) {
-        // TODO(Fran): To be implemented in follow up PRs
-        Log.d(
-            LOG_TAG,
-            "Persist JS error with name:$errorName, message:$message, isFatalIssue:$isFatalIssue," +
-                " engine:$engine, debuggerId:$debuggerId, stack:$stack",
-        )
+        runCatching {
+            val timestamp = System.currentTimeMillis()
```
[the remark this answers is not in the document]
FranAguilera: nice, will update before merging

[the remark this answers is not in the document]
FranAguilera: Okays, discussed internally. I understands now the reason for dateProvider on start. Updated to use that one or the default provided at LoggerImpl

platform/jvm/capture/src/test/kotlin/io/bitdrift/capture/fakes/FakeDateProvider.kt
```diff
@@ -0,0 +1,21 @@
...
+// https://polyformproject.org/wp-content/uploads/2020/06/PolyForm-Shield-1.0.0.txt
+
+package io.bitdrift.capture.fakes
+
+import io.bitdrift.capture.providers.DateProvider
+import java.util.Date
+
+/**
+ * Fake [DateProvider] for use in tests.
+ * Returns a fixed timestamp by default, but can be configured to return different values.
```
murki: /nit: but wouldn't the best way to configure this stub is to receive the value you want it to return in the ctor?

PR #760, opened 2025-11-21
android: stop tracking session_id for app exit in favor of new API

snowp:
With shared-core no longer requiring the session_id to emit a log with the session id for the previous processs run we no longer need to track this.

Updates the JNI interface to pass a bool instead of the previous string

Test session:
https://timeline.bitdrift.dev/session/eac79e6f-920f-4493-bf99-efed1df32ede?utm_source=sdk&utilization=0&expanded=987426626474647588 using fixed session strategy showing that AppExit is still being emitted for the correct session

Depends on https://github.com/bitdriftlabs/shared-core/pull/327

---

- [x] Point shared-core to main once upstream PR merged

### DIFF
--- a/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/LoggerImpl.kt
+++ b/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/LoggerImpl.kt
@@ -455,10 +455,6 @@ internal class LoggerImpl(
         }
     }
 
-    internal fun shutdown() {
-        CaptureJniLibrary.shutdown(this.loggerId)
-    }
-
     internal fun logSessionReplayScreen(
         fields: Map<String, FieldValue>,
         duration: Duration,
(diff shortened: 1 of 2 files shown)


platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/LoggerImpl.kt
```diff
@@ -191,7 +190,7 @@ internal class LoggerImpl(
             bridge.createLogger(
                 sdkDirectory,
                 apiKey,
-                sessionStrategy.createSessionStrategyConfiguration { appExitSaveCurrentSessionId(it) },
+                sessionStrategy.createSessionStrategyConfiguration {},
```
FranAguilera: we could now remove the onSessionIdChanged callback from the method definition

FranAguilera (commented):
overall lgtm, will validate in a bit

FranAguilera (approved):
lgtm just a few nits on naming

Also verified with ActivityBased session https://timeline.bitdrift.dev/session/0d7cc3c4-fc40-4773-b66c-817c84a2ceff?utm_source=sdk&utilization=0

PR #772, opened 2025-11-27
[Android] Fix isMemoryLow reporting + report jvm used percent

FranAguilera:
### What

Part of BIT-6555

[See Related notes here](https://docs.google.com/document/d/15ciNtCaMLDMGeD5hylwZKklDD_V-crmjUn4HYSitXkA/edit?tab=t.0)

Improves accuracy of `_is_memory_low` detection by comparing the JVM memory percent usage against a configurable percent threshold (default is 90%).

####  Memory calculation:

- usedMemory = totalMemory - freeMemory 
- percentage = usedMemory / maxMemory * 100

#### Why `usedMemory = totalMemory - freeMemory` as the numerator instead of just `totalMemory`?

TLDR: While totalMemory represents heap claimed from the OS (system perspective), for detecting app-level memory pressure we are relying on `totalMemory - freeMemory` to measure actual heap usage.

| API |  Context |
| ----| --------|
| [totalMemory](https://developer.android.com/reference/java/lang/Runtime#totalMemory()) | Current heap size (grows dynamically up to maxMemory) |
| [freeMemory](https://developer.android.com/reference/java/lang/Runtime#freeMemory())| Available space within that allocated heap
| `usedJvmMemory` = `totalMemory` - `freeMemory` | Heap space in use |

- For memory pressure detection we are using `totalMemory - freeMemory` because it reflects actual heap usage. After GC runs, `freeMemory` can increase but `totalMemory` stays the same, so using `totalMemory` alone wouldn't reflect reduced pressure.

####  Why `maxMemory` as denominator:

- `maxMemory` is constant for the process lifetime 
- `totalMemory` grows dynamically as the heap expands, which would give us inconsistent percentages if used as denominator

####  Changes:

- Added `_jvm_used_percent` field to memory attributes
- `_is_memory_low` now compares `jvmUsedPercent >= configuredThreshold`
Threshold validation: must be between 50-100% (values outside range are rejected)
- `_is_memory_low` field is only reported when the threshold is configured

### Verification

[See timeline session here for reference](https://timeline.bitdrift.dev/session/3c6659fe-f7ab-42b9-9121-6a29875e5b09?utm_source=sdk&utilization=1&expanded=9007329720549305601)

Modified the `gradle-test-app` to easily trigger this memory pressure situation and validate the sessions report `_is_memory_low` and `_jvm_used_percent` as expected

Screenshot 2025-12-03 at 16 08 54

Screenshot 2025-12-03 at 16 09 40

[See session with ANR after memory pressure increase here ](https://timeline.bitdrift.dev/session/e8a7943b-dee2-4625-be2b-ab93a73728d8?utm_source=sdk&utilization=1&expanded=-6663787307238445174)

[ANR report](https://explorations.bitdrift.dev/issues/5177789378987482565/b17ad681-1c17-4c46-85af-633baea62924?utm_source=sdk)

GC events below for an artificial memory pressure event

Screenshot 2025-12-03 at 16 13 04

## Configuration flag name

Name: `app_low_memory_percent_threshold`
Default value: 90

## Follow up

As part of the same ticket BIT-6555 [will do required frontend changes here ](https://github.com/bitdriftlabs/valve/pull/853)(e.g. update calculation percentage to use max instead of total) and explore equivalent of this work for iOS (see https://github.com/bitdriftlabs/capture-sdk/pull/780)

### DIFF
--- a/platform/jvm/capture/src/test/kotlin/io/bitdrift/capture/events/performance/MemoryMetricsProviderTest.kt
+++ b/platform/jvm/capture/src/test/kotlin/io/bitdrift/capture/events/performance/MemoryMetricsProviderTest.kt
@@ -37,7 +37,7 @@ class MemoryMetricsProviderTest {
             MemoryMetricsProvider(activityManager, jvmMemoryProvider = jvmMemoryProvider)
         memoryMetricsProvider.runtime = runtime
         whenever(runtime.getConfigValue(RuntimeConfig.APP_LOW_MEMORY_PERCENT_THRESHOLD)).thenReturn(
-            95,
+            90,
         )
     }
 
@@ -105,7 +105,7 @@ class MemoryMetricsProviderTest {
 
     @Test
     fun isMemoryLow_whenUsageAtThreshold_shouldReturnTrue() {
-        whenever(jvmMemoryProvider.usedMemoryBytes()).thenReturn(95_000L)
+        whenever(jvmMemoryProvider.usedMemoryBytes()).thenReturn(90_000L)
         whenever(jvmMemoryProvider.maxMemoryBytes()).thenReturn(100_000L)
 
         val result = memoryMetricsProvider.isMemoryLow()
@@ -115,7 +115,7 @@ class MemoryMetricsProviderTest {
 
     @Test
     fun isMemoryLow_whenUsageBelowThreshold_shouldReturnsFalse() {
-        whenever(jvmMemoryProvider.usedMemoryBytes()).thenReturn(50_000L)
+        whenever(jvmMemoryProvider.usedMemoryBytes()).thenReturn(89_000L)
         whenever(jvmMemoryProvider.maxMemoryBytes()).thenReturn(100_000L)
         whenever(runtime.getConfigValue(RuntimeConfig.APP_LOW_MEMORY_PERCENT_THRESHOLD)).thenReturn(
             90,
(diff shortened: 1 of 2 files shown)


platform/jvm/gradle-test-app/src/main/java/io/bitdrift/gradletestapp/ui/compose/StressTestScreen.kt
```diff
@@ -0,0 +1,276 @@
...
+                    .padding(paddingValues)
+                    .padding(horizontal = 16.dp, vertical = 8.dp),
+            verticalArrangement = Arrangement.spacedBy(16.dp),
+            contentPadding = PaddingValues(bottom = 16.dp),
+        ) {
+            item {
+                MemoryPressureCard(onAction = onAction)
+            }
+            item {
+                JankyFramesCard(onAction = onAction)
```
FranAguilera: As I was here I added jank frame and strict mode triggers so is easier to test. As a follow up will add a leak generation one

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/events/performance/MemoryMetricsProvider.kt
```diff
@@ -9,61 +9,74 @@ package io.bitdrift.capture.events.performance
...
+            put("_is_memory_low", if (isMemoryLow()) "1" else "0")
+            put("_jvm_used_percent", "%.3f".format(jvmUsedPercent()))
         }
 
     override fun getMemoryClass(): Map<String, String> = buildMap { put("_memory_class", memoryClassMB().toString()) }
 
     override fun isMemoryLow(): Boolean {
-        val totalUsedMemoryByApp = usedJvmMemoryBytes() + allocatedNativeHeapSizeBytes()
-        return totalUsedMemoryByApp > memoryThresholdBytes
+        val thresholdPercent = appLowMemoryConfigThreshold ?: return false
```
murki: we should probably not report the field al together if it's disabled (that way those data points won't be considered at all in the charts)

PR #779, opened 2025-12-04
[iOS] [Fatal Issue] Report disabled state even for non-real devices

FranAguilera:
## What

While testing this https://github.com/bitdriftlabs/capture-es/pull/177. I've noticed we weren't reporting CLIENT_CONFIG_DISABLED when user opt-out to fatal issue reporting and is using a simulator.

## Verification

Verified on this session https://timeline.bitdrift.dev/session/644B08C9-48D8-44D3-894E-509014FA1BF2?utm_source=sdk&utilization=0&expanded=-6261768846232997110

### DIFF
--- a/platform/swift/source/Logger.swift
+++ b/platform/swift/source/Logger.swift
@@ -232,12 +232,12 @@ public final class Logger {
 
         self.deviceCodeController = DeviceCodeController(client: client)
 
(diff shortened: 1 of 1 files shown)


kattrali (commented):
I'm not sure this is a bug, since the config has no effect whether set or not 🤔

PR #787, opened 2025-12-09
[Android] Avoid lateinit for setting up issueReporterProcessor

FranAguilera:


murki (approved):
thanks I experienced this crash yesterday as well

PR #788, opened 2025-12-09
android: remove ff classes and handle historical app exits missing

snowp:
---


murki (approved):
I'm supportive of this cleanup since it looks like these cases have never been actionable. Let's wait for @FranAguilera to take a pass as well

PR #805, opened 2025-12-18
[Android] Reduce JVM allocations

FranAguilera:
## What 

Related to BIT-5950, BIT-5957

Reduces overall JVM allocations for Android by avoiding internal hashmap creation and passing instead directly arrays to JNI layer.

Main changes:

- Removed `internal typealias InternalFieldsList = List<Field>` and `internal typealias InternalFieldsMap = Map<String, FieldValue>`
- Introduced `ArrayFields`, which internally holds parallel arrays to avoid creating wrapper objects for each field
- Using now`Array<Field>` (before `Map<String,Field>`) only at JNI boundaries where native code expects it (via toLegacyJniFields())
- Updated/Added helper functions: `fieldOf()`, `fieldsOf()`, `combineFields()` for cleaner fields construction
- Update CaptureTree (timber plugin) to use optimized log version

```
e.g.

// Before, a common pattern to construct internal fields
val fields: Map<String, String> = mapOf(
      "_key_1" to "_value_1",
      "_key_2" to "_value_2"
  ).toFields()

// now
val fields = fieldsOf(
   "_key_1" to "_value_1"
   "_key_2" to "_value_2"
   )

// For single field creation you can use to avoid Pair allocation
val field = fieldOf( "_key_1" , "_value_1")
```

## Follow ups

- To provide new public apis that take directly `ArrayFields` instead of `Map<String,String>` for defining fields
- To swap remaining cases of Array<Fields>
- Run benchmarks on ci
- Further improvements with JNI crossing boundaries

### Run `LogBenchmarkTest` on device Samsung SM-S711B

Below results against latest main

| Test name                                            | allocs on Main | allocs on PR | % allocs Reduction |
|------------------------------------------------------|----------------|--------------|-------------|
| trackSpansWithoutFields                              | 117            | 54           | 53.85%      |
| trackSpansWithFields                                 | 3754           | 266          | 92.91%      |
| logNotMatchedNoFields                                | 2              | 0            | 100.00%     |
| logNotMatched5Fields                                 | 34             | 14           | 58.82%      |
| logNotMatched10Fields                                | 48             | 17           | 64.58%      |
| logNotMatched500Fields                               | 1148           | 34           | 97.04%      |
| logNotMatched5000Fields                              | 10034          | 77           | 99.23%      |
| logHttpNetworkLog50FieldsAndHeadersAndFieldProviders | 2223           | 1010         | 54.54%      |

| Test name                                            | duration (ns) on Main | duration (ns) on PR | % duration Reduction |
|------------------------------------------------------|------------------|---------------|-------------|
| trackSpansWithoutFields                              | 39,358 ns        | 9,924 ns      | 74.77%      |
| trackSpansWithFields                                 | 1,173,969 ns     | 661,361 ns    | 43.62%      |
| logNotMatchedNoFields                                | 12,958 ns        | 621 ns        | 95.21%      |
| logNotMatched5Fields                                 | 18,758 ns        | 4,148 ns      | 77.87%      |
| logNotMatched10Fields                                | 23,928 ns        | 7,566 ns      | 68.37%      |
| logNotMatched500Fields                               | 514,746 ns       | 319,711 ns    | 37.87%      |
| logNotMatched5000Fields                              | 5,243,689 ns     | 3,312,847 ns  | 36.79%      |
| logHttpNetworkLog50FieldsAndHeadersAndFieldProviders | 463,147 ns       | 285,884 ns    | 38.25%      |

| Main | PR |
| -----| ----- |
| Screenshot 2025-12-19 at 22 18 59 | Screenshot 2025-12-19 at 22 20 11

### Sanity test on main components 

session replay rendering, spans, timeline logs, workflow matching

- [gradle-test-app session](https://timeline.bitdrift.dev/session/f1b06876-7c7a-4bdf-891c-492e614de67e?utm_source=sdk&utilization=1&expanded=7121421760872114140)  

### Key-Value insertion of `fieldsOf` and `combinedFields`

When using `Fields`, verified that the last updated key value is respected. See https://github.com/bitdriftlabs/capture-sdk/pull/805#discussion_r2635332089

### Additional stress test

- Calling bitdrift public apis on a hot path and from different threads

---

### DIFF
--- a/platform/jvm/core/src/ffi.rs
+++ b/platform/jvm/core/src/ffi.rs
@@ -9,7 +9,7 @@ use crate::jni::{initialize_class, initialize_method_handle, CachedClass, Cached
 use anyhow::bail;
 use bd_client_common::error::InvariantError;
 use bd_logger::{AnnotatedLogField, AnnotatedLogFields, LogFieldKind, LogFieldValue, LogFields};
-use jni::objects::{JMap, JObject, JPrimitiveArray};
+use jni::objects::{JMap, JObject, JObjectArray, JPrimitiveArray};
 use jni::signature::{Primitive, ReturnType};
 use jni::JNIEnv;
 use std::collections::HashMap;
@@ -139,19 +139,16 @@ fn extract_field(
 /// More efficient than List because arrays allow direct indexed access without iterator overhead.
 pub fn jarray_to_annotated_fields(
   env: &mut JNIEnv<'_>,
-  fields_array: &JObject<'_>,
+  fields_array: &JObjectArray<'_>,
   kind: LogFieldKind,
 ) -> anyhow::Result<AnnotatedLogFields> {
-  use jni::objects::JObjectArray;
-
-  let array = unsafe { JObjectArray::from_raw(fields_array.as_raw()) };
-  let len = env.get_array_length(&array)?;
+  let len = env.get_array_length(fields_array)?;
   #[allow(clippy::cast_sign_loss)]
   let mut fields = AnnotatedLogFields::with_capacity(len as usize);
 
   for i in 0 .. len {
     env.with_local_frame(16, |env| -> anyhow::Result<()> {
-      let field_obj = env.get_object_array_element(&array, i)?;
+      let field_obj = env.get_object_array_element(fields_array, i)?;
       let (key, value) = extract_field(env, &field_obj)?;
       fields.insert(key.into(), AnnotatedLogField { value, kind });
       Ok(())
@@ -165,19 +162,15 @@ pub fn jarray_to_annotated_fields(
 /// Similar to `jarray_to_annotated_fields` but returns `LogFields` without annotations.
 pub(crate) fn jarray_to_fields(
   env: &mut JNIEnv<'_>,
-  fields_array: &JObject<'_>,
+  fields_array: &JObjectArray<'_>,
 ) -> anyhow::Result<LogFields> {
-  use jni::objects::JObjectArray;
-
-  // SAFETY: We know this JObject is actually an array passed from Kotlin
-  let array = unsafe { JObjectArray::from_raw(fields_array.as_raw()) };
-  let len = env.get_array_length(&array)?;
+  let len = env.get_array_length(fields_array)?;
   #[allow(clippy::cast_sign_loss)]
   let mut fields = LogFields::with_capacity(len as usize);
 
   for i in 0 .. len {
     env.with_local_frame(16, |env| -> anyhow::Result<()> {
-      let field_obj = env.get_object_array_element(&array, i)?;
+      let field_obj = env.get_object_array_element(fields_array, i)?;
       let (key, value) = extract_field(env, &field_obj)?;
       fields.insert(key.into(), value);
       Ok(())
@@ -195,23 +188,18 @@ pub(crate) fn jarray_to_fields(
 /// The keys and values arrays must have the same length - keys[i] corresponds to values[i].
 pub fn string_arrays_to_annotated_fields(
   env: &mut JNIEnv<'_>,
-  keys: &JObject<'_>,
-  values: &JObject<'_>,
+  keys: &JObjectArray<'_>,
+  values: &JObjectArray<'_>,
   kind: LogFieldKind,
 ) -> anyhow::Result<AnnotatedLogFields> {
-  use jni::objects::JObjectArray;
-
-  let keys_array = unsafe { JObjectArray::from_raw(keys.as_raw()) };
-  let values_array = unsafe { JObjectArray::from_raw(values.as_raw()) };
-
-  let len = env.get_array_length(&keys_array)?;
+  let len = env.get_array_length(keys)?;
   #[allow(clippy::cast_sign_loss)]
   let mut fields = AnnotatedLogFields::with_capacity(len as usize);
 
   for i in 0 .. len {
     env.with_local_frame(4, |env| -> anyhow::Result<()> {
-      let key_obj = env.get_object_array_element(&keys_array, i)?;
-      let value_obj = env.get_object_array_element(&values_array, i)?;
+      let key_obj = env.get_object_array_element(keys, i)?;
+      let value_obj = env.get_object_array_element(values, i)?;
 
       let key = unsafe { env.get_string_unchecked(&key_obj.into()) }?
         .to_string_lossy()
--- a/platform/jvm/core/src/jni.rs
+++ b/platform/jvm/core/src/jni.rs
@@ -40,6 +40,7 @@ use jni::objects::{
   JClass,
   JMethodID,
   JObject,
+  JObjectArray,
   JPrimitiveArray,
   JString,
   JValueGen,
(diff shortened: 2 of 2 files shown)


platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/ILogger.kt
```diff
@@ -116,6 +117,28 @@ interface ILogger {
...
+     * To construct `Fields` instance you can use FieldProvider.fieldsOf("_key", "_value")
+     *
+     * @param level the severity of the log.
+     * @param fields key-value pairs to be added to the log line, constructed via [io.bitdrift.capture.providers.fieldsOf].
+     * @param throwable an optional throwable to include in the log line.
+     * @param message the main message of the log line, evaluated lazily.
+     * @see log for the current logging method that accepts `Map<String, String>`.
+     * @see io.bitdrift.capture.providers.fieldsOf for constructing [Fields] instances.
+     */
+    fun logOptimized(
```
kattrali: Why is the name change needed? Can there not be a method overload for `log()` which accepts `Fields` instead of a `Map`?

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/providers/FieldProvider.kt
```diff
@@ -115,30 +120,202 @@ internal fun ByteArray.toFieldValue() =
...
+ * This is an optimization to avoid creating wrapper objects when passing to JNI.
+ * keys[i] corresponds to values[i].
+ *
+ * Note to construct it please use, fieldsOf(vararg pairs: Pair<String, String>): Fields
+ *
+ * @property keys Array of field keys
+ * @property values Array of field values corresponding to keys
+ */
+@Suppress("UndocumentedPublicFunction")
+class Fields internal constructor(
```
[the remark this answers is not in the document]
FranAguilera: Good observation, the Fields class intentionally allows duplicate keys. this mirrors the behavior of the current List<Field> (on main named InternalFieldsList) which also didn't deduplicate.

The deduplication is handled on the [native/Rust side here in metadata.rs](https://github.com/bitdriftlabs/shared-core/blob/cf38a7165187d7bd4cf8858e803e4442c5f245d7/bd-logger/src/metadata.rs#L150).

I could add a comment to Fields and combineFields() to document this.

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/ILogger.kt
kattrali: woof that is such a bummer 🤦  and there's no way to resolve this, perhaps by making the `Fields` variant require a value instead of having a default value?

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/providers/FieldProvider.kt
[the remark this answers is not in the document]
FranAguilera: Sorry i mean this acts like a combined map as before, the combined fields is what will be send to jni, then jni will handle this per log call (e.g. logInfo) as mentioned in the comment.

Here a demo with existing HashMap API

Screenshot 2025-12-19 at 16 52 30

Screenshot 2025-12-19 at 16 51 44

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/ILogger.kt
FranAguilera: Yes but that will require the caller to pass Fields.EMPTY each time

FranAguilera: yup i mean i can handle internally setting Fields.EMPTY when we expose new apis with Fields later on

kattrali: ah! on the same page, sounds good

murki: +1 I would also prefer if we can use the same method name

platform/jvm/capture/src/test/kotlin/io/bitdrift/capture/utils/FieldTestUtils.kt
```diff
@@ -0,0 +1,27 @@
...
+
+@file:Suppress("INVISIBLE_MEMBER", "INVISIBLE_REFERENCE")
+
+// capture-sdk - bitdrift's client SDK
+// Copyright Bitdrift, Inc. All rights reserved.
+//
+// Use of this source code is governed by a source available license that can be found in the
+// LICENSE file or at:
+// https://polyformproject.org/wp-content/uploads/2020/06/PolyForm-Shield-1.0.0.txt
+
```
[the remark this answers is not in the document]
FranAguilera: Uops i run the license header script a few times. Will fix soon

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/providers/FieldProvider.kt
```diff
@@ -89,14 +85,14 @@ sealed class FieldValue {
     }
 }
 
-typealias Fields = Map<String, String>
+typealias LegacyFields = Map<String, String>
```
murki: isn't this Type publicly exposed in our SDK? This could mean a breaking change if anyone is using it explicitly in any way. Plus why call them legacy, isn't the plan to keep supporting them moving fwd?

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/providers/FieldProvider.kt
```diff
@@ -115,30 +111,238 @@ internal fun ByteArray.toFieldValue() =
...
+ * This is an optimization to avoid creating wrapper objects when passing to JNI.
+ * keys[i] corresponds to values[i].
+ *
+ * Note to construct it please use, fieldsOf(vararg pairs: Pair<String, String>): Fields
+ *
+ * @property keys Array of field keys
+ * @property values Array of field values corresponding to keys
+ */
+@Suppress("UndocumentedPublicClass", "UndocumentedPublicFunction")
+class Fields internal constructor(
```
murki: if this class is indeed public (and you do provide kdocs above) why mark it as `"UndocumentedPublicClass", "UndocumentedPublicFunction"` ?

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/providers/FieldProvider.kt
FranAguilera: good catch, I though it was mainly used by `FieldProvider` only but totally agree customers could be using this somehow and be a breaking change, will revert back to original

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/CaptureJniLibrary.kt
```diff
@@ -205,7 +207,7 @@ internal object CaptureJniLibrary : IBridge, IStreamingReportProcessor {
      */
     external fun writeSessionReplayScreenLog(
         loggerId: Long,
-        fields: Map<String, FieldValue>,
+        fields: Array<Field>,
```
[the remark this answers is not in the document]
murki: oh I see it below

platform/jvm/core/src/ffi.rs
```diff
@@ -150,52 +187,46 @@ pub(crate) fn jobject_list_to_fields(
...
-/// Internally does a lossy conversion into UTF-8 per [`to_string_lossy`](
-/// https://docs.rs/jni/latest/jni/strings/struct.JNIStr.html#method.to_string_lossy).
-pub fn jobject_map_to_fields(
+/// Converts parallel Java String arrays (keys and values) into `AnnotatedLogFields`.
+/// This is more efficient than using Field objects because it avoids:
+/// 1. Creating Field wrapper objects on the Kotlin side
+/// 2. Multiple JNI method calls per field (getKey, getValueType, getValue)
+///
+/// The keys and values arrays must have the same length - keys[i] corresponds to values[i].
+pub fn string_arrays_to_annotated_fields(
```
snowp: This used to support passing binary field values is that no longer required? Maybe it's handled elsewhere?

I haven't read the rest of the PR but how come we can't consolidate the logic to using one way to convert JVM values to log fields? Seems like we can figure out the best way and just use that everywhere?

snowp (commented):
The JNI changes look fine to me except maybe the binary fields part? Other than that just suggestions that would make the code less complicated

murki (approved):
I think you've addressed most of my jvm-related comments

PR #806, opened 2025-12-19
[Android] Add benchmark tests for Timber CaptureTree

FranAguilera:
### What

Adding more microbenchmark test to keep track of allocations at Timber CaptureTree. 

NOTE: As a follow up will try to run these on ci so we can compare PR changes against main

### Test results

Screenshot 2025-12-19 at 14 14 54


FranAguilera:
Actually timber is only adding constant fields so alloc impact should be minimal

PR #808, opened 2025-12-22
fix(ios): use recorded bundle version as build number

kattrali:
Simplify the recording of app version and bundle version to use the versions as recorded from app metadata, which should align with log metadata


FranAguilera:
going through the list of open PRs. Is this still needed @kattrali ?

PR #809, opened 2025-12-23
feat(webviews): add Android MVP WebView support

jacksonhardaker:
This is the feature/release branch for the MVP of WebView support. Moving forward, additional work will be branched from here.

Example session opening amazon.com: https://timeline.bitdrift.dev/session/c321a55f-c7cd-49cf-a99b-e4e34fd4f03e?utm_source=sdk&utilization=0#timeline

Example session opening wikipedia.org: https://timeline.bitdrift.dev/session/7049a0c4-20b9-44c8-8325-2087967f3d04?utm_source=sdk&utilization=0#timeline

---

### DIFF
--- a/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/webview/WebViewCapture.kt
+++ b/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/webview/WebViewCapture.kt
@@ -120,7 +120,10 @@ internal object WebViewCapture {
     private fun ILogger.logInstrumentationNotInitialized(reason: String) {
         log(
             LogLevel.WARNING,
-            fieldsOf("reason" to reason),
+            fieldsOf(
+                "reason" to reason,
+                "_source" to "webview",
+            ),
         ) {
             "webview.notInitialized"
         }


platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/webview/WebViewCapture.kt
```diff
@@ -0,0 +1,258 @@
...
+            }
+            
+            webview.webViewClient = capture
+
+            // Enable JavaScript (required for bridge)
+            webview.settings.javaScriptEnabled = true
+
+            // Register JavaScript interface for receiving bridge messages
+            val bridgeHandler = WebViewBridgeHandler(effectiveLogger, capture)
+            webview.addJavascriptInterface(bridgeHandler, BRIDGE_NAME)
```
murki: if this works well we could just use this approach for everything and replace the current `webview.pageLoad` span and forego the need for `androidx.webkit.WebViewCompat` altogether

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/webview/WebViewCapture.kt
```diff
@@ -0,0 +1,258 @@
...
+                // Fall back to evaluateJavascript after page load on older WebViews
+                logger?.log(LogLevel.DEBUG, emptyMap()) {
+                    "WebView DOCUMENT_START_SCRIPT not supported, using fallback injection"
+                }
+                return
+            }
+
+            try {
+                val script = WebViewBridgeScript.SCRIPT
+                WebViewCompat.addDocumentStartJavaScript(
```
murki: wouldn't we be calling this _in addition_ to  `webview.addJavascriptInterface()` for webviews that support both?

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/webview/WebViewCapture.kt
```diff
@@ -0,0 +1,199 @@
...
+            // Check if we need fallback injection (older WebViews)
+            val needsFallback = !WebViewFeature.isFeatureSupported(WebViewFeature.DOCUMENT_START_SCRIPT)
+
+            // Wrap existing WebViewClient
+            val capture =
+                if (WebViewFeature.isFeatureSupported(WebViewFeature.GET_WEB_VIEW_CLIENT)) {
+                    val original = WebViewCompat.getWebViewClient(webview)
+                    WebViewCapture(original, effectiveLogger, needsFallback)
+                } else {
+                    WebViewCapture(WebViewClient(), effectiveLogger, needsFallback)
```
murki: not sure how this all works but I worry that there might be cases where `WebViewFeature.isFeatureSupported(WebViewFeature.GET_WEB_VIEW_CLIENT)` returns false but the user _still_ provided their own custom `WebViewClient` and we might be just blindly overriding it

platform/swift/source/integrations/webview/WebViewCapture.swift
```diff
@@ -0,0 +1,230 @@
...
+        webView.configuration.userContentController.addUserScript(script)
+
+        // Wrap navigation delegate for page load tracking
+        let navigationDelegate = WebViewNavigationDelegate(
+            original: webView.navigationDelegate,
+            logger: effectiveLogger,
+            messageHandler: capture.messageHandler
+        )
+        capture.navigationDelegate = navigationDelegate
+        webView.navigationDelegate = navigationDelegate
```
kattrali: its not clear to me that this object would not get garbage collected, particularly under memory pressure, since `webView.navigationDelegate` is a [weak ref](https://developer.apple.com/documentation/webkit/wkwebview/navigationdelegate) -- might have to hold on to these? Same for `capture`.

alternately there could be one delegate object that is always around and it keeps a reference to the various webviews and associated data like page load times.

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/webview/WebViewCapture.kt
FranAguilera: This shouldn't be a problem anymore for initial MVP as we removed the fallback logic, more context on https://github.com/bitdriftlabs/capture-sdk/pull/839

PR #866, opened 2026-01-30
[Android] [WebView] Provide safe mode that doesn't require enabling js 

FranAguilera:
## What

Provide a native mode that doesn't rely on injecting js script. 

## Verification

- Testing with gradle-test-app (Will paste link here later)

---

### DIFF
--- a/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/webview/NativeWebViewClient.kt
+++ b/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/webview/NativeWebViewClient.kt
@@ -1,0 +1,359 @@
+// capture-sdk - bitdrift's client SDK
+// Copyright Bitdrift, Inc. All rights reserved.
+//
+// Use of this source code is governed by a source available license that can be found in the
+// LICENSE file or at:
+// https://polyformproject.org/wp-content/uploads/2020/06/PolyForm-Shield-1.0.0.txt
+
+package io.bitdrift.capture.webview
+
+import android.graphics.Bitmap
+import android.net.http.SslError
+import android.os.Build
+import android.os.Message
+import android.view.KeyEvent
+import android.webkit.ClientCertRequest
+import android.webkit.HttpAuthHandler
+import android.webkit.RenderProcessGoneDetail
+import android.webkit.SafeBrowsingResponse
+import android.webkit.SslErrorHandler
+import android.webkit.WebResourceError
+import android.webkit.WebResourceRequest
+import android.webkit.WebResourceResponse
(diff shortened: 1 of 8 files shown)


jacksonhardaker (approved):

jacksonhardaker (approved):
Should we include `onConsoleLog`?

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/webview/NativeWebViewClient.kt
```diff
@@ -0,0 +1,359 @@
...
+    private val config: WebViewConfiguration.NativeOnly,
+) : WebViewClient() {
+    private var pageStartTime: Long = 0L
+
+    override fun onPageStarted(
+        view: WebView,
+        url: String,
+        favicon: Bitmap?,
+    ) {
+        pageStartTime = System.currentTimeMillis()
```
murki: let's make sure to test all these flows very carefully. back when I evaluated a similar approach [here](https://github.com/bitdriftlabs/capture-sdk/pull/803/changes#diff-b897fa2bbc64e01cd879022dddf4f46f230beacd6eef0a073a6dcb5202a2eb8cR24) there were a lot of idiosyncrasies on the order and guarantees of these callbacks. [Here](https://issuetracker.google.com/issues/210920403) are some statements from a google engineer that took me by surprise: 
>I think part of the issue is that onPageStarted is not actually the very start of the page loading process. There are other steps before onPageStarted, but this is the first point when the previous page (if any) is guaranteed to be unloaded and no longer visible. We can't unload the previous page until we have something to swap it with, and we won't know what to swap with until we get the HTTP response.

and

>I would recommend that you do not make an assumption about the order of onPageStarted vs. onReceivedHttpError. I suggest you write code that can handle either order and still have the correct behavior. For example, if onReceivedHttpError was already called for the current page, don't let onPageStarted logic override the error page.

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/webview/NativeWebViewClient.kt
```diff
@@ -0,0 +1,359 @@
...
+            val fields =
+                fieldsOfOptional(
+                    "_url" to url,
+                    "_action" to "end",
+                    "_source" to "webview",
+                    "_mode" to "native",
+                    "_duration_ms" to durationMs?.toString(),
+                )
+            logger.logInternal(LogType.VIEW, LogLevel.DEBUG, fields) {
+                "webview.pageView"
```
murki: shouldn't these be spans?

PR #882, opened 2026-02-12
android: writeLog perf improvements

snowp:
Adds two performance optimizations that should help the inline overhead of logging on Android:
- Pass null instead of empty arrays when there are no fields/matching fields. This allows us to skip the JNI calls for empty arrays
- Manually manage the JNI local references instead of using a local frame, which seems to perform better due to not having to create lots of local frames during field conversion

---


FranAguilera (approved):

FranAguilera:
@snowp is this still being worked on?

snowp:
Going to fold this into what I'm working on now; I think the null checks are worthwhile but not the change to the refs handling

PR #935, opened 2026-04-09
[Android] Add a github action to trigger native crashes

FranAguilera:
## What

A manual workflow to trigger android native crashes on a loop

## Verification

Will test once merged in main


murki (approved):
>Will test once merged in main

you can always test via the gh cli, e.g. `gh workflow run "Workflow Name" --ref your-branch-name`

PR #944, opened 2026-04-15
[Android] Support PreviousRunInfo below OS 11

FranAguilera:
## What

Resolves BIT-7703

Follow up to https://github.com/bitdriftlabs/capture-sdk/pull/922

Update `Capture.Logger.getPreviousRunInfo()` to return if app was terminated due to JVM crash on prior cold launch below OS 11.

## Verification

Verified with these session < OS 11

- [Session with crash](https://timeline.bitdrift.dev/session/84e04b04-1afe-4d91-b4be-5ece1825a5cf?utilization=0&expanded=-7843700512391339268)

Screenshot 2026-04-15 at 16 04 35

- [Next session after crash](https://timeline.bitdrift.dev/session/75d9a466-d8a0-42b3-939a-21cb1db531aa?utm_source=sdk&utilization=0&expanded=-493430855478208199&stacktrace=open)

Screenshot 2026-04-15 at 16 05 36

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3040,9 +3040,9 @@ dependencies = [
 
 [[package]]
 name = "rustls-webpki"
-version = "0.103.10"
+version = "0.103.12"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "df33b2b81ac578cabaf06b89b0631153a3f416b0a886e8a7a1707fb51abbd1ef"
+checksum = "8279bb85272c9f10811ae6a6c547ff594d6a7f3c6c6b02ee9726d1d0dcfcdd06"
 dependencies = [
  "ring",
  "rustls-pki-types",
--- a/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/exitinfo/PreviousRunInfoResolver.kt
+++ b/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/exitinfo/PreviousRunInfoResolver.kt
@@ -9,23 +9,68 @@ package io.bitdrift.capture.reports.exitinfo
 
 import android.os.Build
 import androidx.annotation.RequiresApi
+import io.bitdrift.capture.IPreferences
+import io.bitdrift.capture.reports.jvmcrash.ICaptureUncaughtExceptionHandler
+import io.bitdrift.capture.reports.jvmcrash.IJvmCrashListener
 
 /**
  * Provides [PreviousRunInfo] for the previous app session.
  *
  * On API >= 30, uses [ApplicationExitInfo] directly.
- * On API < 30, returns `null` (no previous run info available yet). Will be implemented in BIT-7703.
+ * On API < 30, relies on manually persisted previous-run state, including when a JVM crash was recorded.
  */
 internal class PreviousRunInfoResolver(
     private val latestAppExitInfoProvider: ILatestAppExitInfoProvider,
-) : IPreviousRunInfoResolver {
+    preferences: IPreferences,
+    captureUncaughtExceptionHandler: ICaptureUncaughtExceptionHandler,
+) : IPreviousRunInfoResolver,
+    IJvmCrashListener {
+    private val previousRunInfoLegacyStore by lazy { LegacyPreviousRunStateStore(preferences) }
+
+    private val legacyPreviousRunState: LegacyPreviousRunState?
+
+    init {
+        if (isBelowApi30()) {
+            legacyPreviousRunState = previousRunInfoLegacyStore.getPreviousState()
+            previousRunInfoLegacyStore.writeState(LegacyPreviousRunState.Started)
+            captureUncaughtExceptionHandler.install(this)
+        } else {
+            legacyPreviousRunState = null
+        }
+    }
+
     override fun get(): PreviousRunInfo? =
-        if (Build.VERSION.SDK_INT < Build.VERSION_CODES.R) {
-            // TODO (BIT-7703): Enable support below OS 11
-            null
+        if (isBelowApi30()) {
+            getFromLegacyState()
         } else {
             getFromAppExitInfo()
         }
+
+    /**
+     * Will only trigger below OS 11
+     */
+    override fun onJvmCrash(
+        thread: Thread,
+        throwable: Throwable,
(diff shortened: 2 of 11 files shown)


Cargo.lock
```diff
@@ -3040,9 +3040,9 @@ dependencies = [
 
 [[package]]
 name = "rustls-webpki"
-version = "0.103.10"
+version = "0.103.12"
```
FranAguilera, line 3043: For now pinned manually given this is only used for tests

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/exitinfo/PreviousRunInfoResolver.kt
```diff
@@ -9,24 +9,69 @@ package io.bitdrift.capture.reports.exitinfo
...
+import io.bitdrift.capture.IPreferences
+import io.bitdrift.capture.reports.jvmcrash.ICaptureUncaughtExceptionHandler
+import io.bitdrift.capture.reports.jvmcrash.IJvmCrashListener
 
 /**
  * Provides [PreviousRunInfo] for the previous app session.
  *
  * On API >= 30, uses [ApplicationExitInfo] directly.
- * On API < 30, returns `null` (no previous run info available yet). Will be implemented in BIT-7703.
+ * On API < 30, relies on manually persisted previous-run state, including when a JVM crash was recorded.
```
murki, line 20: "including" sounds misleading here, isn't it that _only_ in the case of JVM crashes this will report something other than null?

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/exitinfo/PreviousRunInfoResolver.kt
```diff
@@ -62,6 +107,34 @@ data class PreviousRunInfo(
     val terminationReason: ExitReason? = null,
 )
 
+internal class LegacyPreviousRunStateStore(
```
murki, line 110: /namingishard I don't love calling these "legacy". we could take a page from JankStats and be explicit about what it supports, similar to JankStatsApi26, JankStatsApi31, etc

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/exitinfo/PreviousRunInfoResolver.kt
```diff
@@ -9,24 +9,69 @@ package io.bitdrift.capture.reports.exitinfo
...
+     * Will only trigger below OS 11
+     */
+    override fun onJvmCrash(
+        thread: Thread,
+        throwable: Throwable,
+    ) {
+        previousRunInfoLegacyStore.writeState(LegacyPreviousRunState.JvmCrash)
+    }
+
+    private fun isBelowApi30() = Build.VERSION.SDK_INT < Build.VERSION_CODES.R
```
murki, line 59: consider using an injected `BuildVersionChecker` for this -it has the additional advantage that it can be easily mocked for tests

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/exitinfo/PreviousRunInfoResolver.kt
```diff
@@ -9,24 +9,69 @@ package io.bitdrift.capture.reports.exitinfo
...
 import android.os.Build
 import androidx.annotation.RequiresApi
+import io.bitdrift.capture.IPreferences
+import io.bitdrift.capture.reports.jvmcrash.ICaptureUncaughtExceptionHandler
+import io.bitdrift.capture.reports.jvmcrash.IJvmCrashListener
 
 /**
  * Provides [PreviousRunInfo] for the previous app session.
  *
  * On API >= 30, uses [ApplicationExitInfo] directly.
```
murki, line 19: for bracket linking to work on this it needs to be `[android.app.ApplicationExitInfo]`

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/exitinfo/PreviousRunInfoResolver.kt
```diff
@@ -62,6 +107,34 @@ data class PreviousRunInfo(
...
+        val blocking = state == LegacyPreviousRunState.JvmCrash
+        preferences.setString(STATE_KEY, state.value, blocking = blocking)
+    }
+
+    private companion object {
+        private const val STATE_KEY = "io.bitdrift.capture.previous_run_info.state"
+    }
+}
+
+internal enum class LegacyPreviousRunState(
```
murki, line 125: I usually shy away from adding many classes in the same file, unless said classes are only used within the same file (in which case these should be private). if these are re-used we should consider moving them to separate files

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/exitinfo/PreviousRunInfoResolver.kt
FranAguilera: good idea

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/exitinfo/PreviousRunInfoResolver.kt
FranAguilera: good call, will update

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/exitinfo/PreviousRunInfoResolver.kt
```diff
@@ -62,6 +107,34 @@ data class PreviousRunInfo(
...
+    /** Stable text representation*/
+    val value: String,
```
murki, line 127: If this string representation is only used to store the value in sharedprefs we could use the native methods of `Enum.Value.name` and `Enum.valueOf()` to deal with this

murki (approved):
approach lgtm, left a few nits, nothing major

PR #949, opened 2026-04-17
[Prototype] [Android] Prototype an autoinit

FranAguilera:
Do not review, just for sharing

### DIFF
--- a/platform/jvm/capture/build.gradle.kts
+++ b/platform/jvm/capture/build.gradle.kts
@@ -87,6 +87,7 @@ android {
         checkReleaseBuilds = true
         disable.add("GradleDependency")
         disable.add("AndroidGradlePluginVersion")
+        disable.add("EnsureInitializerMetadata")
     }
 }
 
--- a/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/CaptureInitializer.kt
+++ b/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/CaptureInitializer.kt
@@ -1,0 +1,129 @@
+// capture-sdk - bitdrift's client SDK
+// Copyright Bitdrift, Inc. All rights reserved.
+//
+// Use of this source code is governed by a source available license that can be found in the
+// LICENSE file or at:
+// https://polyformproject.org/wp-content/uploads/2020/06/PolyForm-Shield-1.0.0.txt
+
+package io.bitdrift.capture
+
+import android.content.Context
+import android.content.pm.PackageManager
+import android.util.Log
+import androidx.startup.Initializer
+import okhttp3.HttpUrl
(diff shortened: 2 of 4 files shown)


platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/CaptureInitializer.kt
```diff
@@ -0,0 +1,129 @@
...
+ *   io.bitdrift.capture.API_URL               - Custom API URL (default: https://api.bitdrift.io).
+ *   io.bitdrift.capture.SESSION_STRATEGY      - "fixed" or "activity_based" (default: activity_based).
+ *   io.bitdrift.capture.INACTIVITY_THRESHOLD_IN_MINUTES - Minutes for activity_based strategy (default: 30).
+ *   io.bitdrift.capture.ENABLE_SESSION_REPLAY - "true"/"false" (default: true).
+ *   io.bitdrift.capture.ENABLE_FATAL_ISSUE_REPORTING - "true"/"false" (default: true).
+ *   io.bitdrift.capture.SLEEP_MODE            - "enabled"/"disabled" (default: disabled).
+ */
+class CaptureInitializer : Initializer<Unit> {
+    override fun create(context: Context) {
+        val appContext = context.applicationContext
```
murki, line 33: if we do this then we don't need the `ContextHolder` dep, or you could do `val appContext = ContextHolder.APP_CONTEXT` here instead

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/CaptureInitializer.kt
```diff
@@ -0,0 +1,129 @@
...
+            sessionStrategy = sessionStrategy,
+            configuration = configuration,
+            apiUrl = apiUrl ?: defaultUrl(),
+            context = appContext,
+        )
+    }
+
+    override fun dependencies(): List<Class<out Initializer<*>>> = listOf(ContextHolder::class.java)
+
+    private fun buildSessionStrategy(metadata: android.os.Bundle): io.bitdrift.capture.providers.session.SessionStrategy {
```
murki, line 64: if we go down this route I'd like to see us wrap all the metadata strongly-typed parsing logic inside our Configuration class so the properties stay in sync

platform/jvm/capture/build.gradle.kts
```diff
@@ -87,6 +87,7 @@ android {
         checkReleaseBuilds = true
         disable.add("GradleDependency")
         disable.add("AndroidGradlePluginVersion")
+        disable.add("EnsureInitializerMetadata")
```
murki, line 90: is this required? I'm hoping we don't have to ask customers to do this

PR #952, opened 2026-04-20
iOS: Fix timestamps for crash diagnostics

ArielDemarco:
# Overview
Crash report timestamps were captured using `time(NULL)`, which only provides second-level precision. This made it difficult to accurately correlate crash events with other telemetry.

To solve this we:
- Replace `time(NULL)` with `clock_gettime(CLOCK_REALTIME, ...)` to capture both seconds and nanoseconds at crash time.
- Expose a new `cachedCrashDate` method that reconstructs a full-precision `NSDate` from the cached `KSCrash` report.
- Use this higher-precision timestamp when processing `MXCrashDiagnostic`

[Here's an example](https://timeline.bitdrift.dev/session/9f5c6724-a5dd-40f7-96e1-4b0e78fdb7c3?utm_source=explorations.bitdrift.dev&workflow_id=kPSa&utilization=1)

_Side note_: fixed flaky tests

### DIFF
--- a/platform/swift/source/crash_handler/BitdriftKSCrashHandler.m
+++ b/platform/swift/source/crash_handler/BitdriftKSCrashHandler.m
@@ -69,10 +69,8 @@ static void onCrash(struct KSCrash_MonitorContext *monitorContext) {
     struct timespec crashTime = {0};
     if (clock_gettime(CLOCK_REALTIME, &crashTime) != 0) {
         crashTime.tv_sec = time(NULL);
-        crashTime.tv_nsec = 0;
     }
     g_crashHandlerReportContext.metadata.time = crashTime.tv_sec;
-    g_crashHandlerReportContext.metadata.timeNanos = (uint32_t)crashTime.tv_nsec;
     g_crashHandlerReportContext.monitorContext = monitorContext;
     bitdrift_writeKSCrashReport(&g_crashHandlerReportContext);
 }
@@ -215,9 +213,8 @@ static void onCrash(struct KSCrash_MonitorContext *monitorContext) {
     if (!timestamp.available) {
         return nil;
     }
-    NSTimeInterval interval = (NSTimeInterval)timestamp.seconds + ((NSTimeInterval)timestamp.nanoseconds / NSEC_PER_SEC);
     
-    return [NSDate dateWithTimeIntervalSince1970:interval];
+    return [NSDate dateWithTimeIntervalSince1970:(NSTimeInterval)timestamp.seconds];
 }
 
 + (void)stopCrashReporter {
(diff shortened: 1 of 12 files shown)


platform/swift/source/crash_handler/BitdriftKSCrashHandler.m
```diff
@@ -56,8 +65,14 @@ static void onCrash(struct KSCrash_MonitorContext *monitorContext) {
...
-
-    g_crashHandlerReportContext.metadata.time = time(NULL);
+    
+    struct timespec crashTime = {0};
+    if (clock_gettime(CLOCK_REALTIME, &crashTime) != 0) {
+        crashTime.tv_sec = time(NULL);
+        crashTime.tv_nsec = 0;
+    }
+    g_crashHandlerReportContext.metadata.time = crashTime.tv_sec;
+    g_crashHandlerReportContext.metadata.timeNanos = (uint32_t)crashTime.tv_nsec;
```
Reflejo: Do we need nanos? Seems overkill and maybe a false sense of precision, I would personally remove it. tv_nsec is not really accurate anyway and this is looking at a non monotonically increasing clock

ArielDemarco: Sure! I can remove it! It also makes things simpler

platform/swift/source/crash_handler/BitdriftKSCrashHandler.m
```diff
@@ -39,6 +39,14 @@ void ksmemory_notifyUnhandledFatalSignal(void) {}
     CacheResultSuccess = 3
 } CacheResult;
 
+typedef struct {
+    uint64_t seconds;
+    bool available;
+} CachedCrashTimestamp;
+
+/** Get the crash timestamp from the latest/previous-run KSCrash report */
+CachedCrashTimestamp capture_cached_kscrash_timestamp(void);
```
Reflejo: NIT do we need this struct vs just `uint64_t` and then use `> 0` ?

ArielDemarco: you're right! I simplified to just use the integer and removed the whole `struct`

Reflejo (approved):

FranAguilera (approved):
Nice! :)

FranAguilera:
@ArielDemarco Before merging could you also update CHANGELOG.md, this will ease later release notes generation

Also a minor nit but if you could link a timeline session for verification in the PR description it would be great for anyone who stumbles across this PR later on

ArielDemarco:
> @ArielDemarco Before merging could you also update CHANGELOG.md, this will ease later release notes generation
> 
> Also a minor nit but if you could link a timeline session for verification in the PR description it would be great for anyone who stumbles across this PR later on

Sure! Done!

FranAguilera (approved):

PR #963, opened 2026-04-28
[Crash Reporting] Set killswitch flag default to true

FranAguilera:
### What

Depends on https://github.com/bitdriftlabs/shared-core/pull/474

- Setting default killswitch for crash reporting to true when:
  - File doesn't exist (could be initial app installation or user manually clears app cache)
  - When the flag name in the config file is set to true
- Added additional core field at start log (`_is_sdk_directory_first_created`) to determine when sdk was started without prior bitdrift cached files (this should help determine initial app installation or user clearing app cache manually)

Resolves BIT-8082

### Verification

- Setup sample apps with a default key/url that doesn't depend on cache

| Case | crash_reporting.enabled state |Expectation | Android Session | iOS session | 
| -----  | ------------| ------------| ---------------- | ------------ | 
| Initial app install + crash | `true` | Should detect crashes on initial app installation | [Session](https://timeline.bitdrift.dev/session/72ed0c3c-e648-42b7-b730-c837405c9d3c?utm_source=sdk&utilization=0&expanded=5197364081670406992) | [Session](https://timeline.bitdrift.dev/session/58737a55-9364-4779-8b24-faeecf2ccd3b?utilization=0&expanded=-6845442360564193783) |
| After initial app install + valid bitdrift cache | `false` | Issue reporter not started due to runtime config flag set to false | [Session](https://timeline.bitdrift.dev/s/1ea91623-a912-43e9-a25a-90001b1e8564?utm_source=sdk) | [Session](https://timeline.bitdrift.dev/session/5BB4CFF4-1D1E-445C-85B9-F3D895893471?utm_source=sdk&utilization=0&expanded=648358177421955945) |

---

### DIFF
--- a/platform/swift/source/Logger.swift
+++ b/platform/swift/source/Logger.swift
@@ -156,6 +156,8 @@ public final class Logger {
             return nil
         }
 
+        let isSdkDirectoryFirstCreated = !FileManager.default.fileExists(atPath: directoryURL.path)
+
         let network: URLSessionNetworkClient? = enableNetwork
             ? URLSessionNetworkClient(apiBaseURL: configuration.apiURL)
             : nil
@@ -205,6 +207,7 @@ public final class Logger {
             let fields: Fields = [
                 "_fatal_issue_reporting_state": "\(Logger.issueReporterInitResult.0)",
                 "_fatal_issue_reporting_duration_ms": Logger.issueReporterInitResult.1 * Double(MSEC_PER_SEC),
+                "_is_sdk_directory_first_created": isSdkDirectoryFirstCreated,
                 "_session_replay_enabled": (configuration.sessionReplayConfiguration != nil),
             ]
             self.underlyingLogger.logSDKStart(fields: fields, duration: duration)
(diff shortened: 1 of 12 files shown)


platform/swift/source/Logger.swift
```diff
@@ -245,14 +245,13 @@ public final class Logger {
...
-                }
-                guard let enabled = runtimeConfig[RuntimeVariable.crashReporting.name] as? Bool, enabled else {
-                    return .initialized(.runtimeNotEnabled)
+                if let contents = Logger.cachedReportConfigData(base: directoryURL) {
+                    guard let runtimeConfig = readCachedValues(contents) else {
+                        return .initialized(.runtimeInvalid)
+                    }
+                    guard let enabled = runtimeConfig[RuntimeVariable.crashReporting.name] as? Bool, enabled else {
+                        return .initialized(.runtimeNotEnabled)
+                    }
```
ArielDemarco: This means that if there's any problem while reading the cache, we should default to `false` (and not enable crash reporting)?

FranAguilera: sorry went trough a few iterations and this code changed a bit. 

Yes, the idea is that will set the flag to true only when:
- The file doesn't exist (could be initial app installation or the user cleared the app cache manually)
- The files exists and the flag name has a value of true

ArielDemarco (approved):
LGTM

PR #966, opened 2026-04-29
[Tracing] Exclude tracing for configured paths

FranAguilera:
## What

Resolves BIT-8130

## Runtime configuration

`client_config.trace.excluded_path_prefixes_csv`

## Verification

Android verification with runtime config excluding `/data/2021/pep/population` 

[Session](https://timeline.bitdrift.dev/session/c1b082ec-ea15-45d4-bece-a291a6d3bb6b?utm_source=sdk&utilization=0&expanded=-7166318999005497722)

With excluded path 
Screenshot 2026-04-30 at 12 52 48

With valid path

Screenshot 2026-04-30 at 12 49 07

---

### DIFF
--- a/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/network/okhttp/CaptureOkHttpTracingInterceptor.kt
+++ b/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/network/okhttp/CaptureOkHttpTracingInterceptor.kt
@@ -39,6 +39,18 @@ class CaptureOkHttpTracingInterceptor
 
         private val traceContextFactory by lazy { TraceContextFactory() }
 
+        private val excludedPathPrefixes by lazy {
+            val pathPrefixesToExcludeAsCsv =
+                runtimeProvider.getRuntimeStringConfigValue(
+                    RuntimeStringConfig.TRACING_EXCLUDED_PATH_PREFIXES_CSV,
+                )
+            if (pathPrefixesToExcludeAsCsv.isEmpty()) {
+                emptyList()
+            } else {
+                pathPrefixesToExcludeAsCsv.split(",").map { it.trim() }.filter { it.isNotEmpty() }
+            }
+        }
+
         override fun intercept(chain: Interceptor.Chain): Response {
             val currentLogger = Capture.logger()
             val request = chain.request()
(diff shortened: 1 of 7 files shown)


platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/network/okhttp/CaptureOkHttpTracingInterceptor.kt
```diff
@@ -90,6 +94,9 @@ class CaptureOkHttpTracingInterceptor
 
         private fun isBitdriftInternalRequest(request: Request): Boolean = request.header(BITDRIFT_API_KEY_HEADER) != null
 
+        private fun isOpenTelemetryCollectorRequest(request: Request): Boolean =
+            request.url.encodedPath.startsWith("/opentelemetry.proto.collector.")
```
murki: not a big fan of hardcoding these. is there any more generic way to detect this? do they add a header of their own that we can check instead?

FranAguilera: same not a fan of this, will check what they add once i have a repro full setup

FranAguilera: nvm saw this https://github.com/open-telemetry/opentelemetry-java/blob/1b4bf0c1eeb90cbcd2ae8d99b91b2c503d1e42c3/exporters/otlp/all/src/main/java/io/opentelemetry/exporter/otlp/internal/OtlpUserAgent.java#L17

mattklein123: Yeah agreed not optimal and I think maybe there are other paths that people want to ignore? It's more work but I wonder if we should make this configurable somehow.

FranAguilera: yup was thinking something like this as well. Will update later

FranAguilera: added for now a csv list of path to ignore (added verification steps on the PR description)

PR #969, opened 2026-05-04
Add start/failure callback result

FranAguilera:
## What

Part of BIT-8084

Adds optional `startResult` callback to `Capture.Logger.start`. Customers can use this to determine if sdk was started successfully

Separate "status" sync API is implemented in https://github.com/bitdriftlabs/capture-sdk/pull/984

## Verification

Check logcat/println in sample apps

---

### DIFF
--- a/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/Capture.kt
+++ b/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/Capture.kt
@@ -132,7 +132,11 @@ object Capture {
          *               instructed otherwise during discussions with bitdrift. Defaults to bitdrift's hosted
          *               Compose API base URL.
          * @param context an optional context reference. You should provide the context if called from
-         * a [android.content.ContentProvider]
+         * a [android.content.ContentProvider].
+         * @param startResult an optional callback invoked with the result of the SDK initialization.
+         *                     The callback is always called on the calling thread before [start] returns.
+         *                     On success, it receives a [CaptureResult.Success] containing an [ILogger] instance.
+         *                     On failure, it receives a [CaptureResult.Failure] with a [SdkStartFailure].
          */
         @Synchronized
         @JvmStatic
@@ -145,6 +149,7 @@ object Capture {
             dateProvider: DateProvider? = null,
             apiUrl: HttpUrl = defaultCaptureApiUrl,
             context: Context? = null,
+            startResult: ((CaptureResult<ILogger>) -> Unit)? = null,
         ) {
             start(
                 apiKey,
@@ -155,6 +160,7 @@ object Capture {
                 apiUrl,
                 CaptureJniLibrary,
                 context,
+                startResult,
             )
         }
 
@@ -172,13 +178,13 @@ object Capture {
             apiUrl: HttpUrl = defaultCaptureApiUrl,
             bridge: IBridge,
             context: Context? = null,
+            startResult: ((CaptureResult<ILogger>) -> Unit)? = null,
         ) {
             // There's nothing we can do if we don't have yet access to the application context.
             if (hasInvalidContext(context)) {
-                Log.w(
-                    LOG_TAG,
-                    "Attempted to initialize Capture with a null context",
-                )
+                val errorMessage = "Attempted to initialize Capture with a null context"
+                Log.w(LOG_TAG, errorMessage)
+                startResult?.invoke(CaptureResult.Failure(SdkStartFailure(errorMessage)))
                 return
             }
 
@@ -193,6 +199,7 @@ object Capture {
                     apiUrl = apiUrl,
                     bridge = bridge,
                     context = context,
+                    startResult = startResult,
                 )
             } else {
                 Log.w(LOG_TAG, "Multiple attempts to start Capture")
@@ -658,6 +665,7 @@ object Capture {
         apiUrl: HttpUrl,
         bridge: IBridge,
         context: Context?,
+        startResult: ((CaptureResult<ILogger>) -> Unit)? = null,
     ) {
         try {
             val startSdkTimer = TimeSource.Monotonic.markNow()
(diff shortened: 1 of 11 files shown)


platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/Capture.kt
```diff
@@ -172,13 +177,13 @@ object Capture {
...
         ) {
             // There's nothing we can do if we don't have yet access to the application context.
             if (hasInvalidContext(context)) {
-                Log.w(
-                    LOG_TAG,
-                    "Attempted to initialize Capture with a null context",
-                )
+                val errorMessage = "Attempted to initialize Capture with a null context"
+                Log.w(LOG_TAG, errorMessage)
+                completionResult?.invoke(CaptureResult.Failure(SdkNotStartedError(errorMessage)))
```
murki: do we make any guarantees of which thread the result run? looks like we're not being explicitly consistent

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/Capture.kt
```diff
@@ -133,6 +133,9 @@ object Capture {
          *               Compose API base URL.
          * @param context an optional context reference. You should provide the context if called from
          * a [android.content.ContentProvider]
+         * @param completionResult an optional callback invoked with the result of the SDK start operation.
```
murki: /namingishard: but "completionResult" seems a little vague, maybe "initializationResult" ? "startResult"? or even just "result". also we should mention where can the user expect the callback to run on (or let them provide an Executor)

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/Capture.kt
murki: It also feels weird to repurpose the `SdkNotStartedError` error for this, which before was returned to signify "you called this API but this SDK hasn't been started"

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/Models.kt
```diff
@@ -75,5 +75,9 @@ sealed class ApiError(
 
 /**
  * Represents a failed operation due to the SDK not being started
+ *
+ * @param reason The detailed reason for SDK not started
  */
-data object SdkNotStartedError : Error("SDK not started")
+data class SdkNotStartedError(
```
murki: we should probably create a separate type just for starting errors

murki (approved):
Approach looks good to me, just some minor comments

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/Capture.kt
FranAguilera: Yup will update

Regarding "or let them provide an Executor", I would tackle this separately as will need to add again the inMemory caching for Capture.Logger calls during start initialization. Past PR for ref https://github.com/bitdriftlabs/capture-sdk/pull/482/changes#top

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/Capture.kt
FranAguilera: right now that's the same thread where .start gets called.

I agree regarding different enum types for "not started error". I had this in the old [PR](https://github.com/bitdriftlabs/capture-sdk/pull/482/changes#top) so will look into bringing it back

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/Capture.kt
```diff
@@ -710,9 +717,13 @@ object Capture {
                 sdkConfiguredDuration = sdkConfiguredDuration,
                 captureStartThread = Thread.currentThread().name,
             )
-        } catch (e: Throwable) {
-            Log.w(LOG_TAG, "Failed to start Capture", e)
+
+            completionResult?.invoke(CaptureResult.Success(loggerImpl))
```
murki: wait, this is not different than what they can do today (calling Capture.Logger after the start() call returns). What they want is the success callback to be triggered when the **async** initialization of the rust logger is complete (the buffers are ready, the config loaded, etc) @snowp might know best when exactly in the rust code that is

murki (changes requested):
actually on 2nd look this is not right

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/Capture.kt
snowp: If they're just trying to understand when they can send logs and our init system is able to accept it and process it best effort then all that matters is that the jvm logger handle is non-null, as at this point the pre-config buffer is ready to buffer logs even if the logger itself hasn't initialized the buffers etc

If they are looking for something more around validating that for this startup the logger was initialized correctly and we'll be able to process logs then yeah we'd need to integrate more closely with the Rust side of things, but not sure if that's what's being asked?

murki: yes, it's actually the latter, and it even goes further than that; they would like in the callback to also know if our sdk has been able to connect successfully to our backend cc @mattklein123

snowp: Yeah in that case this is more involved and we need to define the points we want to expose, logger init itself is non-deterministic so we'd need to decide which parts are interesting and expose those. The actual plumbing to invoke the callback is easy enough but we need to decide what we care about since there's not really a single "logger is ready" point in the code 

The most interesting parts of init are maybe 1) the main event loop task has kicked off the various sub-tasks 2) config has loaded (starts draining the pre-config buffer) 3) first API handshake received? Each sub-task has their own init that we don't block on and some stuff (like buffer init) is gated on config load

FranAguilera: Discussed offline, will provide a sync api for the different status

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/Capture.kt
kattrali: Apologies for joining an existing convo, but even `startResult` seems a bit stilted/vague. I thought of a few alternatives which were maybe closer to idiomatic, but perhaps it would be easier to have separate callbacks so it could be like `onFailure(msg)` vs `onLoggerInitialized(logger)`?

----

brief offline chat - maybe it makes sense to have these on `Configuration`, since there's also the potential of other event-based callables, like `onDisconnect`?

FranAguilera:
Re-oppening to re trigger for ci issues

murki (approved):

PR #970, opened 2026-05-04
Bump Xcode to 26.1.1 and fix iOS simulator resolution on macos-26 CI

ArielDemarco:
# Overview
The CI was failing on `macos-26` runners because `mac_ci_setup.sh` and `.bazelrc` were pinned to Xcode 26.0.1, whose [iOS 26.0 simulator runtime is not pre-installed on the runner](https://github.com/actions/runner-images/blob/main/images/macos/macos-26-Readme.md#installed-simulators) (only iOS 26.1, 26.2, and 26.4 runtimes are available).                                                        
                                                                                                                                                        
This PR bumps Xcode to 26.1.1 (which has an available iOS 26.1 simulator runtime) and introduces a `resolve-ios-simulator.sh` script to dynamically resolve the correct simulator device for the active Xcode version, avoiding hardcoded device/version values that break whenever the runner image changes.


FranAguilera (approved):
thanks!

.github/workflows/ios.yaml
```diff
@@ -80,10 +80,12 @@ jobs:
...
       - name: "Install dependencies"
         run: ./ci/mac_ci_setup.sh
-      - run: ./bazelw build --config ci --config release-ios //examples/swift/hello_world:ios_app
+      - name: Resolve iOS simulator
+        run: ./ci/resolve-ios-simulator.sh
+      - run: ./bazelw build --config ci --config release-ios --ios_simulator_device="$IOS_SIMULATOR_DEVICE" --ios_simulator_version="$IOS_SIMULATOR_VERSION" //examples/swift/hello_world:ios_app
         name: "Build app"
       # TODO(snowp): Add some kind of assertion that the app does that it's supposed to
-      - run: ./bazelw run --config ci --config release-ios //examples/swift/hello_world:ios_app &> /tmp/envoy.log &
+      - run: ./bazelw run --config ci --config release-ios --ios_simulator_device="$IOS_SIMULATOR_DEVICE" --ios_simulator_version="$IOS_SIMULATOR_VERSION" //examples/swift/hello_world:ios_app &> /tmp/envoy.log &
```
Reflejo: can we just use `--config=force-xcode-version` here instead of having an extra `resolve-ios-simulator`? we should probably use the same across the board

ArielDemarco:
Bumping Xcode is not a good idea; also main issue is related to something we can avoid having in CI.
Will try to fix this in another PR

PR #977, opened 2026-05-08
[BIT-8227] Adjust blocking and flush behavior for iOS

ArielDemarco:
# Overview
1. Reduce flushing timeout to 0.5s (for android)
2. flush non-blocking on iOS when resigning but not on going to background
3. Remove flushing logic on receiving willTerminate notification


Reflejo (approved):

FranAguilera:
lgtm for android too

PR #981, opened 2026-05-13
[Android] [OkHttp] Enable Fast Fallback when available

FranAguilera:
## What

Resolves BIT-8156

Ensure that we always enable fast fallback for apps using OkHttp 5.0.* and up

## Verification

Verified with gradle-test-app 

Screenshot 2026-05-13 at 15 27 49

- [Gradle Test App session with release build](https://timeline.bitdrift.dev/s/6e80355f-f150-447e-91e0-fc997b8ceba6?utm_source=sdk)

- [Bazel Test App session with release build ](https://timeline.bitdrift.dev/session/ecb4cdab-a609-44da-8450-2361a982cc59?utm_source=sdk&debug=false&utilization=0&expanded=-4068042130858119498#timeline)
 
---

### DIFF
--- a/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/network/okhttp/OkHttpNetwork.kt
+++ b/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/network/okhttp/OkHttpNetwork.kt
@@ -8,6 +8,7 @@
 package io.bitdrift.capture.network.okhttp
 
 import io.bitdrift.capture.CaptureJniLibrary
+import io.bitdrift.capture.common.RuntimeFeature
 import io.bitdrift.capture.network.ICaptureNetwork
 import io.bitdrift.capture.network.ICaptureStream
 import io.bitdrift.capture.network.Jni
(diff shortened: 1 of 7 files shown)


platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/network/okhttp/OkHttpNetwork.kt
```diff
@@ -98,6 +98,7 @@ internal class OkHttpNetwork(
                 ).writeTimeout(timeoutSeconds, TimeUnit.SECONDS)
                 .readTimeout(timeoutSeconds, TimeUnit.SECONDS)
                 .retryOnConnectionFailure(false) // Retrying messes up the write pipe state management, so disable.
+                .enableFastFallbackIfAvailable()
```
murki: do we also need to call this inside our `OkHttpApiClient`?

FranAguilera: good point, applied to the shared okhttp instance

murki (approved):
I actually do think this is a reasonably low-risk approach before we make the decision to bump to v5 (or not)

platform/jvm/gradle-test-app/build.gradle.kts
```diff
@@ -40,6 +40,7 @@ dependencies {
     implementation("com.apollographql.apollo:apollo-runtime:4.1.0")
     implementation("com.jakewharton.timber:timber:5.0.1")
     implementation("com.google.android.material:material:1.8.0")
+    implementation("com.squareup.okhttp3:okhttp:5.0.0-alpha.8")
```
murki: this was ok for testing but should we just use the last official version ourselves?

PR #982, opened 2026-05-13
bump shared core for session changes

mattklein123:
Also change "user" to "entity" to better match the forthcoming feature.


FranAguilera (approved):
Android lgtm, just a minor nit in case if we want to rename the api `registerOpaqueEntityId` to `setOpaqueEntityId` 

Also kudos to @ArielDemarco for verifying iOS :)

PR #984, opened 2026-05-14
Expose SDK status

FranAguilera:
## What 

Resolves BIT-8084

Depends on https://github.com/bitdriftlabs/shared-core/pull/491

Exposing API to retrieve the latest SDK status at any time.

```
val status: SdkStatus = Capture.Logger.getSdkStatus()

data class SdkStatus(
    val initializationState: InitializationState,
    val lastHandshakeTimeMs: Long?,       // epoch millis, null if no handshake yet
    val lastConfigDeliveryTimeMs: Long?,   // epoch millis, null if no config delivery yet
)
enum class InitializationState {
    NOT_STARTED,  // SDK has not been started
    LOADED,       // Logger constructed, log processing not yet begun
    RUNNING,      // Fully running and processing logs
    DISABLED,     // The SDK has been force-disabled by the server (e.g., authentication failure).
}
```

## Verification

| Android gradle-test-app | Android Bazel Release | iOS Swift Sample app | iOS Obj-C Sample App
| -------------------------|  -------------------------|  -------------------------|  -------------------------| 
| Screenshot 2026-05-18 at 14 58 04 | Screenshot 2026-05-18 at 15 15 03 | Screenshot 2026-05-18 at 14 57 46 | Screenshot 2026-05-18 at 14 56 02

---

### DIFF
--- a/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/Capture.kt
+++ b/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/Capture.kt
@@ -281,9 +281,6 @@ object Capture {
 
         /**
          * Returns a point-in-time snapshot of the SDK's operational status.
-         *
-         * This is a synchronous polling API that returns the current state of the SDK.
-         * Returns [InitializationState.NOT_STARTED] if the SDK has not been started.
          */
         @JvmStatic
         fun getSdkStatus(): SdkStatus =
--- a/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/Capture.kt
+++ b/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/Capture.kt
@@ -280,6 +280,17 @@ object Capture {
         fun getPreviousRunInfo(): PreviousRunInfo? = (logger() as? LoggerImpl)?.getPreviousRunInfo()
 
         /**
+         * Returns a point-in-time snapshot of the SDK's operational status.
+         *
+         * This is a synchronous polling API that returns the current state of the SDK.
+         * Returns [InitializationState.NOT_STARTED] if the SDK has not been started.
+         */
+        @JvmStatic
+        fun getSdkStatus(): SdkStatus =
+            (logger() as? LoggerImpl)?.getSdkStatus()
+                ?: SdkStatus(InitializationState.NOT_STARTED, null, null)
+
+        /**
          * Adds a field that should be attached to all logs emitted by the logger going forward.
          * If a field with a given key has already been registered with the logger, its value is
          * overridden with the new value.
--- a/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/SdkStatus.kt
+++ b/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/SdkStatus.kt
@@ -1,0 +1,51 @@
+// capture-sdk - bitdrift's client SDK
+// Copyright Bitdrift, Inc. All rights reserved.
+//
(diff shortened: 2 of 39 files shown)


platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/Capture.kt
```diff
@@ -272,6 +272,15 @@ object Capture {
...
         fun getPreviousRunInfo(): PreviousRunInfo? = (logger() as? LoggerImpl)?.getPreviousRunInfo()
 
+        /**
+         * Returns a point-in-time snapshot of the SDK's operational status.
+         *
+         * This is a synchronous polling API that returns the last known state of the SDK.
+         * Returns `null` if the SDK has not been started.
+         */
+        @JvmStatic
+        fun getSdkStatus(): SdkStatus? = (logger() as? LoggerImpl)?.getSdkStatus()
```
murki: is it going to be status or state?

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/SdkStatus.kt
```diff
@@ -0,0 +1,52 @@
...
+ * has not been started yet (i.e. [SdkState.NotStarted]).
+ *
+ * @property connectionState The current connection state with the backend.
+ * @property lastHandshakeTimeMs The wall-clock time (epoch millis) of the last successful
+ *           handshake, or -1 if no handshake has occurred.
+ * @property lastConfigDeliveryTimeMs The wall-clock time (epoch millis) of the last successful
+ *           config delivery from the backend, or -1 if none has occurred.
+ */
+data class SdkStatus(
+    val connectionState: ConnectionState,
```
murki: don't we need another level of indirection here? like what value would we report here if there was an initialization failure. similarly we'd want to have a status in case this gets called before they call Logger.Start() like NotInitialized or something

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/Capture.kt
FranAguilera: didn't push my recent changes. should be status

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/SdkStatus.kt
FranAguilera: forgot to answer here but with the most recent changes the getStatus API will return NOT_STARTED if the customers never starts the SDK. For JVM/JNI errors during .start call we have the `startResult` callback returning the error

platform/jvm/gradle-test-app/src/main/java/io/bitdrift/gradletestapp/init/BitdriftInit.kt
```diff
@@ -70,6 +70,7 @@ object BitdriftInit {
 
             Capture.Logger.setEntityId(userUuid)
 
+            Capture.Logger.getSdkStatus()
```
murki: where is this value used?

murki (approved):
android side lgtm

platform/jvm/gradle-test-app/src/main/java/io/bitdrift/gradletestapp/init/BitdriftInit.kt
FranAguilera: uops good catch, was a leftover from my early changes for the gradle-test-app. Will clean it up

kattrali (approved):

PR #995, opened 2026-05-20
[BIT-8351] Support Xcode 26 Swiftui rendering changes affecting session replay

ArielDemarco:
# Overview
iOS 26 moved several SwiftUI rendering internals from `UIView` subclasses to pure `CALayer` objects. The existing traversal only walked `UIView` subviews, so those elements became invisible in replay captures.

### DIFF
--- a/platform/swift/source/replay/Replay.swift
+++ b/platform/swift/source/replay/Replay.swift
@@ -43,10 +43,16 @@ package final class Replay {
     ///            more information about the bytes structure.
     package func capture() -> Data {
         let startTime = CFAbsoluteTimeGetCurrent()
+        let result = self.capture(windows: UIApplication.shared.sessionReplayWindows())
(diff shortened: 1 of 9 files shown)


platform/swift/source/replay/Replay.swift
```diff
@@ -82,6 +87,15 @@ package final class Replay {
     private func traverse(into buffer: inout Data, parent: UIView, parentPosition: CGPoint, clipTo: CGRect,
                           ignoreViewType: Bool = false)
     {
+        // iOS 26+ (liquid glass): orphan CALayers (not backed by a UIView) must be traversed before
```
Reflejo, line 90: can we gate this to ios 26 maybe? so we don't add layers for all other cases?

ArielDemarco: Done!

Reflejo (approved):

PR #997, opened 2026-05-21
Adding more options for manual releases

FranAguilera:
## What

Adding more flexibility to have more customization for manual releases. Some of these customizations are:

- Trigger Android only, iOS only, React Native only, or all
- When Android is selected, choose between all artifacts or capture-only (without timber, apollo, plugin)
- Option to add a custom capture-suffix for android release (e.g. that becomes `io.bitdrift:capture-your-suffix:0.23.0`)
- GitHub release tag includes the suffix to avoid conflicts with existing releases

## Verification

Tested with a patch release from local branch [here](https://github.com/bitdriftlabs/capture-sdk/actions/runs/26237507549) by running with these parameters below:

`gh workflow run update_sdk_version.yaml --ref android-release-0.23.0-no-jank-stats -f version=0.23.0 -f emergency=true -f platform=android -f android-artifacts=capture-only -f "capture-suffix=no-jank-stats"`

## Mitigation plan

If any issues with standard release, please revert this PR.
 
---

### DIFF
--- a/.github/workflows/release_gh.yaml
+++ b/.github/workflows/release_gh.yaml
@@ -18,7 +18,7 @@ on:
         default: all
       capture-suffix:
         type: string
-        description: 'Custom Android capture artifact suffix (e.g. "no-jank-stats"). Ignored if platform is ios. Leave empty for standard "capture".'
+        description: 'Custom Android capture artifact suffix including separator (e.g. "-no-jank-stats" -> io.bitdrift:capture-no-jank-stats). Ignored if platform is ios. Leave empty for standard "capture".'
         required: false
         default: ''
     secrets:
@@ -51,7 +51,7 @@ on:
           - capture-only
       capture-suffix:
         type: string
-        description: 'Custom Android capture artifact suffix (e.g. "no-jank-stats"). Ignored if platform is ios. Leave empty for standard "capture".'
+        description: 'Custom Android capture artifact suffix including separator (e.g. "-no-jank-stats" -> io.bitdrift:capture-no-jank-stats). Ignored if platform is ios. Leave empty for standard "capture".'
         required: false
         default: ''
 jobs:
@@ -115,10 +115,7 @@ jobs:
         run: sudo apt-get install -y clang
       - name: Build AAR + artifacts
         run: |
-          artifact_name="capture"
-          if [ -n "${{ inputs.capture-suffix }}" ]; then
-            artifact_name="capture-${{ inputs.capture-suffix }}"
-          fi
+          artifact_name="capture${{ inputs.capture-suffix }}"
           ./tools/android_release.sh "$VERSION" "$artifact_name"
         env:
           VERSION: ${{ inputs.version }}
@@ -151,7 +148,6 @@ jobs:
       contents: write
     name: GitHub release
     runs-on: ubuntu-latest
-    if: ${{ always() && !failure() && !cancelled() }}
     needs: [
       "ios-release-build",
       "build-ios-example-apps",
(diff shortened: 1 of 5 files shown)


.github/workflows/release_gh.yaml
```diff
@@ -6,6 +6,21 @@ on:
...
+        required: false
+        default: all
+      android-artifacts:
+        type: string
+        description: Which Android artifacts to publish (all, capture-only)
+        required: false
+        default: all
+      capture-suffix:
+        type: string
+        description: 'Custom Android capture artifact suffix (e.g. "no-jank-stats"). Ignored if platform is ios. Leave empty for standard "capture".'
```
murki: shouldn't this example be `"-no-jank-stats`" ? or are we always appending the `-` ourselves?

.github/workflows/release_gh.yaml
```diff
@@ -72,20 +114,29 @@ jobs:
       - name: Install Clang
         run: sudo apt-get install -y clang
       - name: Build AAR + artifacts
-        run: ./tools/android_release.sh "$VERSION"
+        run: |
+          artifact_name="capture"
+          if [ -n "${{ inputs.capture-suffix }}" ]; then
+            artifact_name="capture-${{ inputs.capture-suffix }}"
```
murki: oh I see now that we're appending the "-", it's a nit but I don't think we should append anything (what if we want to use underscore "_" instead ?)

.github/workflows/release_gh.yaml
```diff
@@ -100,6 +151,7 @@ jobs:
       contents: write
     name: GitHub release
     runs-on: ubuntu-latest
+    if: ${{ always() && !failure() && !cancelled() }}
```
murki, line 161: why is this check needed?

.github/workflows/release_gh.yaml
```diff
@@ -168,22 +225,36 @@ jobs:
...
+          VERSION: ${{ inputs.version }}
+
         # Upload artifacts to the newly created release.
-        # Prefix release version with "v".
+        # Prefix release version with "v". If capture-suffix is set, append it to the tag.
       - name: Create release
         run: |
-          gh release create "v$VERSION" \
+          tag="v$VERSION"
+          if [ -n "${{ inputs.capture-suffix }}" ]; then
```
murki: I'm confused; here we seem to be appending the suffix to the version tag, isn't the artifact name which should be affected?

.github/workflows/release_gh.yaml
FranAguilera: oh yeah good catch, will update to allow that flexibility

.github/workflows/release_gh.yaml
FranAguilera: yeah but this is only for the release published in [here](https://github.com/bitdriftlabs/capture-sdk/releases/tag/v0.23.0) to avoid confusion. But actually for a custom suffix release i will probalby see if i can skip this step

.github/workflows/release_gh.yaml
FranAguilera: uops, will remove

murki (approved):
lgtm, we should just verify nothing in our normal release flow is broken

.github/workflows/release_gh.yaml
FranAguilera: answered below https://github.com/bitdriftlabs/capture-sdk/pull/997#discussion_r3283850565

.github/workflows/release_gh.yaml
FranAguilera: updated now

FranAguilera:
Will be doing more manual testing and merging tomorrow morning if everything looks good

FranAguilera:
lol forgot to merge

PR #999, opened 2026-05-27
[Android] [BIT-8406] Add memory pressure level state to fatal issues

FranAguilera:
## What

Resolves BIT-8406

iOS equivalent here https://github.com/bitdriftlabs/capture-sdk/pull/1001/changes

Depends on https://github.com/bitdriftlabs/shared-core/pull/494 Adding memory pressure level state as global state field to fatal issue reports (will be under `_memory_pressure_level`)

- Captures the latest memory pressure level via ResourceUtilization tick
- Upon a JVM crash capture the latest memory pressure level state

Follow up to store this info as well on the fbs report will be done in BIT-8474

## Verification

See [session](https://timeline.bitdrift.dev/session/5db36dd5-2294-4238-992a-48b905becbd0?utilization=1) 

---

NOTE: Changelog to be updated when the iOS changes are done too


platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/attributes/NetworkAttributes.kt
```diff
@@ -159,7 +159,6 @@ internal class NetworkAttributes(
         permission: String,
     ): String {
         // We'll only attempt to get this value if permission has been granted
-        // TODO: Use androidx.core.content.ContextCompat - fix adding dep to "androidx.core:core-ktx:1.8.0"
```
FranAguilera, line 162: Unrelated, but cleaning up this dead comment

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/LoggerImpl.kt
```diff
@@ -593,6 +594,10 @@ internal class LoggerImpl(
         )
     }
 
+    override fun logMemoryPressureLevel(level: MemoryPressureLevel) {
```
murki: we should definitely not call this method `logX()` since that convention implies that we're actually emitting a new log that gets put in the buffers. We're calling `notify_memory_pressure()` so maybe name it that here as well?

FranAguilera: oh yeah good catch, I think at some point i had a notify in the signature but went back to this. Will update soon

murki (approved):
lgtm

murki (approved):
lgtm

PR #1005, opened 2026-06-03
[Android] Fix ConcurrentModification at ReplayDecorations.addDecorations

FranAguilera:
## What

Resolves BIT-8521

Fixes a rare ConcurrentModificationException on API < 29 when windows change (toasts, dialogs, popups) while session replay is capturing the screen on a background thread.

```
Fatal Exception: java.util.ConcurrentModificationException
   at java.util.ArrayList$Itr.next(ArrayList.java:860)
   at io.bitdrift.capture.replay.internal.ReplayDecorations.addDecorations(ReplayDecorations.kt:54)
   at io.bitdrift.capture.replay.internal.ReplayCaptureEngine$captureScreen$2.run(ReplayCaptureEngine.kt:61)

```

- On API < 29 `WindowManager.getAllRootViews()` uses reflection to read `WindowManagerGlobal.mViews` directly and returning a live reference to the internal list. If the main thread adds or removes a window while the background thread iterates, the iterator can throw the ConcurrentModificationException.

- On API >=29 is unaffected because `WindowInspector.getGlobalWindowViews()` already returns a copy internally. So the fix for the fallback api is to also do a defensive copy.

## Verification

- For the gradle-test-app, created a new entry on StressTab to reproduce the original issue.
- Run on API <29

Screenshot 2026-06-03 at 14 16 21

- Without defensive copy fix the crash is reproduced each time. [See report here](https://explorations.bitdrift.dev/issues/18226480752552072537/0e9627c3-5e04-4e51-83c0-a2985ee61e7c?index=0&range=7d)
- When the fix is apply, the session replay capture continues without crashing.

---

### DIFF
--- a/platform/jvm/common/src/main/kotlin/io/bitdrift/capture/common/WindowManager.kt
+++ b/platform/jvm/common/src/main/kotlin/io/bitdrift/capture/common/WindowManager.kt
@@ -55,7 +55,10 @@ class WindowManager(
         }
         try {
             @Suppress("UNCHECKED_CAST")
-            return getWindowViews.get(windowManagerGlobal) as List<View>
+            // Match WindowInspector.getGlobalWindowViews(), which returns a defensive copy.
+            // This reflection fallback reads WindowManagerGlobal.mViews directly, so returning it
+            // can cause ConcurrentModificationException for consumers if windows change later.
+            return ArrayList(getWindowViews.get(windowManagerGlobal) as List<View>)
         } catch (e: Throwable) {
             errorHandler.handleError("Failed to retrieve windows", e)
             return emptyList()
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,7 +29,7 @@
 
 **Fixed**
 
-- Nothing yet!
+- Session replay: fix a rare `ConcurrentModificationException` on API < 29 when windows change while screen capture is in progress.
 
 ### iOS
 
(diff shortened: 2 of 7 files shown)


murki (approved):

platform/jvm/common/src/main/kotlin/io/bitdrift/capture/common/WindowManager.kt
```diff
@@ -55,7 +55,10 @@ class WindowManager(
         }
         try {
             @Suppress("UNCHECKED_CAST")
-            return getWindowViews.get(windowManagerGlobal) as List<View>
+            // Match WindowInspector.getGlobalWindowViews(), which returns a defensive copy.
+            // This reflection fallback reads WindowManagerGlobal.mViews directly, so returning it
+            // can cause ConcurrentModificationException for consumers if windows change later.
+            return ArrayList(getWindowViews.get(windowManagerGlobal) as List<View>)
```
murki, line 61: no need to synchronize a lock?

FranAguilera: No lock needed here because ArrayList copies the items in one shot at the native level (without iterator). The original crash happens in `addDecorations` which loops over the mutable list one item at a time, so the longer that iteration runs the higher the chance for views to change in that process and to trigger Concurrent Modification exception.

If we still wanted proper synchronization, we'd need to use WindowManagerGlobal's own [mLock](https://android.googlesource.com/platform/frameworks/base/+/master/core/java/android/view/WindowManagerGlobal.java#145), a custom lock on our WindowManager wouldn't help because AOSP WindowManagerGlobal doesn't acquire it when mutating mViews only mLock is shared between both sides. Acquiring it would require additional reflection which im not sure is worth the added complexity/risk.

PR #1008, opened 2026-06-03
andr, sdk: refactor session replay to look for keyboard on single UI pass

murki:
Main changes in `ReplayCaptureEngine` and `ReplayParser`.

Before the traversal was split in two:
1. Main thread: create the ReplayRect tree
2. bg thread: Filter/add layers (remove ignored views, add screen/window, add keyboard)

Now it's been condensed in the single first pass. Also deleted a bunch of confusing extra classes (`ReplayDecoration`, `ReplayFilter`, `FilteredCapture` typealias).

Updated `examples/android/MainActivity.kt` to use the new `List<ReplayRect>` signature in the `onScreenCaptured` override, replacing the removed `FilteredCapture` typealias.

test session: https://timeline.bitdrift.dev/session/f21546b4-95de-485c-a9a1-ceb485292d99

Screenshot 2026-06-03 at 5 33 00 PM

---


platform/jvm/replay/src/main/kotlin/io/bitdrift/capture/replay/internal/ReplayCaptureEngine.kt
```diff
@@ -55,16 +61,25 @@ internal class ReplayCaptureEngine(
...
                 SessionReplayController.L.d("Screen Captured: $replayCaptureMetrics")
-                completion(encodedScreen, screen, replayCaptureMetrics)
+                completion(encodedScreen, filteredCapture, replayCaptureMetrics)
             }
         }
     }
+
+    private fun filter(capture: List<ReplayRect>): List<ReplayRect>? {
+        // This capture is identical to the previous one, filter it out
+        return if (capture == previousCapture) {
```
FranAguilera, line 78: Before we were ignoring when the app is backgrounded

From main: 

```
// This capture is identical to the previous one, or is empty, filter it out
// One interesting case when capture is empty is when the application is backgrounded.
return if (filteredCapture == previousCapture || filteredCapture.isEmpty()) {
    null
} else {
    previousCapture = filteredCapture
    filteredCapture
}
```

platform/jvm/replay/src/main/kotlin/io/bitdrift/capture/replay/internal/ReplayCaptureEngine.kt
```diff
@@ -28,12 +28,18 @@ internal class ReplayCaptureEngine(
...
+        ReplayParser(
+            sessionReplayConfiguration,
+            errorHandler,
+            windowManager,
+            displayManager,
+        ),
     private val replayEncoder: ReplayEncoder = ReplayEncoder(),
     private val clock: IClock = DefaultClock.getInstance(),
 ) {
+    private var previousCapture: List<ReplayRect>? = null
```
FranAguilera, line 41: Looks like this is only accessed from the background thread so probably ok for now to not make it thread safe

platform/jvm/replay/src/main/kotlin/io/bitdrift/capture/replay/internal/ReplayParser.kt
```diff
@@ -7,19 +7,21 @@
...
 import io.bitdrift.capture.common.IWindowManager
 import io.bitdrift.capture.replay.ReplayCaptureMetrics
+import io.bitdrift.capture.replay.ReplayType
 import io.bitdrift.capture.replay.SessionReplayConfiguration
 import io.bitdrift.capture.replay.SessionReplayController
 import io.bitdrift.capture.replay.internal.mappers.ViewMapper
 
-internal typealias Capture = List<List<ReplayRect>>
-
 internal class ReplayParser(
```
FranAguilera, line 20: Wondering if we should rename this now as is doing more things now than just parsing (e.g. computeDisplayRect(), compute IME overlay, etc)

platform/jvm/replay/src/main/kotlin/io/bitdrift/capture/replay/internal/ReplayCaptureEngine.kt
```diff
@@ -55,16 +61,25 @@ internal class ReplayCaptureEngine(
             }
 
         executor.execute {
-            captureFilter.filter(timedValue.value)?.let { filteredCapture ->
+            filter(timedValue.value)?.let { filteredCapture ->
                 replayCaptureMetrics.parseDuration = timedValue.duration
                 replayCaptureMetrics.viewCountAfterFilter = filteredCapture.size
-                val screen = captureDecorations.addDecorations(filteredCapture)
```
FranAguilera, line 61: This was happening in background thread, we should measure the impact on jank for this change (I can take a look)

platform/jvm/replay/src/main/kotlin/io/bitdrift/capture/replay/internal/ReplayCaptureEngine.kt
murki: yea

platform/jvm/replay/src/main/kotlin/io/bitdrift/capture/replay/internal/ReplayParser.kt
murki: fair

platform/jvm/replay/src/main/kotlin/io/bitdrift/capture/replay/internal/ReplayCaptureEngine.kt
murki: yeah in particular there will be 1 extra check for every root view to look for the keyboard, I'd expect it to be very low overhead but probably not 0

platform/jvm/replay/src/main/kotlin/io/bitdrift/capture/replay/internal/ReplayCaptureEngine.kt
murki: yeah good point, this might need more testing

FranAguilera:
Will take a look this week

PR #1016, opened 2026-06-11
[Android] Fix auto instrumentation for webviews subclasses

FranAguilera:
## Problem

Part of BIT-8545

For customers using WebViews subclasses the bytecode manipulation won't work as we are checking for exact WebView match. The fix is to check if the class is a child of `android.webkit.WebView` so we can inject the instrument call at loadUrl

This is now a problem for RN integration subclass `com.reactnativecommunity.webview.RNCWebView`

This will be used in https://github.com/bitdriftlabs/capture-es/pull/261

## Verification

- JUnit tests
- Manually verified with gradle-test-app, CustomWebView and local maven plugin publish. See [demo session here](https://timeline.bitdrift.dev/session/2db28b66-d25c-4c18-8d22-ef6a0e4d45ad?utm_source=sdk&utilization=0&expanded=-3163060467042078438). 

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,7 @@
 **Fixed**
 
 - Ensure the Gradle plugin adds the OkHttp tracing interceptor last during automatic instrumentation.
+- Match Android `WebView` subclass `loadUrl(...)` callsites during automatic WebView instrumentation.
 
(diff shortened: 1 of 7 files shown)


murki (approved):
nice

PR #1024, opened 2026-06-18
add additional fields to reports

kattrali:
(ios tests are going to fail until https://github.com/bitdriftlabs/shared-core/pull/525 is added in)


platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/processor/ILocaleProvider.kt
```diff
@@ -0,0 +1,18 @@
...
+//
+// Use of this source code is governed by a source available license that can be found in the
+// LICENSE file or at:
+// https://polyformproject.org/wp-content/uploads/2020/06/PolyForm-Shield-1.0.0.txt
+package io.bitdrift.capture.reports.processor
+
+/**
+ * Provide a value for device locale
+ */
+interface ILocaleProvider {
```
FranAguilera: Can we expand instead the existing [IClientAttributes](https://github.com/bitdriftlabs/capture-sdk/blob/4d2899512beaf41ba024f58f63e6608feb80dc76/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/attributes/IClientAttributes.kt#L12). e.g. already has full [locale](https://github.com/bitdriftlabs/capture-sdk/blob/4d2899512beaf41ba024f58f63e6608feb80dc76/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/attributes/IClientAttributes.kt#L53)?

kattrali: yeah that makes sense. I think I need to wrap my head around what's going on with locale caching there 😅

FranAguilera: yeah the TLDR is to avoid unnecessary allocation for static fields that won't likely change.

This was added in here https://github.com/bitdriftlabs/capture-sdk/pull/558

ArielDemarco (approved):
iOS LGTM

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/attributes/ClientAttributes.kt
```diff
@@ -59,7 +60,9 @@ internal class ClientAttributes(
 
     override val model: String = Build.MODEL
 
-    override val locale: String by lazy { getCurrentLocale() }
+    override val locale: String by lazy { getCurrentLocale()?.toString() ?: UNKNOWN_FIELD_VALUE }
```
FranAguilera: This was a previous oversight but the exposed locale shouldn't be lazy as technically the locale could be changed during app runtime (e.g. via device settings) and the value will be stale

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/attributes/ClientAttributes.kt
```diff
@@ -150,16 +153,9 @@ internal class ClientAttributes(
         }
 
     /**
-     * Gets the current locale string.
+     * Gets the current locale.
      */
-    private fun getCurrentLocale(): String {
-        val locales = ConfigurationCompat.getLocales(resources.configuration)
-        return if (!locales.isEmpty) {
```
FranAguilera, line 157: Haven't check the internals of `.get` from `getLocales` yet but we are changing a bit the guard we had originally.

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/attributes/ClientAttributes.kt
```diff
@@ -59,7 +60,9 @@ internal class ClientAttributes(
 
     override val model: String = Build.MODEL
 
-    override val locale: String by lazy { getCurrentLocale() }
+    override val locale: String by lazy { getCurrentLocale()?.toString() ?: UNKNOWN_FIELD_VALUE }
+
+    override val localeCountryCode: String by lazy { getCurrentLocale()?.country ?: UNKNOWN_FIELD_VALUE }
```
FranAguilera: Let's make not this lazy as well, to get the most up to date value to any consumers

FranAguilera:
```suggestion
    override val localeCountryCode: String
        get() = currentCountryCode()
```

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/attributes/ClientAttributes.kt
FranAguilera:
```suggestion
    override val locale: String
        get() = currentLocaleValue()
```

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/attributes/ClientAttributes.kt
```diff
@@ -117,7 +120,7 @@ internal class ClientAttributes(
         val configDiff = currentConfig.diff(cachedConfiguration)
 
         if (cachedLocale == null || (configDiff and ActivityInfo.CONFIG_LOCALE == ActivityInfo.CONFIG_LOCALE)) {
-            val updatedLocale = getCurrentLocale()
+            val updatedLocale = getCurrentLocale()?.toString() ?: UNKNOWN_FIELD_VALUE
```
FranAguilera:
```suggestion
            val updatedLocale = currentLocaleValue()
```

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/attributes/ClientAttributes.kt
```diff
@@ -150,16 +153,9 @@ internal class ClientAttributes(
...
      */
-    private fun getCurrentLocale(): String {
-        val locales = ConfigurationCompat.getLocales(resources.configuration)
-        return if (!locales.isEmpty) {
-            locales[0].toString()
-        } else {
-            UNKNOWN_FIELD_VALUE
-        }
-    }
+    private fun getCurrentLocale(): Locale? = ConfigurationCompat.getLocales(resources.configuration).get(0)
```
FranAguilera, line 160:
```suggestion
    private fun getCurrentLocale(): Locale? {
        val locales = ConfigurationCompat.getLocales(resources.configuration)
        return if (locales.isEmpty) {
            null
        } else {
            locales[0]
        }
    }
```

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/attributes/ClientAttributes.kt
```diff
@@ -150,16 +153,9 @@ internal class ClientAttributes(
...
 
     /**
```
FranAguilera, line 162:
```suggestion
         private fun currentLocaleValue(): String = getCurrentLocale()?.toString() ?: UNKNOWN_FIELD_VALUE

    private fun currentCountryCode(): String =
        getCurrentLocale()?.country?.takeIf { it.isNotEmpty() } ?: UNKNOWN_FIELD_VALUE
    /**
```

FranAguilera (changes requested):
Looking good but mainly these regarding Android chnges:

1. Remove lazy for the exposed `locale` and `localeCountryCode` to get the most up to date value to consumers in case of any configuration change.

2. For accessing/sanitizing country code would be good to check for empty string as country code docs mentions `* @return The country/region code, or the empty string if none is defined.`

3. I see we are adding this info now for Native/JVM crash but I don't see in the IClientAttributes passed for generating the ANR report uses the new field in shared-core added in https://github.com/bitdriftlabs/shared-core/pull/525/changes

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/attributes/ClientAttributes.kt
kattrali: I'd changed it since I checked and [it returned `@Nullable Locale`](https://developer.android.com/reference/androidx/core/os/LocaleListCompat#get(int)) ([source](https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:core/core/src/main/java/androidx/core/os/LocaleListCompatWrapper.java;l=48;bpv=1;bpt=1))

FranAguilera (approved):
Can you do a quick manual test with gradle-test-app before merging?

PR #1029, opened 2026-06-19
[BIT-8763] New Crash Reporter Crate (MVP)

ArielDemarco:
# Overview
- New crate to do cross-platform crash handling
- Handle `NSException`s and persist them using `mmap`


platform/crash/src/previous.rs
```diff
@@ -0,0 +1,293 @@
...
+  }
+}
+
+fn parse_nsexception(raw: &RawNSExceptionPayload) -> NSExceptionCrashInfo {
+  NSExceptionCrashInfo {
+    name: raw.name,
+    reason: raw.reason,
+    call_stack: parse_nsexception_call_stack(&raw.call_stack),
+  }
+}
```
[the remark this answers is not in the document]
ArielDemarco: I like the feedback; but I dont like the solution; probably, if it's corrupted, it's better to just zero the name/reason instead.

platform/crash/src/coordinator.rs
```diff
@@ -0,0 +1,154 @@
...
+    }
+  }
+
+  pub(crate) const fn previous_crash_state(&self) -> &PreviousCrashState {
+    &self.previous_crash_state
+  }
+}
+
+#[cfg(test)]
+mod tests {
```
snowp: Typically we put the test files in a separate mod, you can try pulling in the Rust guidelines here https://github.com/bitdriftlabs/ai-instructions

platform/crash/src/monitors/nsexception.rs
```diff
@@ -0,0 +1,314 @@
...
+}
+
+fn previous_handler() -> Option<ExceptionHandler> {
+  let previous = PREVIOUS_HANDLER.load(Ordering::Acquire);
+  if previous == 0 {
+    None
+  } else {
+    Some(unsafe { std::mem::transmute::<usize, ExceptionHandler>(previous) })
+  }
+}
```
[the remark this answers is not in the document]
ArielDemarco: Clippy already passes here, and changing this would be a cleanup rather than a functional fix. I'd prefer to do a more deliberate handler-storage cleanup instead of just swapping cast patterns. But doesn't seem necessary right now.

.github/workflows/crash_reporter.yaml
```diff
@@ -0,0 +1,75 @@
...
+        id: crash_check
+        run: |
+          ./ci/files_changed.sh '^(platform/crash/|Cargo\.toml$|Cargo\.lock$|Cargo\.Bazel\.lock$|MODULE\.bazel$|MODULE\.bazel\.lock$|\.cargo/|\.github/workflows/crash_reporter\.yaml$)'
+        continue-on-error: true
+
+      - name: Determine if tests should run
+        id: check_changes
+        run: |
+          crash_status="${{ steps.crash_check.outputs.check_result }}"
+
```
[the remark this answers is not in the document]
ArielDemarco: wrong; the mismatch you pointed out does not exist in the current file.

platform/crash/src/monitors/nsexception.rs
```diff
@@ -0,0 +1,314 @@
...
+
+#[cfg(test)]
+mod tests {
```
mattklein123: nit: we have a convention that tests go in a parallel test file for rust. In general make sure your agent is pointing to the rust instructions in here: https://github.com/bitdriftlabs/ai-instructions. Have it go through and do a pass on the PR.

mattklein123 (changes requested):
Nice, glad to see this happening.

Can you take a quick pass to adhere to the rust instructions and also add more comments and then I can take a real look? Thanks.

platform/crash/src/monitors/nsexception.rs
ArielDemarco: oh, wasn't aware of this! will take a look

platform/crash/src/schema.rs
```diff
@@ -0,0 +1,109 @@
...
+//
+// Use of this source code is governed by a source available license that can be found in the
+// LICENSE file or at:
+// https://polyformproject.org/wp-content/uploads/2020/06/PolyForm-Shield-1.0.0.txt
+
+pub(crate) const MAGIC: u64 = u64::from_be_bytes(*b"BDCRASH\0");
+pub(crate) const VERSION: u32 = 1;
+pub(crate) const NS_EXCEPTION_NAME_CAPACITY: usize = 128;
+pub(crate) const NS_EXCEPTION_REASON_CAPACITY: usize = 512;
+pub(crate) const MAX_NS_EXCEPTION_CALL_STACK_FRAMES: usize = 64;
```
mattklein123: Is this enough?

platform/crash/src/monitors/nsexception.rs
```diff
@@ -0,0 +1,153 @@
...
+
+  if let Some(snapshot) = snapshot {
+    record_exception_snapshot(&snapshot);
+  }
+
+  chain_previous(exception);
+}
+
+fn extract_exception_snapshot(exception: &NSException) -> ExceptionSnapshot {
+  let return_addresses = exception.callStackReturnAddresses();
```
mattklein123, line 152: So in this case it walks the stack for you? That's nice.

mattklein123 (approved):
This LGTM at a high level. I have questions about how we are going to feature flag this but we can talk about that in the integration PR.

platform/crash/src/schema.rs
ArielDemarco: I had the same question while wiring this up. For the MVP, I picked 64 as a pragmatic bound to keep the persisted record small while still capturing enough context to make the crash useful on next launch. However, I expect to have more info after implementing the full integration with iOS, specially once I can inspect stack traces from longer main-thread crashes (for example, SwiftUI-heavy ones). If `64` turns out to be too little in practice, I can raise the cap before merging

platform/crash/src/monitors/nsexception.rs
ArielDemarco: yep! For `NSException` we can rely on `callStackReturnAddresses()` as it gives you a snapshot of the exception stack without doing our own unwinding right now. That said, this is only from the crashing thread

platform/crash/src/previous.rs
```diff
@@ -0,0 +1,158 @@
...
+  // Persisted strings are expected to be null-terminated. If a terminator is missing, treat the
+  // full field as invalid rather than guessing where a truncated or corrupt string should end.
+  if !bytes.contains(&0) {
+    bytes.fill(0);
+  }
+}
+
+fn parse_nsexception_call_stack(raw: &RawNSExceptionCallStack) -> NSExceptionCallStack {
+  let max_frame_count =
+    u16::try_from(schema::MAX_NS_EXCEPTION_CALL_STACK_FRAMES).unwrap_or(u16::MAX);
```
kattrali: > `u16::MAX`

is this a reasonable thing to do here? Maybe make the schema value a `u16` (since then its infallible to scale up to `usize` when needed)? or just drop in 50 or something?

kattrali (approved):
looking good, one small q

platform/crash/src/schema.rs
ArielDemarco: Bumped them

platform/crash/src/previous.rs
ArielDemarco: Addressed [here](https://github.com/bitdriftlabs/capture-sdk/pull/1042/commits/35ee5623766def3a3830a9cf47267ef89c45ab7b) (next PR). Changed `MAX_NS_EXCEPTION_CALL_STACK_FRAMES` to be a `u16` an cast to `usize` when needed

platform/crash/src/ffi.rs
```diff
@@ -0,0 +1,140 @@
...
+  if COORDINATOR.get().is_none() {
+    return -1;
+  }
+
+  i8::from(previous_crash_state().is_some_and(|state| state.did_crash))
+}
+
+/// Return the cached previous-launch crash timestamp, or `0` when no crash state is available.
+#[no_mangle]
+pub extern "C" fn capture_bitdrift_crash_cached_timestamp() -> u64 {
```
snowp, line 106: If we follow previous patterns we'd be using JNI straight to Rust so we wouldn't be going through these FFI layers. Fine as is, but I think this is more like the objective-c boundary than cross-platform FFI

platform/crash/src/previous.rs
```diff
@@ -0,0 +1,158 @@
...
+
+pub(crate) fn read_previous_state_from_bytes(bytes: &[u8]) -> PreviousCrashState {
+  if bytes.len() < size_of::<CrashRecord>() {
+    return PreviousCrashState::default();
+  }
+
+  // This function defines the acceptance policy for a persisted crash record. The record is
+  // mmap-backed, so reads on a later launch must tolerate truncated or partially-corrupt contents
+  // instead of assuming the bytes were produced by a clean shutdown.
+  let raw: CrashRecord = unsafe { read_unaligned(bytes.as_ptr().cast::<CrashRecord>()) };
```
snowp, line 116: Up to you if you want to change this now but do we want to crc this record as well? I was just dealing with seemingly random memory corruption elsewhere, might be a quick way to avoid junk data?

snowp (approved):
some thoughts, nothing blocking

platform/crash/src/coordinator.rs
ArielDemarco: Done! I thought I did it, but probably missed this file

platform/crash/src/previous.rs
ArielDemarco: done; used the same crate we use in shared-core to do it (`crc32fast`)

PR #1031, opened 2026-06-23
[Android] Reduce report file size 

FranAguilera:
## Problem

If a fatal issue report contains too many stack frames, such as crashes with thousands of threads, the persisted report can become large enough that we risk failing to store it and dropping the report.

## Solution

Resolves BIT-8852

Depends on https://github.com/bitdriftlabs/shared-core/pull/541 and https://github.com/bitdriftlabs/shared-core/pull/543

Mainly these 2 parts

1. Use the builtin fbs string memoization (e.g. `createSharedString`) 
- This deduplicates repeated strings such as symbol names, file names, thread names, image IDs, and other repeated stack trace metadata within a single report.
- In a test report with 7000 threads. This reduces the persisted report size from about 6300 KB to about 2200 KB.

2. Added custom memoization for repeated stack trace structure
- Reuse identical serialized frame offsets when the same frame appears multiple times.
- Reuse identical serialized stack trace vector offsets when multiple threads share the same stack.
- Also for the same 7000-thread case, this reduces the report size from about 2200 KB to about 344 KB.

3. Additionally it brings in Android ANR reduction from shared-core [here](https://github.com/bitdriftlabs/shared-core/pull/541) 

## Verification

1. Go to gradle-test-app, Stress tab, create 7000 threads and trigger JVM/Native crash report. Verify the report is ingested on a low end device. [See report here](https://explorations.bitdrift.dev/issues/11985232642810234142/dc9f8244-cfa3-4127-950c-9bd496919855?utm_source=sdk).
2. Verify manually the *.cap report file size.
3. Verified other regular fatal issues are not dropped. See [here](https://explorations.bitdrift.dev/issues?f=app_version...is...3.0)

## Killswitch flag

`client_feature.android.optimize_fatal_issue_report_size`, enabled by default

## Follow-up

- BIT-8854: For iOS, see if we can get similar improvements.

---


platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/processor/NativeCrashProcessor.kt
```diff
@@ -206,21 +218,21 @@ internal object NativeCrashProcessor {
...
         val signalName = tombstone.signalInfo.name
         val errorName = signalName.ifEmpty { description ?: "Native crash" }
-        val reason = builder.createString(errorName)
+        val reason = builder.createSharedString(errorName)
         val causeText =
             tombstone.causesList.firstOrNull()?.humanReadable ?: tombstone.abortMessage.ifEmpty {
                 signalsByNumber.values.firstOrNull { it.name == signalName }?.description ?: "Native crash"
             }
-        val cause = builder.createString(causeText)
+        val cause = builder.createSharedString(causeText)
```
snowp: I'm not sure if there is extra overhead per shared string vs not shared string, but for stuff like this (exception cause) it seems unlikely that it will be repeated? If there is overhead then it seems optimal to only use shared strings for values that are typically repeated within the report

platform/jvm/capture/src/test/kotlin/io/bitdrift/capture/reports/processor/IssueReporterProcessorTest.kt
```diff
@@ -151,6 +143,65 @@ class IssueReporterProcessorTest {
...
+            }
+
+        val report =
+            persistJvmCrashAndGetReport(
+                callerThread = callerThread,
+                throwable = RuntimeException("crash"),
+                allThreads = allThreads,
+            )
+        assertThat(report.threadDetails?.threadsLength).isEqualTo(allThreads.size)
+        assertThat(report.threadDetails?.threads(0)?.stackTraceLength).isEqualTo(sharedStackTrace.size)
```
snowp: do you want these to validate the contents of the stack trace as well? Seems like a stronger test than just checking the number of threads 

Same for some of the other tests

snowp (approved):
lgtm, some minor comments but nothing blocking

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/processor/NativeCrashProcessor.kt
FranAguilera: good catch, yes there is a small overhead after checking internal impl. Will update this portion to keep as before and retest

platform/jvm/capture/src/test/kotlin/io/bitdrift/capture/reports/processor/IssueReporterProcessorTest.kt
FranAguilera: good call, will update before merging

PR #1035, opened 2026-06-26
[BIT-8890] Create a flag to manage blocking timeout in `willTerminateNotification` log

ArielDemarco:


platform/swift/source/RuntimeVariable.swift
```diff
@@ -78,6 +78,11 @@ extension RuntimeVariable<Bool> {
         defaultValue: true
     )
 
+    static let blockingWillTerminateLog = RuntimeVariable(
+        name: "client_feature.ios.blocking_will_terminate_log",
+        defaultValue: true
```
mattklein123: Should we default to false? This seems to match what the rest of the code does now?

ArielDemarco: I was going to leave it as true so it could act as a control variable, but considering it seems to be causing issues, I'll switch it to false

mattklein123: I could go either way we can see what people think

FranAguilera (approved):

PR #1038, opened 2026-06-26
[Android] Add more benchmarks tests

FranAguilera:
Adding a few more benchmarks for other common Capture.Logger public APIs

---


platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/Capture.kt
```diff
@@ -326,8 +326,8 @@ object Capture {
          * a user is exposed to a specific feature flag variant in your application. The exposure
          * is recorded with a timestamp and tracked for the duration of the process.
          *
-         * @param name the name of the flag being exposed
-         * @param variant the variant of the flag being exposed
+         * @param name the name of the flag being exposed.
```
FranAguilera, line 329: Just to trigger full Android ci

murki (approved):

PR #1040, opened 2026-06-26
Reduce AppExit Blocking Duration

ArielDemarco:


Reflejo (approved):

murki:
@ArielDemarco can you add a little note in the release notes for iOS that we made fixes to reduce the chances of bla bla?

ArielDemarco:
> @ArielDemarco can you add a little note in the release notes for iOS that we made fixes to reduce the chances of bla bla?

Done!

CHANGELOG.md
```diff
@@ -41,7 +41,7 @@
 
 **Changed**
 
-- Nothing yet!
+- Reduced the default wait time for the AppExit log during app termination from 1s to 500ms
```
murki: Could we reword it to make it more customer centric? Like what it achieves "Added a fix that should reduce the chances of bitdrift being executing when the app is about to be terminated" or something like that

ArielDemarco: Done!

murki (approved):
ty!

PR #1044, opened 2026-07-01
[BIT-8758] PreviousRunInfo revamp

ArielDemarco:
# Overview

Changed how we check the previous run's exit status on iOS (aka `Capture.previousRunInfo`).

## Description
Now `previousRunInfo` not only exposes `hasFatallyTerminated`, but also provides a broader concept: `wasCleanExit`. When the user manually terminates the app, we consider it a clean exit. Unclean exits include crashes, but there are other reasons why a run could be considered unclean, which are exposed through `PreviousRunStatus`.

Whenever we do not have enough information, instead of relying on heuristics, we explicitly mark the unclean exit as `.unknown`.

The resolution priority is:
```
fatalCrash > cleanExit > appUpdate > osUpdate > (reboot*) > debuggerAttached > unknown
```

## Open questions

1. **Forced reboot status**: the boot-time-mismatch check already detects "rebooted, no crash/clean-exit" but reports plain `.unknown`. Worth a dedicated case; detection logic already exists.
2. **Watchdog kills** : `DiagnosticEventReporter` already flags these via MetricKit (`crash_is_hang_termination`) but doesn't feed `previousRunInfo`. MetricKit delivery is async/delayed, so: keep `previousRunInfo` startup-only and treat this separately, or allow a post-hoc override of an already-resolved `.unknown`?


platform/swift/source/reports/previousRunInfo/PreviousRunInfo.swift
```diff
@@ -0,0 +1,35 @@
...
+// Use of this source code is governed by a source available license that can be found in the
+// LICENSE file or at:
+// https://polyformproject.org/wp-content/uploads/2020/06/PolyForm-Shield-1.0.0.txt
+
+import Foundation
+
+/// Snapshot of the previous app run status.
+public struct PreviousRunInfo: Equatable {
+    /// The best deterministic launch-time status for the previous app run.
+    public let status: PreviousRunStatus
```
murki: we should try to align with Android here https://docs.bitdrift.io/sdk/features/fatal-issues#return-value

can we call this `terminationReason: ExitReason`

ArielDemarco: sure

platform/swift/source/Logger.swift
```diff
@@ -262,10 +268,10 @@ public final class Logger {
                 let kscrashReportDir = Logger.kscrashReportDirectory(base: directoryURL)
                 do {
                     try BitdriftKSCrashWrapper.configure(withCrashReportDirectory: kscrashReportDir)
-                    Logger.hasFatallyTerminatedOnPreviousRun = BitdriftKSCrashWrapper.didCrashLastLaunch()?.boolValue
+                    didCrashLastLaunch = BitdriftKSCrashWrapper.didCrashLastLaunch()?.boolValue == true
                     try BitdriftKSCrashWrapper.startCrashReporter()
                 } catch {
-                    // hasFatallyTerminatedOnPreviousRun may already be set if configure() succeeded
+                    // The previous-run result has already been computed without crash evidence.
```
snowp: This bit seems confusing - how does hitting this catch branch imply this? Or is it just saying that we'll end up computing the value without looking at kscrash?

platform/swift/source/reports/previousRunInfo/PreviousRunResolver.swift
```diff
@@ -0,0 +1,68 @@
...
+        }
+
+        // A clean exit marker means the previous run terminated gracefully (received willTerminate).
+        if previousState.wasCleanExit {
+            return PreviousRunInfo(status: .cleanExit)
+        }
+
+        // We consider that a change in the binary UUID or a change in the app version is an update
+        if previousState.appVersion != currentState.appVersion ||
+            previousState.binaryUUID != currentState.binaryUUID {
```
snowp: This makes sense but surprised it's necessary to check both, maybe document why if you know?

platform/swift/source/reports/previousRunInfo/PreviousRunInfoController.swift
```diff
@@ -0,0 +1,65 @@
...
+    private let currentState: PreviousRunCurrentState
+    private let previousState: PreviousRunStoredState?
+    private let terminationObserver: PreviousRunTerminationObserver
+    private let resolver = PreviousRunResolver()
+
+    private let previousRunInfoStorage = Atomic<PreviousRunInfo?>(nil)
+    var previousRunInfo: PreviousRunInfo { self.previousRunInfoStorage.load() ?? .unknown }
+
+    init?(baseDirectory: URL, appVersion: String, osVersion: String) {
+        let storeDirectory = baseDirectory.appendingPathComponent("reports/previous_run", isDirectory: true)
```
snowp: I would probably give it a different location than the reports dir so that there's no confusion over file directory ownership. Not sure if I added it but I wouldn't be surprised if the crash monitor clears out non-report files from this directory, if so seems better to avoid contention

 You can put this anywhere in the base directory, whatever makes sense

platform/swift/source/reports/PreviousRunInfoRepository.m
snowp, line 1: For my understanding why is this not in Swift?

platform/swift/source/reports/PreviousRunInfoRepository.m
```diff
@@ -0,0 +1,326 @@
...
+    uint8_t is_terminating;
+    uint8_t reserved[3];
+    uint64_t boot_time;
+    char app_version[BDPreviousRunInfoStringCapacity];
+    char os_version[BDPreviousRunInfoStringCapacity];
+    char binary_uuid[BDPreviousRunInfoUUIDCapacity];
+    uint8_t is_initialized;
+    uint8_t was_debugger_attached;
+    uint8_t trailing_reserved[6];
+} BDPreviousRunInfoRecord;
```
snowp, line 37: does this need a crc checksum or similar to ensure file integrity?

platform/swift/source/reports/PreviousRunInfoRepository.m
```diff
@@ -0,0 +1,326 @@
...
+    }
+
+    size_t bytes_to_copy = MIN((size_t)data.length, capacity - 1);
+    memcpy(destination, data.bytes, bytes_to_copy);
+}
+
+// Downgrades from NSURLFileProtectionComplete (the default) to CompleteUnlessOpen so the file stays
+// writable while the device is locked. Without this, markTerminating() would fail to write to the
+// mapped record if any write (like willTerminate) happens the screen locked.
+static BOOL bdpri_disable_file_protection(NSString *path) {
```
snowp, line 67: For my understanding is there some global file lock in play when the screen is locked?  Is this something we can just do or does this need extra privileges? I know we did something to the directory in general to have it bypass some of these checks

platform/swift/source/reports/PreviousRunInfoRepository.m
```diff
@@ -0,0 +1,326 @@
...
+        return;
+    }
+
+    size_t bytes_to_copy = MIN((size_t)data.length, capacity - 1);
+    memcpy(destination, data.bytes, bytes_to_copy);
+}
+
+// Downgrades from NSURLFileProtectionComplete (the default) to CompleteUnlessOpen so the file stays
+// writable while the device is locked. Without this, markTerminating() would fail to write to the
+// mapped record if any write (like willTerminate) happens the screen locked.
```
snowp:
```suggestion
// mapped record if any write (like willTerminate) happens while the screen is locked.
```

is it supposed to be this?

platform/swift/source/reports/previousRunInfo/PreviousRunResolver.swift
ArielDemarco: I can document it, for sure; but the main reason is when distributing internal adhoc builds (or even while coding and doing changes + build + run) the `appVersion` will probably remain the same; however, the UUID of the main binary will change

platform/swift/source/reports/previousRunInfo/PreviousRunInfoController.swift
ArielDemarco: Will do!

platform/swift/source/reports/PreviousRunInfoRepository.m
ArielDemarco: It's way much easier to deal with C bridging in objc, and considering that this will be the one communicating with rust via ffi, I decided to go with objc

platform/swift/source/reports/PreviousRunInfoRepository.m
ArielDemarco: I think it's a great idea; i'll do it (I also think we do it on the rust layer, so makes sense that this matches that way of working)

platform/swift/source/reports/PreviousRunInfoRepository.m
ArielDemarco: Yes; applying suggestion!

platform/swift/source/Logger.swift
ArielDemarco: Oh, this is outdated (I had all the logic to get the previousRunInfo inside the `do`/`try` block; i'll change it

platform/swift/source/reports/PreviousRunInfoRepository.m
ArielDemarco: It's not a global lock or something like that. It's "iOS Data Protection", that is a per-file encryption where the key's availability depends on the device state. `NSURLFileProtectionComplete` destroys the decryption key when the screen locks, so anybody with access to it would loose it (and in the case of trying to access the memory of an mmaped file, it'll generate a SIGBUS). `NSURLFileProtectionCompleteUnlessOpen` keeps the key alive as long as the file is open

platform/swift/source/Logger.swift
ArielDemarco: Changed!

platform/swift/source/reports/previousRunInfo/PreviousRunResolver.swift
ArielDemarco: Added some docs

platform/swift/source/reports/previousRunInfo/PreviousRunInfo.swift
ArielDemarco: Done

platform/swift/source/reports/previousRunInfo/PreviousRunInfoController.swift
ArielDemarco: Done

platform/swift/source/reports/previousRunInfo/PreviousRunInfo.swift
```diff
@@ -0,0 +1,35 @@
...
+    /// Compatibility shim for the legacy experimental API.
+    ///
+    /// This only returns `true` when the in-process crash reporter captured a fatal crash.
+    /// Other non-clean outcomes, such as `.unknown`, intentionally return `false`.
+    public var hasFatallyTerminated: Bool {
+        return self.terminationReason == .fatalCrash
+    }
+
+    /// Whether the previous run definitely terminated cleanly.
+    public var wasCleanExit: Bool {
```
murki: similarly should we rename Exit to Termination to be consistent

ArielDemarco: agreed

ArielDemarco: done

platform/swift/source/reports/PreviousRunInfoRepository.m
ArielDemarco: Done! Good catch!

platform/swift/source/reports/previousRunInfo/PreviousRunResolver.swift
```diff
@@ -0,0 +1,69 @@
...
+            return PreviousRunInfo(terminationReason: .osUpdate)
+        }
+
+        if previousState.bootTime != 0,
+           currentState.bootTime != 0,
+           Self.bootTimeDelta(previousState.bootTime, currentState.bootTime) > Self.bootTimeToleranceMicroseconds {
+            // The device rebooted between runs while the previous run ended without a clean-exit
+            // marker or captured crash. This currently collapses to `.unknown`, same as the
+            // fallthrough below.
+            // TODO: introduce a dedicated status (e.g. forced reboot / power-off) if we want to
```
FranAguilera: nit: can you add a linear ticket for easier future reference?

platform/swift/source/Logger.swift
```diff
@@ -262,10 +268,11 @@ public final class Logger {
                 let kscrashReportDir = Logger.kscrashReportDirectory(base: directoryURL)
                 do {
                     try BitdriftKSCrashWrapper.configure(withCrashReportDirectory: kscrashReportDir)
-                    Logger.hasFatallyTerminatedOnPreviousRun = BitdriftKSCrashWrapper.didCrashLastLaunch()?.boolValue
+                    didCrashLastLaunch = BitdriftKSCrashWrapper.didCrashLastLaunch()?.boolValue == true
                     try BitdriftKSCrashWrapper.startCrashReporter()
                 } catch {
-                    // hasFatallyTerminatedOnPreviousRun may already be set if configure() succeeded
+                    // KSCrash failed to initialize; previousRunInfo will be resolved without crash
```
FranAguilera: nit: should we log some internal error here?

FranAguilera (commented):
overall lgtm but will defer to iOS folks for final stamp. The only thing to consolidate after this will be the RN api

platform/swift/source/reports/previousRunInfo/PreviousRunResolver.swift
kattrali: Why not just check uuid? is app version adding anything?

platform/swift/source/reports/previousRunInfo/PreviousRunResolver.swift
kattrali: maybe just add the state now, it was actually my first question when I saw the enum without (`.reboot` or similar)

platform/swift/source/reports/previousRunInfo/PreviousRunResolver.swift
```diff
@@ -0,0 +1,69 @@
...
+            // distinguish a reboot-induced termination from a genuinely unknown one.
+            return .unknown
+        }
+
+        // Checked last since it's the least specific signal we have: only reported in debug builds,
+        // for a launch that otherwise looked unclean but wasn't explained by a crash, reboot, or
+        // app/OS update.
+        if previousState.wasDebuggerAttached {
+            return PreviousRunInfo(terminationReason: .debuggerAttached)
+        }
```
kattrali, line 59: more of a future thing (probably), but this case should go away in a world where we only derive did crash from out-of-process reporting. At least, I don't see why it wouldn't be reported if you (for example) disconnect a debugger then crash an app if not dependent on an in-process check which skips enabling crash reporting in the presence of a debugger.

platform/swift/source/reports/PreviousRunInfoRepository.m
```diff
@@ -0,0 +1,344 @@
...
+static NSString *bdpri_make_string(const char *buffer, size_t capacity) {
+    size_t length = strnlen(buffer, capacity);
+    NSData *data = [NSData dataWithBytes:buffer length:length];
+    NSString *value = [[NSString alloc] initWithData:data encoding:NSUTF8StringEncoding];
+    return value ?: @"";
+}
+
+static void bdpri_write_string(char *destination, size_t capacity, NSString *value) {
+    memset(destination, 0, capacity);
+    NSData *data = [value dataUsingEncoding:NSUTF8StringEncoding allowLossyConversion:YES];
```
kattrali:
```suggestion
    NSData *data = [value dataUsingEncoding:NSUTF8StringEncoding allowLossyConversion:NO];
```

Probably want to return an error type for this (maybe promote `BDPreviousRunInfoInvalidFileDescriptor` into an enum of error values?), but I don't see why we would want it to be lossy.

platform/swift/source/reports/previousRunInfo/PreviousRunResolver.swift
ArielDemarco: Will do!

platform/swift/source/reports/PreviousRunStateCaptureSupport.m
```diff
@@ -0,0 +1,88 @@
...
+
+        NSString *candidate = [NSString stringWithUTF8String:imageName];
+        if ([candidate isEqualToString:executablePath]) {
+            return header;
+        }
+    }
+
+    return fallbackHeader;
+}
+
```
kattrali: the tiniest nit but I think we know this isn't the last time we're gonna use "image for name" logic so might as well make it its own function separate from "image for main bundle"

platform/swift/source/reports/PreviousRunStateCaptureSupport.m
```diff
@@ -0,0 +1,88 @@
...
+    uintptr_t cursor = 0;
+    uint32_t commandCount = 0;
+    if (header->magic == MH_MAGIC_64 || header->magic == MH_CIGAM_64) {
+        const struct mach_header_64 *header64 = (const struct mach_header_64 *)header;
+        cursor = (uintptr_t)(header64 + 1);
+        commandCount = header64->ncmds;
+    } else {
+        cursor = (uintptr_t)(header + 1);
+        commandCount = header->ncmds;
```
kattrali: Will this run on anything which would exercise this branch?

test/platform/swift/unit_integration/core/default/PreviousRunInfoRepositoryTests.swift
```diff
@@ -0,0 +1,450 @@
...
+// https://polyformproject.org/wp-content/uploads/2020/06/PolyForm-Shield-1.0.0.txt
+
+@testable import CaptureLoggerBridge
+import Foundation
+import XCTest
+import zlib
+
+final class PreviousRunInfoRepositoryTests: XCTestCase {
+    private var directoryURL: URL!
+    private var sut: BDPreviousRunInfoRepository!
```
kattrali, line 15: ...sut?

kattrali (commented):
looks good to me, a few comments

test/platform/swift/unit_integration/core/default/PreviousRunInfoRepositoryTests.swift
ArielDemarco: [System/Subject Under Test](http://xunitpatterns.com/SUT.html)

platform/swift/source/reports/PreviousRunStateCaptureSupport.m
ArielDemarco: I did it without thinking about the context tbh; but yeah, I think i can remove this branch and have no problems at all considering we support iOS 15+

platform/swift/source/reports/PreviousRunStateCaptureSupport.m
ArielDemarco: Will do

platform/swift/source/reports/previousRunInfo/PreviousRunResolver.swift
ArielDemarco: it's a good call; i'll add documentation so we remember to remove this in the future

platform/swift/source/reports/PreviousRunInfoRepository.m
ArielDemarco: make sense; better safe than sorry hehehe

platform/swift/source/reports/previousRunInfo/PreviousRunResolver.swift
ArielDemarco: good point! I think thats much smarter than checking the app version. It also gets rid of a lot of unnecessary boilerplate

test/platform/swift/unit_integration/core/default/PreviousRunInfoRepositoryTests.swift
kattrali: Ah!! I was racking my brain trying to think of what that stood for. thanks!

platform/swift/source/reports/previousRunInfo/PreviousRunResolver.swift
ArielDemarco: Done

platform/swift/source/reports/previousRunInfo/PreviousRunResolver.swift
ArielDemarco: Done

platform/swift/source/reports/PreviousRunStateCaptureSupport.m
ArielDemarco: Done

platform/swift/source/reports/PreviousRunStateCaptureSupport.m
ArielDemarco: Removed

platform/swift/source/Logger.swift
ArielDemarco: The logger may not be fully initialized at this point so logging directly i dont think is safe here (maybe it would with `underlyingLogger`?). However, we're currently blind here, so I'll add a `degraded` case to `ReporterInitResolution` that gets set when KSCrash fails to start. This is going to be reported inside the existing `_fatal_issue_reporting_state` field in the SDK started event as rather than the misleading `.monitoring`/`"CRASH_REPORT_MONITORING"`. Seems the most convenient way to address this

kattrali (approved):

PR #1047, opened 2026-07-02
Preventing breaking change

ArielDemarco:
# Overview
[This PR](https://github.com/bitdriftlabs/capture-sdk/pull/1044/changes) introduces a breaking change. Even though the API is marked as experimental, avoiding the breaking change requires very little effort.

It's unlikely that any production code relies on the `PreviousRunInfo` constructor directly (perhaps only tests mocking those states), so this PR preserves backwards compatibility at virtually no cost.

Just in case, marked the API as deprecated:
image


murki (approved):

ArielDemarco:
We'll change it to `internal`; it makes no sense to have this publicly exposed.
Closing.

PR #1054, opened 2026-07-08
Add support for DD network tracing format

murki:
## Goal

Add Datadog trace propagation support.

## Summary

- add Datadog propagation support
- in `dd` mode, keep `_trace_id` Datadog-linkable even when requests already have W3C headers
- add runtime-configurable request ignore rules so Datadog uploader requests can be excluded (`client_config.network.request_ignore_match_paths` and `client_config.network.request_ignore_match_headers`)

## Verification

Use `gradle-test-app`.

Screenshot 2026-07-10 at 11 38 35

Checks:
- `client_config.trace.propagation_mode=dd`
- requests without existing trace headers log DD decimal `_trace_id`
- requests with existing sampled W3C headers still log DD-linkable `_trace_id`
- existing DD headers pass through unchanged
- ignored uploader requests do not show up as app spans

Example session:
- [session](https://timeline.bitdrift.dev/s/7678d855-4980-4f1e-914f-615ed50ac914?utm_source=sdk)

---

### DIFF
--- a/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/network/okhttp/CaptureOkHttpEventListenerFactory.kt
+++ b/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/network/okhttp/CaptureOkHttpEventListenerFactory.kt
@@ -8,9 +8,12 @@
 package io.bitdrift.capture.network.okhttp
 
 import io.bitdrift.capture.Capture
+import io.bitdrift.capture.CaptureRuntimeProvider
 import io.bitdrift.capture.ILogger
+import io.bitdrift.capture.IRuntimeProvider
 import io.bitdrift.capture.common.DefaultClock
 import io.bitdrift.capture.common.IClock
+import io.bitdrift.capture.common.RuntimeStringConfig
 import okhttp3.Call
 import okhttp3.EventListener
 
@@ -32,9 +35,17 @@ class CaptureOkHttpEventListenerFactory internal constructor(
     private val targetEventListenerFactory: EventListener.Factory?,
     private val logger: ILogger?,
     private val clock: IClock,
+    private val runtimeProvider: IRuntimeProvider,
     private val requestFieldProvider: OkHttpRequestFieldProvider,
     private val responseFieldProvider: OkHttpResponseFieldProvider,
 ) : EventListener.Factory {
+    private val configuredPropagationMode by lazy {
+        TracePropagationMode.fromRuntimeValue(
+            runtimeProvider.getRuntimeStringConfigValue(RuntimeStringConfig.TRACE_PROPAGATION_MODE),
+        )
+    }
+    private val requestIgnorePolicy by lazy { RuntimeOkHttpRequestIgnorePolicy(runtimeProvider) }
+
     /**
      * Initializes a new instance of the Capture event listener with an existing event listener factory.
      *
@@ -51,6 +62,7 @@ class CaptureOkHttpEventListenerFactory internal constructor(
         targetEventListenerFactory = targetEventListenerFactory,
         logger = Capture.logger(),
         clock = DefaultClock.getInstance(),
+        runtimeProvider = CaptureRuntimeProvider,
         requestFieldProvider = requestFieldProvider,
         responseFieldProvider = responseFieldProvider,
     )
@@ -58,13 +70,14 @@ class CaptureOkHttpEventListenerFactory internal constructor(
     override fun create(call: Call): EventListener {
         val currentLogger = getLogger()
         val targetEventListener = targetEventListenerFactory?.create(call)
-        if (currentLogger == null) {
+        if (currentLogger == null || requestIgnorePolicy.shouldIgnore(call.request())) {
             return targetEventListener ?: EventListener.NONE
         }
         return CaptureOkHttpEventListener(
             logger = currentLogger,
             clock = clock,
             targetEventListener = targetEventListener,
+            configuredPropagationMode = configuredPropagationMode,
             requestExtraFieldsProvider = requestFieldProvider,
             responseExtraFieldsProvider = responseFieldProvider,
         )
--- a/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/network/okhttp/CaptureOkHttpRequestIgnorePolicy.kt
+++ b/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/network/okhttp/CaptureOkHttpRequestIgnorePolicy.kt
@@ -1,0 +1,52 @@
+// capture-sdk - bitdrift's client SDK
+// Copyright Bitdrift, Inc. All rights reserved.
+//
+// Use of this source code is governed by a source available license that can be found in the
+// LICENSE file or at:
+// https://polyformproject.org/wp-content/uploads/2020/06/PolyForm-Shield-1.0.0.txt
+
+package io.bitdrift.capture.network.okhttp
+
+import io.bitdrift.capture.CaptureRuntimeProvider
+import io.bitdrift.capture.IRuntimeProvider
+import io.bitdrift.capture.common.RuntimeStringConfig
+import okhttp3.Request
+
+internal interface CaptureOkHttpRequestIgnorePolicy {
+    fun shouldIgnore(request: Request): Boolean
+}
+
+internal class RuntimeOkHttpRequestIgnorePolicy(
+    private val runtimeProvider: IRuntimeProvider = CaptureRuntimeProvider,
+) : CaptureOkHttpRequestIgnorePolicy {
+    private val ignoredPaths by lazy { readRuntimeCsvEntries(RuntimeStringConfig.NETWORK_REQUEST_IGNORE_PATHS_CSV) }
+
+    private val requiredHeaders by lazy {
+        readRuntimeCsvEntries(RuntimeStringConfig.NETWORK_REQUEST_IGNORE_REQUIRED_HEADERS_CSV)
+    }
+
+    override fun shouldIgnore(request: Request): Boolean {
+        val hasIgnoredPaths = ignoredPaths.isNotEmpty()
+        val hasRequiredHeaders = requiredHeaders.isNotEmpty()
+        if (!hasIgnoredPaths && !hasRequiredHeaders) {
+            return false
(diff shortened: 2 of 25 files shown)


platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/network/okhttp/CaptureOkHttpTracingInterceptor.kt
```diff
@@ -74,6 +74,13 @@ class CaptureOkHttpTracingInterceptor
                     requestBuilder.header("X-B3-Sampled", "1")
                 }
 
+                TracePropagationMode.DD -> {
+                    requestBuilder.header("x-datadog-trace-id", traceContext.traceIdAsDecimal())
+                    requestBuilder.header("x-datadog-parent-id", traceContext.spanIdAsDecimal())
```
murki: blocker: is this header required?

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/network/okhttp/CaptureOkHttpTracingInterceptor.kt
```diff
@@ -74,6 +74,13 @@ class CaptureOkHttpTracingInterceptor
...
+                    // Using 2 to mean USER_KEEP from https://datadoghq.dev/dd-trace-rb/Datadog/Tracing/Sampling/Ext/Priority.html
+                    requestBuilder.header("x-datadog-sampling-priority", "2")
```
murki, line 81: blocker: is this the right sampling priority to use?

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/network/okhttp/TracePropagation.kt
```diff
@@ -41,6 +43,20 @@ internal object TracePropagation {
...
         }
 
+        val ddTraceId = request.header("x-datadog-trace-id")
+        if (ddTraceId != null) {
+            val sampled = request.header("x-datadog-sampling-priority")
+            // Datadog sampling priority values: 1 = AUTO_KEEP, 2 = USER_KEEP
+            // https://datadoghq.dev/dd-trace-rb/Datadog/Tracing/Sampling/Ext/Priority.html
+            if (sampled == "1" || sampled == "2") {
+                return try {
+                    BigInteger(ddTraceId).toString(16).padStart(16, '0')
```
FranAguilera: i'm not sure we should convert this to hex. Will check their docs

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/network/okhttp/CaptureOkHttpTracingInterceptor.kt
murki: @FranAguilera after our testing, the "x-datadog-parent-id" header will not be needed so we should remove it from this PR

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/network/okhttp/CaptureOkHttpTracingInterceptor.kt
FranAguilera: I think so, this is similar to what we do with the other formats

murki: Oh yeah we tested this was right, cane be left as 2

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/network/okhttp/CaptureOkHttpTracingInterceptor.kt
```diff
@@ -74,6 +75,12 @@ class CaptureOkHttpTracingInterceptor
                     requestBuilder.header("X-B3-Sampled", "1")
                 }
 
+                TracePropagationMode.DD -> {
+                    requestBuilder.header("x-datadog-trace-id", traceContext.traceIdAsDecimal())
+                    // Using 2 to mean USER_KEEP from https://datadoghq.dev/dd-trace-rb/Datadog/Tracing/Sampling/Ext/Priority.html
+                    requestBuilder.header("x-datadog-sampling-priority", "2")
+                }
```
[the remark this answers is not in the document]
FranAguilera: This is intended. We tested and parent id is not needed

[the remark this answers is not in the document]
FranAguilera: https://github.com/bitdriftlabs/capture-sdk/pull/1054#discussion_r3554592536

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/network/okhttp/CaptureOkHttpRequestIgnorePolicy.kt
```diff
@@ -0,0 +1,52 @@
...
+
+        val matchesIgnoredPath = request.url.encodedPath in ignoredPaths
+        val matchesRequiredHeader = requiredHeaders.any { request.header(it) != null }
+
+        return when {
+            hasIgnoredPaths && hasRequiredHeaders -> matchesIgnoredPath && matchesRequiredHeader
+            hasIgnoredPaths -> matchesIgnoredPath
+            else -> matchesRequiredHeader
+        }
+    }
```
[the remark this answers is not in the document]
FranAguilera: This is ok given the runtime config is only loaded in cold launch and it won't be changed during the ongoing session

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/network/okhttp/CaptureOkHttpEventListenerFactory.kt
```diff
@@ -32,9 +35,17 @@ class CaptureOkHttpEventListenerFactory internal constructor(
...
     private val clock: IClock,
+    private val runtimeProvider: IRuntimeProvider,
     private val requestFieldProvider: OkHttpRequestFieldProvider,
     private val responseFieldProvider: OkHttpResponseFieldProvider,
 ) : EventListener.Factory {
+    private val configuredPropagationMode by lazy {
+        TracePropagationMode.fromRuntimeValue(
+            runtimeProvider.getRuntimeStringConfigValue(RuntimeStringConfig.TRACE_PROPAGATION_MODE),
+        )
+    }
```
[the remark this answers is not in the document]
FranAguilera: Same as above

platform/swift/source/integrations/url_session/extensions/URLSessionTask+Swizzling.swift
```diff
@@ -34,15 +34,24 @@ extension URLSessionTask {
             return
         }
 
+        let ignorePolicy = RuntimeURLSessionIgnorePolicy(
+            ignorePathsCSV: integration.requestIgnorePathsCSV,
+            requiredHeadersCSV: integration.requestIgnoreRequiredHeadersCSV,
+            )
```
kattrali: This is maybe not the place to rework runtime value propagation entirely, but it seems like unnecessary overhead to have to fetch and resplit this value on every request, rather than having it push an update to a swift-side cache when the value is updated and have it ready as an array to validate against rather than text.

platform/swift/source/integrations/url_session/TracePropagation.swift
```diff
@@ -97,14 +117,89 @@ enum URLSessionTracePropagation {
...
+        switch (hasIgnoredPaths, hasRequiredHeaders) {
+        case (true, true):
+            return matchesIgnoredPath && matchesRequiredHeader
+        case (true, false):
+            return matchesIgnoredPath
+        case (false, true):
+            return matchesRequiredHeader
+        case (false, false):
+            return false
+        }
```
kattrali: nit: exactly the same but slightly more readable to not have to refer to what the two case values represent

```suggestion
        if hasIgnoredPaths && hasRequiredHeaders {
            return matchesIgnoredPath && matchesRequiredHeader
        } else if hasIgnoredPaths {
            return matchesIgnoredPath
        }
        return matchesRequiredHeader
```

platform/swift/source/integrations/url_session/URLSessionIntegration.swift
```diff
@@ -79,6 +79,22 @@ final class URLSessionIntegration {
...
+    }
+
+    var requestIgnoreRequiredHeadersCSV: String {
+        guard let logger = Logger.getShared() as? Logger else {
+            return RuntimeVariable<String>.networkRequestIgnoreRequiredHeadersCSV.defaultValue
+        }
+
+        return logger.runtimeValue(.networkRequestIgnoreRequiredHeadersCSV)
+    }
+
```
kattrali: technically, a comma could appear in a [URL path](https://www.rfc-editor.org/info/rfc3986/#section-3.3) or a [header](https://www.rfc-editor.org/info/rfc9110/#section-5.5). Its no big deal if we're using a CSV lib or parsing escaped values, but it might bite us sometime here

platform/swift/source/integrations/url_session/extensions/URLSessionTask+Swizzling.swift
FranAguilera: good idea will cache

platform/swift/source/integrations/url_session/URLSessionIntegration.swift
FranAguilera: good call, will move away from csv

FranAguilera: after thinking carefully a bit more not sure if worth the trouble now given the extra complexity (e.g. having an json array with values could be a solution). Maybe in the future we can update runtime better and formalize a list of items/Array of values instead of raw string

kattrali: Makes sense. JSON array would be a good solution at some point.

kattrali (approved):
iOS side lgtm

PR #1055, opened 2026-07-09
[Android] [release] Consolidate maven central artifacts into a single bundle

FranAguilera:
Resolves BIT-8921

## Verification

Verified locally:
- One bundle is created dist/maven-central/android-0.23.9.maven-central.zip
- The zip contains all expected root artifaccts
- Artifacts are signed

Screenshot 2026-07-09 at 15 54 14

Screenshot 2026-07-09 at 15 55 09

---


murki (approved):
nice

PR #1056, opened 2026-07-10
[Android] Optimize Capture.Logger.start call

FranAguilera:
## Problem

Capture.Logger.start() was doing unnecessary eager startup work on the caller thread.

## Solution

- Creating lazily `OkHttpCaptureApiClient` (before named OkHttpClient) which is only called when deviceCode is requested or there is an internal error to report. The internal Gson contructor is also lazy until first `OkHttpCaptureApiClient` usage.
- Creating whole session URL as potentially can never be called.

Additionally renamed some classes to clarify their intention:
- `OkHttpApiClient` <> `OkHttpCaptureApiClient`. Given this is internal to capture and used API backend communication (for now device code generation and internal error reporting)
- `OkHttpNetwork` to `OkHttpCaptureStream`. Also internal to capture and responsible to keep the stream connection with the backend

### Local benchmark improvement with a Fairphone 6

| Metric | Before | After | Diff |
|--------| -------| ------| -------|
| duration |  1,880,327 ns |  1,312,276 ns    | -30.2%    | 
| allocs |  617 allocs   |   530 allocs     |  -14.1%  |

## Verification

- Verified okhttp calls via gradle-test-app (device code/artificial errors/etc)
- Added a new benchmark test to compare internal of Capture.Logger.start via LoggerImpl instance creation

---

### DIFF
--- a/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/LoggerImpl.kt
+++ b/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/LoggerImpl.kt
@@ -39,8 +39,8 @@ import io.bitdrift.capture.events.span.Span
 import io.bitdrift.capture.experimental.ExperimentalBitdriftApi
 import io.bitdrift.capture.network.HttpRequestInfo
 import io.bitdrift.capture.network.HttpResponseInfo
-import io.bitdrift.capture.network.okhttp.OkHttpApiClient
-import io.bitdrift.capture.network.okhttp.OkHttpNetwork
+import io.bitdrift.capture.network.okhttp.OkHttpCaptureApiClient
+import io.bitdrift.capture.network.okhttp.OkHttpCaptureStream
 import io.bitdrift.capture.network.okhttp.buildSharedOkHttpClient
 import io.bitdrift.capture.providers.ArrayFields
 import io.bitdrift.capture.providers.DateProvider
@@ -94,7 +94,12 @@ internal class LoggerImpl(
         ),
     preferences: IPreferences = Preferences(context),
     sharedOkHttpClient: OkHttpClient = buildSharedOkHttpClient(),
-    private val apiClient: OkHttpApiClient = OkHttpApiClient(apiUrl, apiKey, client = sharedOkHttpClient),
+    // This client is only used for temporary device code creation and internal
+    // SDK error reporting, so defer constructing it until one of those paths runs.
+    private val apiClient: Lazy<OkHttpCaptureApiClient> =
+        lazy {
+            OkHttpCaptureApiClient(apiUrl, apiKey, client = sharedOkHttpClient)
+        },
     private var deviceCodeService: DeviceCodeService = DeviceCodeService(apiClient),
     activityManager: ActivityManager = context.getSystemService(Context.ACTIVITY_SERVICE) as ActivityManager,
     bridge: IBridge = CaptureJniLibrary,
@@ -116,9 +121,17 @@ internal class LoggerImpl(
     private val runtime: JniRuntime
     private var jankStatsMonitor: JankStatsMonitor? = null
 
-    // we can assume a properly formatted api url is being used, so we can follow the same pattern
-    // making sure we only replace the first occurrence
-    private val sessionUrlBase: HttpUrl
+    // Session URLs are only needed when queried externally, so derive the
+    // timeline base URL on first access. We replace only the first "api."
+    // prefix when converting the Capture API host to the timeline host.
+    private val sessionUrlBase: HttpUrl by lazy {
+        HttpUrl
+            .Builder()
+            .scheme("https")
+            .host(apiUrl.host.replaceFirst("api.", "timeline."))
+            .addQueryParameter("utm_source", "sdk")
+            .build()
+    }
 
     private val resourceUtilizationTarget: ResourceUtilizationTarget
     private val eventsListenerTarget = EventsListenerTarget()
@@ -148,14 +161,6 @@ internal class LoggerImpl(
     internal val loggerId: LoggerId
 
     init {
-        this.sessionUrlBase =
-            HttpUrl
-                .Builder()
-                .scheme("https")
-                .host(apiUrl.host.replaceFirst("api.", "timeline."))
-                .addQueryParameter("utm_source", "sdk")
-                .build()
-
         val networkAttributes = NetworkAttributes(context)
 
         metadataProvider =
@@ -173,7 +178,7 @@ internal class LoggerImpl(
             )
 
         val network =
-            OkHttpNetwork(
+            OkHttpCaptureStream(
                 apiBaseUrl = apiUrl,
                 okHttpClient = sharedOkHttpClient,
             )
--- a/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/events/lifecycle/AppExitLogger.kt
+++ b/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/events/lifecycle/AppExitLogger.kt
@@ -17,6 +17,7 @@ import io.bitdrift.capture.IInternalLogger
 import io.bitdrift.capture.LogAttributesOverrides
 import io.bitdrift.capture.LogLevel
 import io.bitdrift.capture.LogType
+import io.bitdrift.capture.common.IBackgroundThreadHandler
 import io.bitdrift.capture.common.Runtime
 import io.bitdrift.capture.common.RuntimeFeature
 import io.bitdrift.capture.events.performance.IMemoryMetricsProvider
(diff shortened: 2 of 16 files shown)


platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/events/lifecycle/AppExitLogger.kt
```diff
@@ -56,17 +56,23 @@ internal class AppExitLogger(
...
+    private val isFatalIssueReportingInitialized = IssueReporterState.Initialized == issueReporter?.initializationState()
+
     @SuppressLint("NewApi")
     fun installAppExitLogger() {
         if (!runtime.isEnabled(RuntimeFeature.APP_EXIT_EVENTS)) {
             return
         }
-        captureUncaughtExceptionHandler.install(this)
+        if (!isFatalIssueReportingInitialized) {
+            captureUncaughtExceptionHandler.install(this)
```
FranAguilera, line 67: A bit out of scope but this is not needed when fatal issue is initialized

platform/jvm/gradle-test-app/build.gradle.kts
```diff
@@ -3,10 +3,22 @@ plugins {
     alias(libs.plugins.apollo.graphql)
     alias(libs.plugins.kotlin.android)
     alias(libs.plugins.compose.compiler)
-    id("com.google.firebase.crashlytics") version "3.0.6"
+    id("com.google.firebase.crashlytics") version "3.0.6" apply false
```
FranAguilera, line 6: Without it, release builds are broken given i'm not adding the firebase json config to the root and intiallizing at runtime. Only using this for debug builds either way

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/events/lifecycle/AppExitLogger.kt
```diff
@@ -56,17 +59,26 @@ internal class AppExitLogger(
...
             return
         }
-        captureUncaughtExceptionHandler.install(this)
-        logPreviousExitReasonIfAny()
+        if (!isFatalIssueReportingInitialized) {
+            captureUncaughtExceptionHandler.install(this)
+        }
+        backgroundThreadHandler.runAsync {
+            // Ordering is handled at shared-core layer
+            logPreviousExitReasonIfAny()
```
murki: I see your comment here but I just would like to double click on the importance on making sure that this added concurrency doesn't change any of the current assumptions with regards of ordering, last know session id, etc cc @snowp to confirm

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/network/okhttp/OkHttpNetwork.kt
```diff
@@ -76,12 +76,12 @@ internal fun newDuplexRequestBody(contentType: MediaType): PipeDuplexRequestBody
...
     apiBaseUrl: HttpUrl,
     timeoutSeconds: Long = 2L * 60,
-    private val okHttpClient: OkHttpClient,
+    private val okHttpClient: Lazy<OkHttpClient>,
     private val networkDispatcher: CaptureDispatchers.Network = CaptureDispatchers.Network,
 ) : ICaptureNetwork {
-    private val client: OkHttpClient =
-        run {
-            val builder = okHttpClient.newBuilder()
+    private val client: Lazy<OkHttpClient> =
```
murki: I'm not sure what the benefit of lazying is here since it looks like we access this instance straight from `init`?

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/LoggerImpl.kt
```diff
@@ -93,8 +93,11 @@ internal class LoggerImpl(
             ProcessLifecycleOwner.get(),
         ),
     preferences: IPreferences = Preferences(context),
-    sharedOkHttpClient: OkHttpClient = buildSharedOkHttpClient(),
-    private val apiClient: OkHttpApiClient = OkHttpApiClient(apiUrl, apiKey, client = sharedOkHttpClient),
+    private val sharedOkHttpClient: Lazy<OkHttpClient> = lazy { buildSharedOkHttpClient() },
```
murki: similarly I'm having a hard time understanding the benefits of lazying here since we're immediately loading it via `OkHttpNetwork().init`. additionally if I'm following the code correctly now we have redundant chained lazying: sharedOkHttpClient -> apiClient -> ErrorReporterService / DeviceCodeService. In general I would rather be very surgical and only apply intentional lazies on things that we know a) are expensive to create and b) are not called / needed immediately, and then add a comment of why

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/network/okhttp/OkHttpNetwork.kt
FranAguilera: It should be still deferrred as `client.value` is not called until startStream() is called. Either way will double check again if this is overkill

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/LoggerImpl.kt
murki: I wonder if could test a simpler approach. From what I remember when looking at the ANR stacktraces it seemed that the `Gson()` instantiation inside `OkHttpApiClient` which is shared by  `ErrorReporterService` and `DeviceCodeService` indirectly lazy and more efficient

FranAguilera: will double check

murki (commented):
in general can we add a comment to each added `lazy` to justify its rationale. it's hard to follow what each one accomplishes

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/LoggerImpl.kt
```diff
@@ -118,7 +121,14 @@ internal class LoggerImpl(
 
     // we can assume a properly formatted api url is being used, so we can follow the same pattern
     // making sure we only replace the first occurrence
-    private val sessionUrlBase: HttpUrl
+    private val sessionUrlBase: HttpUrl by lazy {
```
murki, line 127: 👍

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/network/okhttp/OkHttpNetwork.kt
FranAguilera: Okays, removed lazy for OkHttpClient given the stream is always called right away (btw now renamed OkHttpNetwork to OkHttpCaptureStream)

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/LoggerImpl.kt
FranAguilera: Simplified this a bit, will be only creating lazily the `OkHttpApiClient` (now named `OkHttpCaptureApiClient`) given that can never be called if customer never calls device code or no internal errors are ever called.

Also regarding Gson instantiation now with the lazy init is also effectively lazy 

```
    private val apiClient: Lazy<OkHttpCaptureApiClient> =
        lazy {
            OkHttpCaptureApiClient(apiUrl, apiKey, client = sharedOkHttpClient)
        },
```

And that the value from the lazy instance is called immediately with perform

1. `DeviceCodeService`

```
    fun createTemporaryDeviceCode(
        deviceId: String,
        completion: (CaptureResult<String>) -> Unit,
    ) {
        val typedRequest = DeviceCodeRequest(deviceId)

        apiClient.value.perform<DeviceCodeRequest, DeviceCodeResponse>(
```
        
2. `ErrorReporterService`

```
    override fun reportError(
        message: String,
        details: String?,
        fields: Map<String, String>,
    ) {
        val typedRequest = ErrorReportRequest(message, details)

        val allFields =
            buildMap {
                putAll(headers())
                putAll(fields)
            }

        apiClient.value.perform<ErrorReportRequest, Unit>(      
```

murki (approved):
lgtm barring that question about writing the appexit log async

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/events/lifecycle/AppExitLogger.kt
snowp: I think this does break the contract that the 
```
                logger.logInternal(
                    LogType.LIFECYCLE,
                    lastExitInfo.reason.toLogLevel(),
                    buildAppExitFields(lastExitInfo),
                    attributesOverrides = LogAttributesOverrides.PreviousRunSessionId(timestampMs),
                ) { APP_EXIT_EVENT_NAME }
```
log will be emitted before custom logs are sent to the logger, which could result in session thrashing. I think ideally we'd have a way to queue up work that runs before the logger accepts custom logs but as for now it seems safest to keep this as is

FranAguilera: sounds good, will revert for now. As we have to do the pre init / background init will take a look as part of that

PR #1065, opened 2026-07-21
[BIT-7195] iOS WebView Instrumentation 

ArielDemarco:
# Overview
Provide support to webview instrumentation via Swizzling.


platform/swift/source/integrations/webview/messages/ConsoleMessage.swift
```diff
@@ -0,0 +1,37 @@
...
+    let level: String
+    let message: String
+    let args: [String]?
+
+    func makeLoggingAction(context: WebViewLoggingContext) -> WebViewLoggingAction? {
+        let logLevel: LogLevel = switch level {
+        case "error": .error
+        case "warn": .warning
+        case "info": .info
+        default: .debug
```
kattrali, line 25: no [trace](https://developer.mozilla.org/en-US/docs/Web/API/console/trace_static)?


```suggestion
        case "info": .info
        case "trace": .trace
        default: .debug
```

platform/swift/source/integrations/webview/messages/LongTaskMessage.swift
```diff
@@ -0,0 +1,48 @@
...
+    let type: WebViewMessageType
+    let timestamp: Int64
+    let durationMs: Double
+    let startTime: Double
+    let attribution: LongTaskAttribution?
+
+    func makeLoggingAction(context: WebViewLoggingContext) -> WebViewLoggingAction? {
+        let logLevel: LogLevel = switch durationMs {
+        case 200...: .warning
+        case 100...: .info
```
kattrali, line 22: do these duration thresholds live somewhere (on other platforms/docs)? I could see it being something which may need to be known (or even configurable)?

platform/swift/source/integrations/webview/messages/WebViewLoggableMessage.swift
```diff
@@ -0,0 +1,61 @@
...
+    let currentPageViewSpanID: String?
+    let activePageViewSpans: [String: Span]
+
+    func parentLoggerSpanID(for webViewSpanID: String?) -> UUID? {
+        if let webViewSpanID, let activeSpan = activePageViewSpans[webViewSpanID] {
+            return activeSpan.id
+        }
+
+        if let webViewSpanID {
+            return UUID(uuidString: webViewSpanID)
```
kattrali, line 24: is falling back to nil the only thing we can do here if the ID is not a valid UUID? (maybe it is?)

platform/swift/source/integrations/webview/WebViewIntegration.swift
```diff
@@ -0,0 +1,194 @@
...
+    ) {
+        underlyingLogger.update { storedLogger in
+            storedLogger = logger
+        }
+
+        guard !hasBeenSwizzled.load() else {
+            return
+        }
+
+        hasBeenSwizzled.update { swizzled in
```
kattrali, line 42: Is it possible to end up here twice if two threads encounter this block before the exchange completes? maybe using a static block would make sense here?

platform/webview/scripts/generate-native.ts
```diff
@@ -1,37 +1,43 @@
-import * as fs from 'node:fs';
-import * as path from 'node:path';
-import { fileURLToPath } from 'node:url';
+import * as fs from "node:fs";
+import * as path from "node:path";
+import { fileURLToPath } from "node:url";
```
kattrali: probably should have a defined style for these going forward - for example, single quotes match what we're doing [elsewhere](https://github.com/bitdriftlabs/bitdrift-frontend/blob/f4bc3acb6771d40abb4d6583ad39b712abc94d2a/.prettierrc.js#L2)

platform/swift/source/integrations/webview/WebViewIntegration.swift
```diff
@@ -0,0 +1,194 @@
...
+            injectionTime: .atDocumentStart,
+            forMainFrameOnly: false
+        )
+
+        let userContentController = configuration.userContentController
+        userContentController.addUserScript(script)
+        userContentController.add(ScriptMessageHandler(loggingProvider: WebViewLoggingProvider()), name: "BitdriftLogger")
+        let webView = cap_init(frame: frame, configuration: configuration)
+        if #available(iOS 16.4, *) {
+            webView.isInspectable = true
```
kattrali, line 80: is this necessary? maybe it should be a debug option somehow?

platform/swift/source/integrations/webview/messages/ConsoleMessage.swift
ArielDemarco: Yeah, tbh, i just copied android behavior, but makes total sense to apply this

platform/swift/source/integrations/webview/messages/LongTaskMessage.swift
ArielDemarco: Similar to the ones before, i basically copied android behavior. Im open to make this configurable, or at least add the info on the docs. Wdyt @FranAguilera ?

platform/swift/source/integrations/webview/messages/WebViewLoggableMessage.swift
ArielDemarco: Again I basically copied the graceful degradation to match Android's behavior which also silently drops on not having a valid UUID. Seems reasonable to me

platform/swift/source/integrations/webview/WebViewIntegration.swift
ArielDemarco: sure, good catch; will fix it

platform/webview/scripts/generate-native.ts
ArielDemarco: it _should_ be single quotes based on `biome.json`; i'll re-run it and see how it goes. Good catch

platform/swift/source/integrations/webview/WebViewIntegration.swift
ArielDemarco: It's a leftover from debugging; good call. Wrapping it in `#if DEBUG` so it doesnt run in release builds and we can still inspect when testing.

platform/swift/source/integrations/webview/WebViewIntegration.swift
```diff
@@ -0,0 +1,195 @@
...
+}
+
+struct WebViewLoggingProvider: LoggingProvider {
+    func getLogging() -> (any Logging)? {
+        WebViewIntegration.shared.getLogger()
+    }
+}
+
+final class WebViewIntegration {
+    private var hasBeenSwizzled = Atomic(false)
```
kattrali, line 29: microscopic nit:


```suggestion
    private var hasBeenConfigured = Atomic(false)
```

since this value will be set to true even if swizzling is disabled

ArielDemarco: Changing in the [last PR](https://github.com/bitdriftlabs/capture-sdk/pull/1069)

ArielDemarco: Done there!

kattrali (approved):

PR #1068, opened 2026-07-21
[BIT-9194] iOS - Implement manual WKWebView Instrumentation

ArielDemarco:
# Overview
Associated PRs include automatic instrumentation via swizzling. However, some may prefer not to rely on automatic instrumentation because of the risks associated with swizzling. That's why we enable the possibility to do manual `WKWebView` instrumentation


jacksonhardaker (approved):
Makes sense to me, but I'm not an expert

FranAguilera (approved):

PR #1060, opened 2026-07-16
[Android] Add optional executor to start call

FranAguilera:
## Proposal

[Past investigation doc](https://docs.google.com/document/d/1Ir3rpBaJeA3ZX6dY7AKUd8_RdmzXVM4pfnSUU46dmp0/edit?tab=t.0#heading=h.a2phuhp8qtm6)

Allowing to provide an executor to `Capture.Logger.start` where the internal initialization logic will take place.

- Prevent blocking main thread for callers on Application.onCreate etc
- Prevents dropping logs for consumers that wraps the start call in their background thread
- The in memory logs keep the original timestamp emission via `LogAttributesOverrides.OccurredAt`
- Limit the inMemory queue size and report internal log if that threshold is ever reached.

## Verification

1. With an added artificial delay at LoggerImpl constructor

- [Session](https://timeline.bitdrift.dev/session/b2349cc1-e7cf-4f57-956d-8d9e3a153beb?utm_source=sdk&utilization=0&expanded=6299595603258992535) with sdk start wrapper in background thread, you can see logs are not being collected while sdk is being initialized
- [Session](https://timeline.bitdrift.dev/session/1824948b-9c21-436e-91be-7c63f9d347bb?utm_source=sdk&utilization=0) with executor passed in to start method 

2. Microbenchmark tests (without artificial delay)

| start on main thread | start on passed executor |
| ---------------------| -------------------------|
| Screenshot 2026-07-16 at 11 11 51 | Screenshot 2026-07-16 at 11 11 40 |

3. Also did full e2e manual verification.

4. Pass an shutdown executor. startResult callback should emit detailed exception
---

- [ ] `CHANGELOG.md`'s "Unreleased" section has been updated, if applicable.

PR #1077, opened 2026-08-04
[Android] Capture.Logger.preInit

FranAguilera:
A slight variation of https://github.com/bitdriftlabs/capture-sdk/pull/1060

Here we expose a new public API Capture.Logger.preInit, that can be called before initial Capture.Logger.start call so any logs are not missed during initialization of start (in case this gets called off main thread)


FranAguilera:
mainly used a prototyping sharing, will reopen when ready

PR #1080, opened 2026-08-06
Implementation of the new swift based MetricKit APIs

ArielDemarco:
# Overview
The idea of this PR is to use the new MetricKit API that's going to be released in iOS 27. This is not only a replacement so crashes are handled in the new swift based API, it also includes handling memory exceptions (OOMs) and the possibility to use all the new features (as the old Objc API is marked as deprecated in Xcode 27).

## Details
- As we'll keep using Xcode 26 to build this, everything usage of APIs from iOS 27 is gated behind the `if compiler(>=6.4)`
- A refactor was made to simplify `DiagnosticEventReporter`, and move the logic to two separate files so they can be shared between the new handler (`MetricKitDiagnosticManager.swift`) and the old handler. One of those files is the one that handles all the complexity to write the report to disk (`DiagnosticReportWriter`) and the other one is a helper
- The new handler adapts to the old data shape instead of the other way around. `MetricKitLegacyCrashAdapter` converts the new iOS 27 `CrashDiagnostic`/`MemoryExceptionDiagnostic` structs into the same dictionary shape the old `MXDiagnostic` API produced, so it can reuse the writer and the existing enrichment.


platform/swift/source/reports/MetricKitDiagnosticManager.swift
```diff
@@ -0,0 +1,255 @@
...
+            let capturedCrash = self.crashReporting.cachedPreviousCrash()
+            self.processCrash(
+                diagnostic,
+                environment: report.environment,
+                timestamp: timestamp,
+                capturedCrash: capturedCrash
+            )
+        case .memoryException(let diagnostic):
+            self.processMemoryException(
+                diagnostic, environment: report.environment, timestamp: timestamp)
```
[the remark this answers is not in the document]
ArielDemarco: Tests are going to be done on next PR of the stack.

platform/swift/source/reports/MetricKitDiagnosticParsing.m
```diff
@@ -0,0 +1,205 @@
...
+  *seconds = (uint64_t)whole_seconds;
+  *nanos = (uint32_t)((timestamp - whole_seconds) * NSEC_PER_SEC);
+}
+
+- (int8_t)architectureConstantFor:(NSString *)platformArchitecture {
+  if (!platformArchitecture) {
+    return /* Architecture.Unknown */ 0;
+  }
+  return [platformArchitecture containsString:@"arm64"] ? /* Architecture.arm64 */ 2 : /* Architecture.x86_64 */ 4;
+}
```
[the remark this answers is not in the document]
ArielDemarco: Just moved this exact logic from one place to another; is not really affecting today.

platform/swift/source/reports/MetricKitDiagnosticManager.swift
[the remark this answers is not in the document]
ArielDemarco: Not going to be included in the changelog as we're not going to release it (we'll compile xcframework with xcode 26, so the future is not going to be compiled)

platform/swift/source/reports/CrashReporterService.swift
```diff
@@ -138,6 +150,7 @@ private struct RuntimeFileState {
         if isBitdriftCrashHandlerEnabled {
             self.bitdriftCrashHandler.stopCrashReporter()
         }
+        self.metricKitDiagnosticManager?.stop()
```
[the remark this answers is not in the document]
ArielDemarco: Done!

platform/swift/source/reports/CrashReporterService.swift
```diff
@@ -102,7 +103,18 @@ private struct RuntimeFileState {
             }
 
             lastRunResult = didCrashLastLaunch()
+
+            #if compiler(>=6.4)
+            if #available(iOS 27.0, *) {
+                let manager = makeMetricKitDiagnosticManager(sdkBaseURL: sdkBaseURL, underlyingLogger: underlyingLogger)
+                manager.start()
```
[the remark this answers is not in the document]
ArielDemarco: Done!

platform/swift/source/reports/CrashReporterService.swift
```diff
@@ -296,3 +317,48 @@ private extension CrashReporterService {
...
+
+#if compiler(>=6.4)
+extension CrashReporterService {
+    @available(iOS 27.0, *)
+    func makeMetricKitDiagnosticManager(
+        sdkBaseURL: URL,
+        underlyingLogger: CoreLogging
+    ) -> MetricKitDiagnosticManager {
+        let useStackOverlapMatching = underlyingLogger.runtimeValue(.crashThreadMatchingByStackOverlap)
+        let fileSizeOptimizationEnabled = underlyingLogger.runtimeValue(.optimizeFatalIssueReportSize)
```
snowp, line 329: Just to double check, these are initialized at the right time to be visible here?

platform/swift/source/reports/MetricKitLegacyCrashAdapter.swift
```diff
@@ -0,0 +1,104 @@
...
+// Use of this source code is governed by a source available license that can be found in the
+// LICENSE file or at:
+// https://polyformproject.org/wp-content/uploads/2020/06/PolyForm-Shield-1.0.0.txt
+
+import Foundation
+import MetricKit
+
+#if compiler(>=6.4)
+
+@available(iOS 27.0, *)
```
snowp, line 13: So I understand why do we need both the ifdef and the available guard?

test/platform/swift/unit_integration/core/reports/CrashReporterServiceTests.swift
```diff
@@ -282,6 +291,13 @@ private extension CrashReporterServiceTests {
         )
     }
 
+    func supportsIOS27() -> Bool {
+        if #available(iOS 27.0, *) {
```
snowp, line 295: Not sure if you've set this up but should we be running tests on iOS 27 as well? With the change to build the release + the recent bazel cache I don't think it would be that expensive

platform/swift/source/reports/MetricKitLegacyCrashAdapter.swift
```diff
@@ -0,0 +1,104 @@
...
+                "\(environment.osVersion.platform) \(environment.osVersion.number) (\(environment.osVersion.buildNumber))",
+            "lowPowerModeEnabled": environment.lowPowerModeEnabled,
+        ]
+    }
+
+    private static func makeCallStackTreeDict(_ tree: CallStackTree) -> [String: Any] {
+        let threads = tree.callStackThreads.map { thread -> [String: Any] in
+            [
+                "threadAttributed": thread.threadAttributed ?? false,
+                "callStackRootFrames": [self.frameDict(thread.rootFrames.first, tree: tree)]
```
snowp, line 63: Something flagged by my agent: What does it mean for there to be multiple rootFrames? Is it just one frame here always or do we need to handle this?

snowp (approved):

platform/swift/source/reports/CrashReporterService.swift
ArielDemarco: Probably it'd be worth to check them out; I just moved them to a separate service, but the timing is the same.
But, I tested the new crash reporter and it had some problems to do it _before_ the underlying logger it's initialized. So maybe worth checking it out (mostly for the `fileSizeOptimization`, probably we can get rid of the thread matching one)

test/platform/swift/unit_integration/core/reports/CrashReporterServiceTests.swift
ArielDemarco: I created a new xcode 27 build step ([run example](https://github.com/bitdriftlabs/capture-sdk/actions/runs/33075234427/job/98527813602?pr=1151)) since I didn't do that many iOS 27 specific tests; given how flaky they've been lately, I didn't want to introduce a second chance in each PR for this to fail. If the recent changes I merged help make CI a bit more stable, I can definitely add it too

platform/swift/source/reports/MetricKitLegacyCrashAdapter.swift
ArielDemarco: In practise, whenever we're dealing with this kind of crash/oom diagnostic reports (not metrics) there's only one  rootFrame. We have a similar logic in different places. If we ever want to handle metrics or other types of diagnostics, we might have to change this

platform/swift/source/reports/MetricKitLegacyCrashAdapter.swift
ArielDemarco: They answer different questions; `#if compiler(>=6.4)` is build-time and checks whether the code exists at all as the code being wrapped only exists in the iOS 27 SDK. Since swift has no `#if sdk(>=)`, the compiler version is the (not so clear) proxy for "is using Xcode 27 toolchain". Without this, there'll type-check errors when trying to compile in Xcode 26. On the other hand, `@available(iOS 27.0, *)` works in runtime and checks if it should execute or not based on some local conditions. This is important as we still have our minimum deployment in iOS 15, so even an Xcode 27 build needs the OS check for devices with older OSes

test/platform/swift/unit_integration/core/reports/CrashReporterServiceTests.swift
snowp: If flakiness of some of the other tests are the concern we could do dedicated test targets (e.g. using test filters) to specify which ones we want to run on xcode 27 and bake that into the build. Will leave this all to you

ArielDemarco: I didnt think about it, that's a really good idea. I'll do this on the [last PR of the stack](https://github.com/bitdriftlabs/capture-sdk/pull/1151) as seems not so complicated

PR #1089, opened 2026-08-10
build: upgrade to Bazel 9

snowp:
Upgrades us to the latest Bazel version and updates a number of rules to ensure compatibility

- The latest version of rules_rust embeds the stdlib into the generated allocator lib. We patch this out to ensure that for the cases that we rely on this allocator library (ios) we'll exclude the stdlib binaries. This avoids shipping two copies of the Rust stdlib
- Updates one of the iOS tests to be less likely to time out: by carefully tearing down expectations we avoid them firing after the test has completed and breaking the test harness. 
- Enables the use of vendored protoc, this should avoid having to recompile this all the time locally
- Adds regression tests for capturing duplicate stdlib symbols in the ios build

Co-Authored-By: GPT-5 <codex@openai.com>


MODULE.bazel
```diff
@@ -36,7 +36,8 @@ COMPOSE_VERSION = "1.7.6"
 
 COMPOSE_COMPILER_VERSION = "1.5.15"
 
-KOTLIN_COMPILE_VERSION = "2.1.21"
+# Matches the compiler bundled by rules_kotlin 2.3.20; source language remains 1.9.
+KOTLIN_COMPILE_VERSION = "2.3.20"
```
snowp, line 40: @murki I think this is fine since we're only updating the compiler version and not the language version?

MODULE.bazel
```diff
@@ -106,15 +130,15 @@ new_git_repository(
 )
 
 git_override(
-    module_name = "google_bazel_common",
-    commit = "c35b0339ae7d7ac95761f69f4a0eed033163cc80",
-    remote = "https://github.com/google/bazel-common",
+    module_name = "hedron_compile_commands",
+    commit = "868c433f823d2d4eb7ce68198cd9d38bf75d61ae",
+    remote = "https://github.com/helly25/bazel-compile-commands-extractor.git",
```
snowp, line 135: Using a new fork since the old one is somewhat unmaintained and doesn't work with Bazel 9

MODULE.bazel
murki: should be (famous last words). can you make sure to update platform/jvm/gradle/libs.versions.toml to keep these in sync? (and test it with the gradle app)

snowp: I did this, seems to work but I had to disable a lint saying Kotlin 1.9 is deprecated. Seems fine for now but might mean we have to think about what we need to do to upgrade

murki: yeah, there's a ticket somewhere to move to 2 eventually, we were mostly waiting for more customer to transition

mattklein123 (approved):

PR #1090, opened 2026-08-10
session: add mobile session configuration APIs

snowp:
Instead of relying on two distinct strategy options with different APIs this consolidates the session API into a simpler configuration based API:

- For activity-like session strategies, the caller specifies the inactivity to invalidate sessions with. Upon invalidation the SDK will always generate a standard UUID session ID.
- For fixed-like workflows, the caller can provide the initial session ID and update it with startNewSession(id). If a session is not provided the SDK will default to a standard UUID session ID.
- In all cases, calling startNewSession with an ID will update the session to that ID - for activity-like workflows this means that the caller would potentially mix user-provided and SDK provided IDs which would be discouraged through docs.
- The existing strategy-based APIs become compatibility shims for the new API
- We no longer provide a session generator callback: users that expect to control the session should set no activity threshold, provide an initial ID and pass the new ID through startNewSession 

- [x] `CHANGELOG.md`'s "Unreleased" section has been updated.


platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/providers/session/SessionCallback.kt
```diff
@@ -0,0 +1,21 @@
...
+    private val onSessionIdChanged: (String) -> Unit,
+    private val mainThreadHandlerOverride: MainThreadHandler? = null,
+) {
+    private val mainThreadHandler by lazy { mainThreadHandlerOverride ?: MainThreadHandler() }
+
+    fun sessionIdChanged(sessionId: String) {
+        mainThreadHandler.run {
+            onSessionIdChanged.invokeCatchingOrThrowOnDebug(sessionId)
+        }
+    }
```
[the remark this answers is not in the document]
snowp: This is how it was done before so I left this as-is. iOS always does an async call to main but there didn't seem to be any reason to change this

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/providers/session/SessionCallback.kt
```diff
@@ -0,0 +1,25 @@
...
+
+package io.bitdrift.capture.providers.session
+
+import io.bitdrift.capture.common.MainThreadHandler
+import io.bitdrift.capture.utils.invokeCatchingOrThrowOnDebug
+
+/** Dispatches session changes from the Rust core to the app callback on the Android main thread. */
+internal class SessionCallback(
+    private val onSessionIdChanged: (String) -> Unit,
+    private val mainThreadHandlerOverride: MainThreadHandler? = null,
```
murki: why is this an optional param here?

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/providers/session/SessionConfiguration.kt
```diff
@@ -0,0 +1,69 @@
...
+ * This is the canonical session API. [SessionStrategy] is retained only as a compatibility shim.
+ *
+ * Capture creates a session when the SDK starts and whenever
+ * [io.bitdrift.capture.Capture.Logger.startNewSession] is called. An SDK-created session ID is a
+ * UUID. Without [inactivityTimeout], Capture uses [initialSessionId] whenever the SDK starts; it
+ * never reuses a persisted session. With [inactivityTimeout], [initialSessionId] seeds only the
+ * first session. Later SDK starts reuse the persisted session while it remains active and create
+ * an SDK UUID only after the inactivity period has elapsed.
+ *
+ * When [inactivityTimeout] is set, a period of inactivity rotates the session to an SDK-created
```
murki, line 26: I wonder if we should make more explicit the "persisting" logic here. Basically call out that without this param the sessionID will never survive process restarts

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/providers/session/SessionStrategy.kt
```diff
@@ -7,29 +7,24 @@
...
 sealed class SessionStrategy {
+    internal data class Configuration(
+        val configuration: SessionConfiguration,
+    ) : SessionStrategy()
+
     /**
-     * A session strategy that never expires the session ID but does not survive process restart.
-     *
-     * The initial session ID is retrieved by calling the passed closure.
+     * Deprecated compatibility shim for a session that does not expire during a process but is
```
murki, line 21: if we actually want to warn users for this we should use the annotation `@Deprecated("[message]")`

platform/swift/source/Capture.swift
```diff
@@ -68,6 +68,38 @@ extension Logger {
...
+    /// - parameter dateProvider:         An optional date provider for the default logger.
+    /// - parameter startResult:          An optional callback invoked with the SDK initialization result.
+    ///
+    /// - returns: A logger integrator that can enable SDK integrations.
+    @discardableResult
+    public static func start(
+        withAPIKey apiKey: String,
+        sessionConfiguration: SessionConfiguration = .init(),
+        configuration: Configuration = .init(),
+        fieldProviders: [FieldProvider] = [],
```
ArielDemarco, line 90: Based on internal conversation: maybe we want to avoid exposing this in new `start` APIs to encourage the use of `addField` instead and pass an empty array directly to the `self.start` method.

platform/swift/source/Capture.swift
```diff
@@ -104,12 +136,26 @@ extension Logger {
         return Self.getShared()?.sessionURL
     }
 
-    /// Initializes a new session within the currently configured logger.
+    /// Creates a new session with an SDK-created ID within the currently configured logger.
+    ///
     /// The logger must be started before this operation for it to take effect.
     public static func startNewSession() {
         Self.getShared()?.startNewSession()
```
ArielDemarco: Nit: 
```suggestion
        Self.startNewSession(sessionID: nil)
```

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/Capture.kt
```diff
@@ -166,6 +167,49 @@ object Capture {
...
+        /**
+         * Initializes Capture with the canonical session configuration API.
+         *
+         * Calling other SDK methods has no effect unless the logger has been initialized.
+         * Subsequent calls to this function have no effect. See [SessionConfiguration] for the
+         * session-ID lifecycle contract.
+         *
+         * @param apiKey The API key provided by bitdrift. This is required.
+         * @param sessionConfiguration Session lifecycle configuration. By default, Capture starts
+         * with an SDK-created UUID and does not rotate sessions due to inactivity.
```
murki: maybe should add that new session UUID are generated per-process

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/Capture.kt
```diff
@@ -244,14 +288,27 @@ object Capture {
...
 
+        /**
+         * Creates a new session with an app-provided ID within the currently running logger.
+         *
+         * A non-empty [sessionId] becomes the new session ID. When [sessionId] is null or empty,
+         * Capture generates a UUID. Every call invokes the configured session-ID callback, even if
+         * [sessionId] equals the current ID. If no logger is started, this is a no-op.
+         */
+        @JvmStatic
+        fun startNewSession(sessionId: String?) {
```
murki: why not just add an optional param to the existing method above? `@JvmStatic` should do it's thing

murki (approved):
android lgtm - this is so much nicer

platform/swift/source/CaptureRustBridge.h
```diff
@@ -42,7 +45,9 @@ void capture_report_error(const char *message);
 logger_id capture_create_logger(
     const char *_Nullable path,
     const char *api_key,
-    id<SessionStrategyProvider> session_strategy_provider,
+    NSString *_Nullable initial_session_id,
+    double inactivity_timeout_seconds,
+    _Nullable id session_callback,
```
ArielDemarco: The replacement here from `id<SessionStrategyProvider>` to a plain `id` is somewhat insecure. As far as I can see, the way the rust layer is invoking this through `msg_send![*self.swift_object, sessionIDChanged:*session_id]`, so if for some reason, somebody changes the name of the method in the receiver object, this wont work. 

To make is a bit more "secure" (and also easy to catch issues on compilation time) I'd create a simple protocol in this bridge, e.g.:
```objc
@protocol CAPSessionCallbackProvider <NSObject>
 - (void)sessionIDChanged:(NSString *)sessionID;
@end
```
And then, you make `SessionCallbackBridge` to conform to that protocol (and also the factory method to return this interface rather than the brdige). In that way you can repplace this `_Nullable id` to a `_Nullable id<CAPSessionCallbackProvider>`

platform/swift/source/Logger.swift
```diff
@@ -88,6 +88,24 @@ public final class Logger {
...
+    ) {
+        self.init(
+            withAPIKey: apiKey,
+            configuration: configuration,
+            sessionStrategy: .configuration(sessionConfiguration),
+            dateProvider: dateProvider,
+            fieldProviders: fieldProviders,
+            loggerBridgingFactoryProvider: loggerBridgingFactoryProvider
+        )
+    }
```
ArielDemarco: Seems nobody is using this, and is internal. we might want to remove it

platform/swift/source/Logger.swift
```diff
@@ -419,8 +437,20 @@ extension Logger: Logging {
         (self.underlyingLogger as? CoreLogger)?.isTracingActive == true
     }
 
+    /// Creates a new session with an SDK-created ID.
     public func startNewSession() {
-        self.underlyingLogger.startNewSession()
+        self.underlyingLogger.startNewSession(sessionID: nil)
```
ArielDemarco: Nit:
Call the other local method instead
```swift
        Self.startNewSession(sessionID: nil)
```

platform/swift/source/Logging.swift
```diff
@@ -189,6 +195,12 @@ public protocol Logging {
     ) -> Span
 }
 
+public extension Logging {
+    func startNewSession(sessionID _: String?) {
+        self.startNewSession()
+    }
+}
```
ArielDemarco: Is this necessary?  I think we can remove it (and if we use it for testing purposes, we can remove the `public`)

platform/swift/source/SessionConfiguration.swift
```diff
@@ -0,0 +1,91 @@
...
+/// discouraged unless the application can handle both forms.
+/// Empty initial and explicit IDs are treated as absent, so Capture generates a UUID instead.
+///
+/// `onSessionIDChanged` is called after Capture starts the initial session, rotates after
+/// inactivity, and on every explicit session start. An explicit start invokes the callback even
+/// when it supplies the current session ID. It is always called asynchronously on the main queue.
+/// When session starts overlap, callbacks can arrive in a different order from their state
+/// transitions. Treat the callback ID as belonging to that individual start; use the logger's
+/// `sessionID` property when the current session ID is required.
+public struct SessionConfiguration {
```
ArielDemarco: I'd make this an object that is able to be seen by the objective-c runtime. In this way we can entirely remove the `SessionConfigurationObjc`  and you keep all the core logic in a single object. Also, all the properties are objc compatible.

```suggestion
@objc(CAPSessionConfiguration)
public class SessionConfiguration: NSObject {
```

Important: if you implement this change, you'll probably have to change al the references to `underlyingConfiguration` in `LoggerObjc`

platform/swift/source/SessionStrategy.swift
```diff
@@ -5,25 +5,24 @@
...
-    /// - parameter sessionIDGenerator: The closure that returns the session ID to use. Upon the
-    ///                                 initialization of the logger the closure is called on the thread
-    ///                                 that's used to configure the logger. Subsequent closure calls are
-    ///                                 performed every time `Logger.startNewSession` method is called using
-    ///                                 the thread on which the method is called.
-    case fixed(sessionIDGenerator: (() -> String) = { UUID().uuidString })
+    /// - returns: The fixed-session compatibility shim.
+    public static func fixed() -> Self {
+        .fixed
+    }
```
ArielDemarco, line 27: This seems unnecessary. All the places where the `.fixed()` is being called could be replaced by `.fixed` directly

platform/swift/source/SessionStrategy.swift
```diff
@@ -35,69 +34,27 @@ public enum SessionStrategy {
...
         case .fixed:
-            CapturePassable.SessionStrategyType.fixed
-        case .activityBased:
-            CapturePassable.SessionStrategyType.activityBased
+            SessionConfiguration()
+        case let .activityBased(inactivityThresholdMins, onSessionIDChanged):
+            SessionConfiguration(
+                inactivityTimeout: TimeInterval(inactivityThresholdMins * 60),
+                onSessionIDChanged: onSessionIDChanged
+            )
```
ArielDemarco, line 64: Nit: As far as i can see, we're validating this in the rust layer; but we could validate that the inactivityThresholdMins provided by the user is a "valid" theshold (e.g. > 0). We could provide a sane default and log a warning or something. Seems better than failing

platform/swift/source/Logger.swift
snowp: Removed, good callout

platform/swift/source/Logging.swift
snowp: I think Logging is a public protocol right now so this guards against downstream compile errors, not sure if we care about that?

platform/swift/source/SessionStrategy.swift
snowp: This is another public API compatibility thing I believe since .fixed() is exposed via the API and used to compile, but maybe us marking this all as deprecated is fine and we can skip that? wdyt

platform/swift/source/SessionStrategy.swift
snowp: Will have this default to disabled on invalid value and document it, make sense to me

platform/swift/source/Capture.swift
snowp: I think I'll punt on this for now, I would like to make some other API changes to better support static fields known at initialization time before deprecating FieldProvider, will handle that there

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/Capture.kt
snowp: My agent complains that it won't work, let me dig into it a bit

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/providers/session/SessionCallback.kt
snowp: Removed, left over from a previous test iteration

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/providers/session/SessionConfiguration.kt
snowp: Updated kotlin + ios to call this out explicitly

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/providers/session/SessionStrategy.kt
snowp: Updated kotlin + ios to actually mark these as deprecated

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/Capture.kt
snowp: needs JvmOverloads, updating this

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/Capture.kt
snowp: making this clearer

platform/swift/source/Capture.swift
ArielDemarco: sounds good to me!

platform/swift/source/Logging.swift
ArielDemarco: Protocol extensions give a "default" implementation to an implementer. So if an implementer of `Logging` doesn't implement `startNewSession(sessionID: String)`, it'll still have this implementation. Risk is that if they implement `startNewSession`, and that calls `startNewSession(sessionID: String)` it'd create a loop.
IMO it's harmeless right now because `Logger` implements it

snowp: Makes sense, removed it

platform/swift/source/SessionStrategy.swift
ArielDemarco: Oh I see; the older `SessionStrategy` had a `.fixed()` value so you mimicking it. That's smart. I'd leave it as is; you may wanna mark it as deprecated but that's up to you if you want to do it now or in another PR

platform/swift/source/SessionConfiguration.swift
```diff
@@ -0,0 +1,117 @@
...
+/// inactivity, and on every explicit session start. An explicit start invokes the callback even
+/// when it supplies the current session ID. It is always called asynchronously on the main queue.
+/// When session starts overlap, callbacks can arrive in a different order from their state
+/// transitions. Treat the callback ID as belonging to that individual start; use the logger's
+/// `sessionID` property when the current session ID is required.
+@objc(CAPSessionConfiguration)
+public final class SessionConfiguration: NSObject {
+    private static let logger = os.Logger(
+        subsystem: "io.bitdrift.capture.SessionConfiguration",
+        category: "configuration"
```
ArielDemarco: Nit 
FWIW: there's an `OSLogger` abstraction

ArielDemarco (approved):
PR LGTM. Loved the change 👍🏻

platform/swift/source/SessionConfiguration.swift
snowp: nice, updated!

PR #1093, opened 2026-08-10
avoid rebuilding apk during size CI check

snowp:
Previously we'd build it multiple times with different --config options, which would prevent cache reuse. Instead this explicitly builds it all once then hands the AAR over to the binary size script


murki (approved):
chef's kiss

PR #1095, opened 2026-08-12
use bazel to build the rust library used by gradle

snowp:


.gitignore
```diff
@@ -38,4 +38,3 @@ rust-project.json
 /perf.data
 /perf.data.old
 vendor
-platform/jvm/capture/src/main/jniLibs/
```
murki, line 41: what was this about?

platform/jvm/capture/build.gradle.kts
```diff
@@ -96,100 +97,107 @@ android {
...
+    // Build the Bazel test target, which adds test helpers to the production JNI library.
     commandLine(
-        "cargo",
+        "./bazelw",
         "build",
-        "--package",
-        "test_jni",
-        "--target",
-        platformConfig.rustTarget,
+        "//test/platform/jvm:capture",
```
murki, line 193: so nice

.gitignore
snowp: The cargo-ndk plugin would stage the changes into the repo before putting them into the gradle build, now it just copies straight from the bazel-bin directory into the gradle build so much cleaner and we don't need this .gitignore

platform/jvm/capture/build.gradle.kts
snowp: it's all covered by the bazel build graph, so it's all pretty nice

murki (approved):
unblocking, tests and iterations will catch whatever breaks!

snowp:
Seems to work in most cases I tested and simplifies stuff like no longer having to deal with missing Cargo paths etc., seems more robust. Updated the wiki with the new instructions: just less things to do and less troubleshooting

PR #1107, opened 2026-08-14
bump shared core again for more stats changes

mattklein123:
---


platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/LoggerImpl.kt
```diff
@@ -505,9 +505,11 @@ internal class LoggerImpl(
                 matchingArrayFields.keys,
                 matchingArrayFields.values,
                 previousRunSessionId,
-                occurredAtTimestampMs,
-                blocking,
+                occurredAtTimestampMs
             )
+            if (blocking) {
+                flush(true)
```
murki:
```suggestion
                flush(blocking = true)
```

murki: named params help with readability in this case

murki: or even better, just pass the blocking var down

murki (approved):

PR #1113, opened 2026-08-15
improve CI cache reuse

snowp:
1. Remove the bazel repository cache. This is not very useful anymore with BB and we were thrashing the cache at 10G.
2. Have ./bazelw reuse the bazelisk and Android SDK cache used elsehwere. This allows us to effectively share the cache and avoids downloading the Android SDK/NDK for more tests.
3. Support passing extra flags to bazel-diff to minimize analysis-cache churn
4. Avoid writing most caches during PR builds: these are not reusable by other PRs/main and they end up eating up cache storage causing cache evictions


.github/workflows/format.yaml
```diff
@@ -30,11 +30,6 @@ jobs:
           key: ${{ runner.os }}-formatters-${{ hashFiles('./ci/setup_linux_format.sh') }}
       - run: ./ci/setup_linux_format.sh
         if: steps.cache-formatters.outputs.cache-hit != 'true'
-      - run: |
```
snowp, line 33: No longer necessary

murki (approved):
maybe we should do a fake release dry-run? just to make sure all the release job still work properly

PR #1119, opened 2026-08-18
Upgrade jni to 0.22

snowp:
This is a fairly involved breaking change in the jni API but should provide better safety guarantees

---


snowp:
This slows down the hot path logging path by quite a bit, so blocked on figuring that out

PR #1120, opened 2026-08-18
[BIT-9591] Include extra information in crash report to support downloadable CR

ArielDemarco:
---


platform/swift/source/crash_handler/BitdriftKSCrashHandler.m
```diff
@@ -174,7 +189,10 @@ - (BOOL)startCrashReporterWithError:(NSError **)error {
     memset(&g_crashHandlerReportContext, 0, sizeof(g_crashHandlerReportContext));
     // This gets allocated once and lives forever.
     g_crashHandlerReportContext.reportPath = strdup(self.kscrashReportFilePath.UTF8String);
+    g_crashHandlerReportContext.bundlePath = strdup(NSBundle.mainBundle.bundlePath.UTF8String);
```
kattrali: I think its possible for this to be null in a unit test bundle, might be a problem when the package is installed in a host app since `strdup(nil)` would not behave.

platform/swift/source/reports/MetricKitDiagnosticManager.swift
```diff
@@ -74,11 +75,14 @@ final class MetricKitDiagnosticManager: MetricKitDiagnosticManaging {
         self.fileManager = fileManager
         self.crashEnrichmentSummaryHandler = crashEnrichmentSummaryHandler
         self.completionHandler = completionHandler
+        let appDistribution = appDistributionInspector.inspect()
```
kattrali, line 78: this is the data from the subsequent launch, not the crash session, right?

platform/swift/source/crash_handler/BitdriftKSCrashHandler.m
ArielDemarco: Fixed! Good catch :)

platform/swift/source/reports/MetricKitDiagnosticManager.swift
ArielDemarco: Yep! good catch. I've modified it so it saves it on the crash report now

kattrali (approved):

PR #1130, opened 2026-08-19
andr; flush logger on window focus lost

murki:
Flush the logger (non-blocking) when the app's window loses focus. The existing backgrounding
flush hangs off `ProcessLifecycleOwner`'s `ON_STOP`, which never fires while the recents overview
is showing — an app swiped away from there loses whatever hadn't reached disk. Window focus is the
only signal that fires in that case.

- `WindowFocusFlushLogger` (`IEventListenerLogger`) registers a `ViewTreeObserver.OnWindowFocusChangeListener`
  (API 18) per started activity via an injectable `IWindowFocusRegistrar`; over-firing (IME, dialogs,
  transitions) is fine — flushes are debounced downstream.
- Kill switch: `client_feature.android.logger_flushing_on_window_focus_loss` (default on, re-read per callback).
- Unit tests (explicit fakes) + Robolectric tests mirror the on-device scenario matrix.
- Test app: `FocusMatrixActivity` exercises the IME / permission-dialog / transition focus-loss cases.

Verified on a Pixel 10: recents, recents+kill, home, and activity-navigation all show
`focus=false` → flush within 1ms, upload acked before process death. Rotation drops no focus (measured).

---

### DIFF
--- a/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/events/lifecycle/WindowFocusFlushLogger.kt
+++ b/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/events/lifecycle/WindowFocusFlushLogger.kt
@@ -1,0 +1,116 @@
+// capture-sdk - bitdrift's client SDK
+// Copyright Bitdrift, Inc. All rights reserved.
+//
+// Use of this source code is governed by a source available license that can be found in the
+// LICENSE file or at:
+// https://polyformproject.org/wp-content/uploads/2020/06/PolyForm-Shield-1.0.0.txt
+
+package io.bitdrift.capture.events.lifecycle
+
+import android.app.Activity
+import android.app.Application
+import android.os.Bundle
+import io.bitdrift.capture.IInternalLogger
+import io.bitdrift.capture.common.IWindowManager
+import io.bitdrift.capture.common.MainThreadHandler
+import io.bitdrift.capture.common.Runtime
+import io.bitdrift.capture.common.RuntimeFeature
+import io.bitdrift.capture.events.IEventListenerLogger
+
+/**
+ * Makes buffered data durable when the app's window loses focus.
+ *
+ * The gap this closes: the SDK's backgrounding flush hangs off `ProcessLifecycleOwner`'s `ON_STOP`,
+ * and `ON_STOP` never fires for the app switcher — while the overview is showing, the recents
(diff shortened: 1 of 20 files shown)


platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/events/lifecycle/WindowFocusFlushLogger.kt
```diff
@@ -0,0 +1,102 @@
...
+        // Defensive: covers an activity destroyed without a matching stop.
+        focusRegistrar.unregister(activity)
+    }
+
+    private fun onWindowFocusLost() {
+        if (!isStarted || !runtime.isEnabled(RuntimeFeature.LOGGER_FLUSHING_ON_WINDOW_FOCUS_LOSS)) {
+            return
+        }
+        // Focus callbacks are dispatched on the main thread; a blocking flush would stall it.
+        logger.flush(blocking = false)
```
murki: this is the crux of the change

platform/jvm/capture/src/test/kotlin/io/bitdrift/capture/events/lifecycle/WindowFocusFlushLoggerTest.kt
```diff
@@ -0,0 +1,198 @@
...
+        windowFocusFlushLogger.start()
+        windowFocusFlushLogger.onActivityStarted(activity)
+        windowFocusFlushLogger.onActivityStarted(secondActivity)
+
+        windowFocusFlushLogger.stop()
+
+        assertThat(focusRegistrar.registeredActivities).isEmpty()
+    }
+
+    // The tests below mirror the on-device scenario matrix (recents is the plain focus-loss case
```
snowp: This is referring to the testing you did right? Feels like this comment will become hard to understand quickly if we throw away the test scenarios

platform/jvm/gradle-test-app/src/main/res/values/strings.xml
```diff
@@ -53,4 +53,9 @@
     <string name="restart_warning_summary">Click here to restart and apply your configuration changes</string>
     <string name="log_message_toast">Log message sent with level: %1$s</string>
     <string name="webview_testing">WebView Testing</string>
+    <!-- Debug-only focus-matrix harness strings; not translated. -->
```
snowp: we can probably drop this haha

snowp (approved):

PR #1135, opened 2026-08-19
android/ios: migrate OOTB fields to stateful updates

snowp:
Migrates Android and iOS foreground, locale, and network out-of-the-box fields from per-log metadata providers to the native OOTB field store. Each platform seeds the store during logger creation and publishes only later state changes, avoiding per-log provider and FFI evaluation while preserving the existing field values. Custom field providers remain available for application-defined fields.

Validated against live Android and iOS simulator sessions on `.dev`: [Android](https://timeline.bitdrift.dev/session/71aca7f7-4de5-441d-a49b-15357defdaad) covers foreground, Wi-Fi/cellular, and `fr_CA` locale updates; [iOS](https://timeline.bitdrift.dev/session/de2e0613-38cd-43a2-a901-7156c193520f) confirms the initial OOTB field set and startup foreground transition.


platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/events/lifecycle/ForegroundStateListener.kt
```diff
@@ -0,0 +1,47 @@
...
+import androidx.lifecycle.LifecycleEventObserver
+import androidx.lifecycle.LifecycleOwner
+import io.bitdrift.capture.IInternalLogger
+import io.bitdrift.capture.attributes.ClientAttributes
+import io.bitdrift.capture.common.MainThreadHandler
+
+/**
+ * Keeps the foreground OOTB field current independently of configurable lifecycle-event logging.
+ */
+internal class ForegroundStateListener(
```
snowp: To discuss: we have an existing listener for this but it's gated (off by default, on via remote config) which means switching over is technically a regression for fresh app installs. Might be fine or we do a targeted listener here that only updates the field

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/events/lifecycle/ForegroundStateListener.kt
```diff
@@ -0,0 +1,47 @@
...
+        }
+    }
+
+    override fun onStateChanged(
+        source: LifecycleOwner,
+        event: Lifecycle.Event,
+    ) {
+        when (event) {
+            Lifecycle.Event.ON_START -> logger.updateOotbField(ClientAttributes.FOREGROUND_KEY, "1")
+            Lifecycle.Event.ON_STOP -> logger.updateOotbField(ClientAttributes.FOREGROUND_KEY, "0")
```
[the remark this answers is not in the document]
snowp: This is true but an explicit trade-off for now: state updates (both now and in the future) are unlikely to be dropped at the frequency that we're emitting them, so for now we'll forego any kind of reconciliation logic to keep things simple

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/LoggerImpl.kt
```diff
@@ -255,6 +257,7 @@ internal class LoggerImpl(
         this.loggerId = loggerId
 
         runtime = JniRuntime(this.loggerId)
+        ForegroundStateListener(this, ProcessLifecycleOwner.get()).start()
```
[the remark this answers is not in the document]
snowp: We snap this as part of clientAttributes.iniitalOotbFields(), so we're only worried about this tiny inconsistency window. Seems fine but I could be convinced otherwise

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/events/lifecycle/ForegroundStateListener.kt
murki: Wait, isn't that `AppLifecycleListenerLogger` ? that one uses `client_feature.android.application_lifecycle_reporting` that I see has a default value of `true`. or were you thinking of a different one?

snowp: I don't know if this is well-known anymore but this is all gated by `platform_events.enabled` and enabled via shared-core once this flips on. Maybe we should just flip that to on by default at this point and then we can reuse the existing code freely and live with the fact that these state changes would disappear if we turned that off?

murki: yeah i wasn't aware of that. no reason the default for that would be off at this point

snowp: ok will make that change

murki: with that being said it might still make sense to have separate flags for the fields vs the logs

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/attributes/ClientAttributes.kt
```diff
@@ -132,7 +109,8 @@ internal class ClientAttributes(
     /**
      * Gets the current locale.
      */
-    private fun getCurrentLocale(): Locale? = ConfigurationCompat.getLocales(resources.configuration).get(0)
+    private fun getCurrentLocale(configuration: Configuration = resources.configuration): Locale? =
```
murki, line 112: why is this even required? wasn't the locale provider logic extracted to `LocaleAttributes`?

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/attributes/LocaleAttributes.kt
```diff
@@ -0,0 +1,58 @@
...
+     * The initial locale is installed synchronously when the logger is created, so starting this
+     * listener does not need to replay it.
+     */
+    fun start(logger: IInternalLogger) {
+        this.logger = logger
+        context.registerComponentCallbacks(this)
+    }
+
+    override fun onConfigurationChanged(newConfig: Configuration) {
+        val updatedLocale = currentLocale(newConfig)
```
murki, line 43: if there's any cost associated with checking the OS locale (e.g. IPC) we might want to bring back the diff logic from `updateLocaleIfNeeded()`

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/attributes/ClientAttributes.kt
snowp: It's because the crash reporting logic calls back into this and it goes through the ClientAttributes interface to do it, seemed simpler to keep this as-is instead of introducing another JNI interface for LocaleAttributes. Would be nice to make this all a bit better

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/attributes/LocaleAttributes.kt
snowp: Yeah let me check what the overhead of this is

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/LoggerImpl.kt
```diff
@@ -264,6 +261,8 @@ internal class LoggerImpl(
         this.loggerId = loggerId
 
         runtime = JniRuntime(this.loggerId)
+        localeAttributes.start(this)
+        networkAttributes.start(this)
```
murki: /nit, but if this is to become a common pattern, we might want to put it in an collection of interfaced classes that all get started centrally somewhere in here similar to EventsListenerTarget / IEventListenerLogger

platform/swift/source/Logger.swift
```diff
@@ -416,6 +418,17 @@ public final class Logger {
...
+    let size = UnsafeMutablePointer<Int>.allocate(capacity: 1)
+    sysctlbyname("hw.machine", nil, size, nil, 0)
+
+    var machine = [CChar](repeating: 0, count: size.pointee)
+    sysctlbyname("hw.machine", &machine, size, nil, 0)
+    size.deallocate()
+
+    return String(cString: machine)
+}
+
```
ArielDemarco: TBH, I'd move this back to `DeviceAttribute` or create a new abstraction (similar to `Debugger`) for this kind of syscalls

ArielDemarco (approved):
lgtm

platform/swift/source/Logger.swift
```diff
@@ -183,6 +181,9 @@ public final class Logger {
             bufferDirectoryPath: directoryURL.path,
             sessionStrategy: sessionStrategy,
             metadataProvider: metadataProvider,
+            initialOotbFields: self.appStateAttributes.initialOotbFields()
+                + self.localeAttributes.initialOotbFields()
+                + self.networkAttributes.initialOotbFields(),
```
ArielDemarco: Maybe we can do an interface/abstraction to actually group these. Should be simpler to add and follow the code too. All of those share the same behavior of having the `initialOotbFields`

snowp: Yeah will add this, should make this a bit simpler to reason about

platform/swift/source/Logger.swift
snowp: Will do, I moved it since it only had that static but if you prefer it there I don't mind

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/LoggerImpl.kt
snowp: Will make this change

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/attributes/LocaleAttributes.kt
snowp: From what I can tell reading Configuration is not going to invoke an IPC so this seems safe

murki (approved):
lgtm for android

platform/swift/source/default/attributes/TelephonyNetworkInfo.swift
```diff
@@ -25,15 +25,17 @@ final class TelephonyNetworkInfo: NSObject {
...
 
         super.init()
 
         // Keep track of the active data service for cases when device uses multiple SIMs.
         // All delegate's callbacks are dispatched asynchronously to a global queue with `default` QoS.
         self.underlyingNetworkInfo.delegate = self
-        self.updateDataServiceNetworkInfo()
+        self.stateQueue.sync {
+            self.updateDataServiceNetworkInfo()
+        }
```
ArielDemarco: You could probably get rid of this? you have a guard in the `updateDataServiceNetworkInfo`:
```swift
guard self.radioType != radioType else {
            return
        }
```
In line 28 you are already setting the `radioType` so this is an unnecessary queue jump

platform/swift/source/default/attributes/TelephonyNetworkInfo.swift
```diff
@@ -46,42 +48,52 @@ final class TelephonyNetworkInfo: NSObject {
                 // identifier of the currently active data service provider. Otherwise, the update
                 // is for a SIM (cellular provider) that's not actively used for data transfer
                 // and we ignore it.
-                if identifier == self.dataServiceIdentifier.load() {
-                    self.updateDataServiceNetworkInfo()
+                self.stateQueue.async {
+                    if identifier == self.dataServiceIdentifier {
+                        self.updateDataServiceNetworkInfo()
+                    }
```
ArielDemarco: You can probably remove this block (and the whole subscription as it's deprecated, and the delegate is covering iOS 15 too)

platform/swift/source/default/attributes/TelephonyNetworkInfo.swift
```diff
@@ -46,42 +48,52 @@ final class TelephonyNetworkInfo: NSObject {
...
 }
 
 extension TelephonyNetworkInfo: CTTelephonyNetworkInfoDelegate {
     public func dataServiceIdentifierDidChange(_ identifier: String) {
-        self.dataServiceIdentifier.update { $0 = identifier }
-        self.updateDataServiceNetworkInfo()
+        self.stateQueue.sync {
+            self.dataServiceIdentifier = identifier
+            self.updateDataServiceNetworkInfo()
+        }
```
ArielDemarco: Queue es serial, so i'd chagne this to `async` to avoid waiting on a global queue

platform/swift/source/default/attributes/TelephonyNetworkInfo.swift
```diff
@@ -46,42 +48,52 @@ final class TelephonyNetworkInfo: NSObject {
...
             }
     }
 
     func start(with logger: CoreLogging) {
-        self.logger = logger
-        self.publishRadioType()
+        self.stateQueue.sync {
+            self.logger = logger
+            self.publishRadioType()
+        }
```
ArielDemarco: I don't know in this particular context (because you might be calling first `start` and after the `initialOotbdFields`, but if that's not the case, then as the queue es serial, you could change this to `async` to avoid waiting on a global serial queue

platform/swift/source/default/attributes/TelephonyNetworkInfo.swift
```diff
@@ -46,42 +48,52 @@ final class TelephonyNetworkInfo: NSObject {
...
 
-        let radioType = self.underlyingNetworkInfo.radioType(for: identifier)
-        self.radioType.update { $0 = radioType }
+        self.radioType = radioType
         self.publishRadioType()
     }
 
     private func publishRadioType() {
-        self.logger?.updateOotbField(withKey: "radio_type", value: self.radioType.load() ?? "unknown")
+        self.logger?.updateOotbField(withKey: "radio_type", value: self.radioType ?? "unknown")
```
ArielDemarco, line 70: As you removed the `Atomic`, it'd be better to read this from inside the queue (or bring back the atomic as it's using a lock inside)

snowp: I think this is fine at the moment since publishRadioType is internal and only called from within the queue

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/providers/MetadataProvider.kt
```diff
@@ -43,7 +43,11 @@ internal class MetadataProvider(
                 errorHandler.handleError(message, e)
             }
         }
-        if (result.isEmpty()) return emptyArray()
+        if (result.isEmpty()) return EMPTY_FIELDS
         return result.toTypedArray()
     }
+
+    private companion object {
+        val EMPTY_FIELDS = emptyArray<Field>()
```
FranAguilera, line 51: maybe this could be moved into [FieldProvider](https://github.com/bitdriftlabs/capture-sdk/blob/be061430ff1564a00e17746e0845e5faf9fb94cb/platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/providers/FieldProvider.kt#L164) for consistency and to allow reuse if needed

snowp: I think for now since this is just an optimization on the JNI bridging I'll leave it here; I might end up redoing this to do the null optimization I did elsewhere but will do that in a separate PR

PR #1136, opened 2026-08-20
remove dependency on ndk from gradle

snowp:
We build the .so with Bazel so gradle doesn't need the NDK tooling

We preserve the binary stripping levels used by Gradle, allowing the level to be changed just through a Gradle config change

### DIFF
--- a/platform/jvm/capture/build.gradle.kts
+++ b/platform/jvm/capture/build.gradle.kts
@@ -65,6 +65,14 @@ android {
         }
     }
 
+    packaging {
+        jniLibs {
+            // Bazel already selects libcapture's strip level for each build type. Preserve the
+            // generated library so Android Gradle Plugin does not invoke an NDK strip tool.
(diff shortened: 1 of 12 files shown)


platform/jvm/capture/build.gradle.kts
```diff
@@ -117,7 +129,7 @@ val bazelAndroidTarget =
         else -> throw GradleException("Unsupported Rust Android target: $rustTarget")
     }
 
-fun registerBazelRustBuild(buildType: String, release: Boolean) =
+fun registerBazelRustBuild(buildType: String, release: Boolean, stripLevel: String? = null) =
```
murki: so what are the possible values for stripLevel? could it be an enum?

murki (approved):

platform/jvm/capture/build.gradle.kts
snowp: Good idea, I switched it to an enum so now it's all pretty well documented

PR #1137, opened 2026-08-20
simplify populating reports with static metadata

snowp:
Instead of passing ClientAttributes to the report builder and having to JNI out the values we can pull them straight from the metadata object. This avoids startup registration of these JNI handles and simplifies the code quite a bit.

The remaining values could also be moved into `Metadata` but I kept this light for now


platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/reports/processor/IStreamingReportProcessor.kt
```diff
@@ -6,40 +6,43 @@
...
-        stream: InputStream?,
-        timestampMillis: Long,
-        destinationPath: String,
-        attributes: IClientAttributes,
-        runningState: String?,
-        appExitDescription: String?,
-        memoryPressureLevel: Int,
-        isFileSizeOptimizationEnabled: Boolean,
-    )
+    fun processAndPersistANR(report: AnrReport)
```
FranAguilera, line 42: thanks for cleaning this up

FranAguilera (approved):

PR #1149, opened 2026-08-21
andr/ios: introduce initial fields and deprecate FieldProviders

snowp:
This adds `initialFields` to the constructor on both iOS and Android, which should make it possible to set fields that should be set from the start of the logger, then update them later on with addField. 

With this we can move to deprecate FieldProviders which are much more expensive due to high FFI overhead per field, especially on Android.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,11 +7,11 @@
 
 **Added**
 
-- Nothing yet!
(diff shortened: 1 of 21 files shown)


snowp:
Superseded by #1150, which contains this patch plus follow-up compatibility fixes and now uses shared-core main.

PR #1150, opened 2026-08-22
andr/ios: introduce initial fields and deprecate FieldProviders

snowp:
This adds `initialFields` to the constructor on both iOS and Android, which should make it possible to set fields that should be set from the start of the logger, then update them later on with addField.

With this we can move to deprecate FieldProviders which are much more expensive due to high FFI overhead per field, especially on Android.

Since we are still using FieldProviders internally they are redone as `FieldGetters`: this will go away as we migrate more of the internal ootb flags to rely in updateOotbField like #1135


platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/Capture.kt
```diff
@@ -140,30 +140,98 @@ object Capture {
...
+        )
         @Synchronized
         @JvmStatic
         @JvmOverloads
         fun start(
             apiKey: String,
             sessionStrategy: SessionStrategy,
             configuration: Configuration = Configuration(),
-            fieldProviders: List<FieldProvider> = listOf(),
+            fieldProviders: List<io.bitdrift.capture.providers.FieldProvider> = listOf(),
```
murki, line 199: why is this namespaced

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/Capture.kt
```diff
@@ -140,30 +140,98 @@ object Capture {
...
+
+        /**
+         * Initializes Capture with the specified API key, session strategy, and startup fields.
+         *
+         * This overload keeps startup fields ergonomic for Java callers without changing the existing
+         * Java overloads that accept legacy field providers.
+         */
+        @Synchronized
+        @JvmStatic
+        fun start(
```
murki: why so many explicit overloads, can't we collapse all non deprecated to a single @JvmOverloads

murki (approved):
good stuff

test/platform/swift/unit_integration/mocks/MockFieldProvider.swift
```diff
@@ -7,6 +7,11 @@
 
 import Capture
 
+@available(
+*,
+deprecated,
+message: "Use Logger.start(initialFields:) to seed fields at startup and Logger.addField(withKey:value:) to update them."
+)
```
ArielDemarco: I think you can actually delete this file

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/Capture.kt
snowp: This is to avoid having to do `@file:Suppress("DEPRECATION")` on this whole file, instead it puts it on the deprecated functions and avoids a package-level import

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/Capture.kt
snowp: I cleaned up one, apparently we still need some of them to avoid making initialFields required in the jvm overloads

platform/swift/source/Capture.swift
```diff
@@ -77,30 +78,149 @@ extension Logger {
...
+    /// - parameter dateProvider:         An optional date provider for the default logger.
+    /// - parameter startResult:          An optional callback invoked with the SDK initialization result.
+    ///
+    /// - returns: A logger integrator that can enable SDK integrations.
+    @discardableResult
+    public static func start(
+        withAPIKey apiKey: String,
+        sessionConfiguration: SessionConfiguration = .init(),
+        configuration: Configuration = .init(),
+        initialFields: Fields,
```
ArielDemarco: If you add a default value to `initialFields`, you might be able to get rid of the other `start` at line 86:
```suggestion
        initialFields: Fields = [:],
```

platform/swift/source/Logger.swift
```diff
@@ -81,13 +84,80 @@ public final class Logger {
...
             timeProvider: SystemTimeProvider(),
             loggerBridgingFactoryProvider: loggerBridgingFactoryProvider
         )
     }
 
+    @available(
+    *,
+    deprecated,
+    message: "Use Logger.start(initialFields:) to seed fields at startup and Logger.addField(withKey:value:) to update them."
+    )
```
ArielDemarco: Marking this in an internal class might be unnecessary

platform/swift/source/Logger.swift
```diff
@@ -81,13 +84,80 @@ public final class Logger {
...
+    convenience init?(
+        withAPIKey apiKey: String,
+        configuration: Configuration,
+        sessionStrategy: SessionStrategy,
+        dateProvider: DateProvider?,
+        fieldProviders: [FieldProvider],
+        initialFields: Fields = [:],
+        loggerBridgingFactoryProvider: LoggerBridgingFactoryProvider = LoggerBridgingFactory()
```
ArielDemarco: Nobody is using this method, we can remove it

platform/swift/source/Logger.swift
```diff
@@ -81,13 +84,80 @@ public final class Logger {
...
+        configuration: Configuration,
+        sessionStrategy: SessionStrategy,
+        dateProvider: DateProvider?,
+        fieldProviders: [FieldProvider],
+        initialFields: Fields = [:],
+        enableNetwork: Bool = true,
+        storageProvider: StorageProvider,
+        timeProvider: TimeProvider,
+        networkDelegateQueue: DispatchQueue = .network,
+        loggerBridgingFactoryProvider: LoggerBridgingFactoryProvider = LoggerBridgingFactory()
```
ArielDemarco: Nobody is using this method, we can remove it

platform/swift/source/src/bridge.rs
```diff
@@ -550,6 +551,7 @@ extern "C" fn capture_create_logger(
             .to_string(),
         },
       ));
+      let initial_custom_fields = unsafe { ffi::convert_fields(initial_fields) }?;
```
ArielDemarco: Shall the logger creation fail if we cannot convert the custom field? wouldn't be better to not include them and inform in some way? (e.g. logging in console or sending a log afterwards to the back)

ArielDemarco (approved):
Made some simple comments. PR looks good!

PR #1151, opened 2026-08-24
Changes to CI to support multiple xcframeworks on release

ArielDemarco:

### DIFF
--- a/.github/workflows/ios.yaml
+++ b/.github/workflows/ios.yaml
@@ -146,6 +146,7 @@ jobs:
           ./bazelw build \
             --config ci \
             --xcode_version="$XCODE_VERSION" \
+            --ios_simulator_version=27.0 \
             //platform/swift/source:capture_ios_static_framework
         env:
           XCODE_VERSION: ${{ steps.plan.outputs.xcode_version }}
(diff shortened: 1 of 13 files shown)


FranAguilera (approved):

kattrali (approved):

.github/workflows/ios.yaml
```diff
@@ -149,6 +149,19 @@ jobs:
             //platform/swift/source:capture_ios_static_framework
         env:
           XCODE_VERSION: ${{ steps.plan.outputs.xcode_version }}
+
+      - name: Test the iOS 27 code paths with Xcode 27
+        run: |
+          ./bazelw test \
```
snowp, line 156: You can see this in the CI logs 
```
WARNING: Build options --ios_simulator_version and --java_runtime_version have changed, discarding analysis cache (this can be expensive, see https://bazel.build/advanced/performance/iteration-speed).
``` 
which means that we're spending time reconfiguring the build. I think we can update .bazelrc to do

```
common --java_runtime_version=remotejdk_11
```
instead of putting it on `test` and then pass the same --io_simulator_version to the build command. The simulator version on the build command doesn't do anything but it will avoid invalidating the analysis call on the next call

snowp (approved):
Seems right to me, one suggestion to speed up CI that you can do here or in a followup

.github/workflows/ios.yaml
ArielDemarco: Done!

PR #1157, opened 2026-08-26
perf(android): batch JNI local reference cleanup

snowp:
Batches string-field conversion in bounded 64-entry JNI local-reference frames, avoiding a frame transition for every field. Each batch reserves only its 128 actual element refs, and `aarch64-linux-android` maps to the arm64 Gradle build. Focused field benchmarks remain available for diagnostics but are omitted from the CI PR report.

On the same locally connected Pixel 9a, the 64-entry frame improved the 10/50/100-field samples by 6.8%/7.7%/7.4%, respectively. An API 23 arm64 emulator also passed the 100- and 1000-field scenarios without local-reference or CheckJNI errors.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,8 +14,6 @@
 
 **Changed**
 
-- Nothing yet!
-
 - Legacy fixed-session strategy generator callbacks have been removed. Use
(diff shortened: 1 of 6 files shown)


FranAguilera (approved):
nice

PR #1161, opened 2026-08-26
andr;sdk: Fix partial transparency issue in ModalBottomSheet for session replay

murki:
### Session Replay: fix overlay windows used in modals to no longer occlude the replay fram

A dialog, bottom sheet or popup was captured as an opaque rect spanning its whole window, painting over every window beneath it and rendering the frame empty.

original: 
Screenshot_20260826-134002

|[before](https://timeline.bitdrift.io/session/8c636bf6-c0f3-4783-9c1c-ff1d5ddc88e3#sessionReplay)|[after](https://timeline.bitdrift.io/session/ab5c5e7f-070c-4630-b0cf-bf91adac5057#sessionReplay)|
|---|---|
|Screenshot 2026-08-26 at 11 02 38 AM|Screenshot 2026-08-26 at 3 53 44 PM|
|Screenshot 2026-08-26 at 11 02 55 AM|Screenshot 2026-08-26 at 4 01 16 PM|

**Legacy Android views:**

- `DecorView` resolved through the mapper table straight to `ReplayType.View`, so its background opacity was never consulted, it is now reported as `ReplayType.TransparentView`.
- Opacity handling was split between `BackgroundMapper` and the mapper table, and the two disagreed. Centralized in `BackgroundOpacity`, now consulted by both paths.
- Opaque backgrounds map to `View` (previously `BackgroundImage`); transparent, unknown and absent backgrounds map to `TransparentView` (previously `Ignore` or nothing). `BackgroundImage` is no longer emitted on Android.

**Compose views:**

- Every layout container with no text and no role fell through to `ReplayType.View`, so full-screen containers inside an overlay — the `ModalBottomSheet` scrim, for instance — occluded the app.
- Overlay windows are now detected once per window via `SemanticsProperties.IsDialog` / `IsPopup`, and generic containers inside them are emitted as `TransparentView`. Containers outside an overlay are unchanged.
- An interop `AndroidView` inside an overlay was no longer handed to the Android traversal, dropping its entire subtree. The interop check now runs against the semantic type, before the overlay adjustment.

Fixes BIT-9568 and BIT-6866

---

### DIFF
--- a/.github/workflows/android.yaml
+++ b/.github/workflows/android.yaml
@@ -322,6 +322,9 @@ jobs:
     - name: Gradle capture-plugin unit tests
       working-directory: ./platform/jvm
       run: ./gradlew capture-plugin:test --info
+    - name: Gradle replay unit tests
+      working-directory: ./platform/jvm
+      run: ./gradlew replay:testDebugUnitTest --info
     - name: Check Licenses for modules
       working-directory: ./platform/jvm
       run: ./gradlew replay:checkLicense common:checkLicense capture:checkLicense
--- a/.github/workflows/android.yaml
+++ b/.github/workflows/android.yaml
@@ -322,9 +322,6 @@ jobs:
     - name: Gradle capture-plugin unit tests
       working-directory: ./platform/jvm
       run: ./gradlew capture-plugin:test --info
-    - name: Gradle replay unit tests
-      working-directory: ./platform/jvm
-      run: ./gradlew replay:testDebugUnitTest --info
     - name: Check Licenses for modules
       working-directory: ./platform/jvm
       run: ./gradlew replay:checkLicense common:checkLicense capture:checkLicense
--- a/platform/jvm/gradle-test-app/build.gradle.kts
+++ b/platform/jvm/gradle-test-app/build.gradle.kts
@@ -77,9 +77,8 @@ dependencies {
     implementation("com.squareup.retrofit2:converter-gson:3.0.0")
 
     debugImplementation("androidx.compose.ui:ui-test-manifest")
-    // ui-test-manifest drags androidx.test:core 1.5.0 onto debugRuntimeClasspath, and AGP's
-    // consistent resolution then pins androidTest to it. Espresso <3.7 calls
-    // InputManager.getInstance(), removed in Android 17, so the floor has to come up here too.
+    // Raises the floor AGP's consistent resolution pins androidTest to; Espresso <3.7 calls
+    // InputManager.getInstance(), removed in Android 17.
     debugImplementation("androidx.test:core:1.7.0")
     debugImplementation("androidx.fragment:fragment-testing:1.6.2")
     debugImplementation("com.squareup.leakcanary:leakcanary-android:2.14")
--- a/platform/jvm/gradle-test-app/build.gradle.kts
+++ b/platform/jvm/gradle-test-app/build.gradle.kts
@@ -77,6 +77,10 @@ dependencies {
     implementation("com.squareup.retrofit2:converter-gson:3.0.0")
 
     debugImplementation("androidx.compose.ui:ui-test-manifest")
+    // ui-test-manifest drags androidx.test:core 1.5.0 onto debugRuntimeClasspath, and AGP's
+    // consistent resolution then pins androidTest to it. Espresso <3.7 calls
+    // InputManager.getInstance(), removed in Android 17, so the floor has to come up here too.
+    debugImplementation("androidx.test:core:1.7.0")
     debugImplementation("androidx.fragment:fragment-testing:1.6.2")
     debugImplementation("com.squareup.leakcanary:leakcanary-android:2.14")
     debugImplementation("com.github.chuckerteam.chucker:library:4.2.0")
@@ -87,14 +91,14 @@ dependencies {
     androidTestImplementation(project(":common"))
     androidTestImplementation("com.google.truth:truth:1.1.4")
     androidTestImplementation("androidx.compose.ui:ui-test-junit4")
-    androidTestImplementation("androidx.test:core:1.5.0")
-    androidTestImplementation("androidx.test:core-ktx:1.5.0")
-    androidTestImplementation("androidx.test.ext:junit:1.1.5")
-    androidTestImplementation("androidx.test.ext:junit-ktx:1.1.5")
-    androidTestImplementation("androidx.test:runner:1.5.0")
+    androidTestImplementation("androidx.test:core:1.7.0")
+    androidTestImplementation("androidx.test:core-ktx:1.7.0")
+    androidTestImplementation("androidx.test.ext:junit:1.3.0")
+    androidTestImplementation("androidx.test.ext:junit-ktx:1.3.0")
+    androidTestImplementation("androidx.test:runner:1.7.0")
     androidTestImplementation("androidx.tracing:tracing-ktx:1.0.0")
     androidTestImplementation("androidx.tracing:tracing:1.0.0")
-    androidTestImplementation("androidx.test.espresso:espresso-core:3.5.1")
+    androidTestImplementation("androidx.test.espresso:espresso-core:3.7.0")
 }
 
 android {
--- a/platform/jvm/gradle-test-app/build.gradle.kts
+++ b/platform/jvm/gradle-test-app/build.gradle.kts
@@ -139,6 +139,7 @@ android {
     }
     lint {
         checkDependencies = true
+        checkTestSources = true
         disable.add("GradleDependency")
         disable.add("AndroidGradlePluginVersion")
     }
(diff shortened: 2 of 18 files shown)


.github/workflows/android.yaml
```diff
@@ -322,6 +322,9 @@ jobs:
     - name: Gradle capture-plugin unit tests
       working-directory: ./platform/jvm
       run: ./gradlew capture-plugin:test --info
+    - name: Gradle replay unit tests
+      working-directory: ./platform/jvm
+      run: ./gradlew replay:testDebugUnitTest --info
```
murki: we were not running the unit tests inside the replay module in CI

platform/jvm/gradle-test-app/build.gradle.kts
```diff
@@ -136,6 +139,7 @@ android {
     }
     lint {
         checkDependencies = true
+        checkTestSources = true
```
murki, line 142: make sure to run the list rules inside the test modules

platform/jvm/replay/src/main/kotlin/io/bitdrift/capture/replay/internal/mappers/ViewMapper.kt
```diff
@@ -76,15 +77,16 @@ internal class ViewMapper(
                 "invalid_resource_id"
             }
 
-        val type = viewMapperConfiguration.mapper[this.javaClass.simpleName]
+        val className = this.javaClass.simpleName
```
murki, line 79: unrelated, but we reuse the computed class names below

platform/jvm/replay/src/main/kotlin/io/bitdrift/capture/replay/internal/compose/ComposeTreeParser.kt
```diff
@@ -162,6 +169,46 @@ internal object ComposeTreeParser {
...
+        }
+
+    /**
+     * Whether this node draws a surface of its own, which an overlay's scrim and wrappers do not.
+     * Matched by key name because [SemanticsProperties] gained `Shape` after the Compose version
+     * this module compiles against; on older runtimes it is simply absent.
+     */
+    private val SemanticsConfiguration.paintsSurface: Boolean
+        // TODO(@murki): Migrate to using strongly-typed [SemanticsProperties.Shape] check when updating compose-ui to >= 1.9.0
+        get() = any { it.key.name == SHAPE_KEY }
```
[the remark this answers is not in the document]
murki: This is ok since we want to stay in compose:ui 1.7.6 for the time being, so this is fwd compatible but doesn't break older versions

platform/jvm/replay/src/main/kotlin/io/bitdrift/capture/replay/internal/mappers/BackgroundMapper.kt
```diff
@@ -7,30 +7,16 @@
...
-                    }
-                    else -> {
-                        // is Transparent or unknown
-                        ReplayType.Ignore
-                    }
-                }
+        // Runs alongside ButtonMapper and TextMapper, so only contribute a rect when the
+        // background actually paints.
+        BackgroundOpacity.paintedType(view)?.let { type ->
             list.add(ReplayRect(type, viewOriginX, viewOriginY, view.width, view.height))
```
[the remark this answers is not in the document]
murki: fixed

platform/jvm/replay/src/main/kotlin/io/bitdrift/capture/replay/internal/compose/ComposeTreeParser.kt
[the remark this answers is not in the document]
murki: we could also decide to bump our min compose version to 1.10.0 (released Dec 2025) but customers that are not yet would complain https://developer.android.com/jetpack/androidx/releases/compose-ui#1.10.0
wdyt @FranAguilera ?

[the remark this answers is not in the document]
FranAguilera: let's discuss internally

platform/jvm/replay/src/main/kotlin/io/bitdrift/capture/replay/internal/compose/ComposeTreeParser.kt
```diff
@@ -162,6 +169,46 @@ internal object ComposeTreeParser {
...
+
+    /**
+     * Downgrades a generic container inside an overlay window. Compose semantics carry no background
+     * information, and guessing opaque lets a full-screen container occlude the windows beneath it.
+     */
+    private fun ReplayType.occlusionAdjustedFor(
+        window: ComposeWindow,
+        config: SemanticsConfiguration,
+    ): ReplayType =
+        if (window.isOverlay && this == ReplayType.View && !config.paintsSurface) {
```
FranAguilera, line 190: Maybe I missed somewher but how do we know `paintsSurface` is accurate here? If it is wrong we may hide a dialog background in the replay.

FranAguilera (commented):
Initial comments

platform/jvm/replay/src/main/kotlin/io/bitdrift/capture/replay/internal/compose/ComposeTreeParser.kt
```diff
@@ -54,7 +56,14 @@ internal object ComposeTreeParser {
         SessionReplayController.L.d(
             "Found Compose SemanticsNode root. Parsing Compose tree. Window offset: (${windowOffset[0]}, ${windowOffset[1]})",
         )
-        return rootNode.toScannableView(windowOffset[0], windowOffset[1], layoutNodeMap)
+        return rootNode.toScannableView(
+            ComposeWindow(
+                offsetX = windowOffset[0],
+                offsetY = windowOffset[1],
+                layoutNodes = layoutNodeMap,
+                isOverlay = rootNode.hostsOverlayWindow(),
```
FranAguilera: This adds a second tree traversal on the main thread for every capture, wondering of the perf implications (did you profile this?)

murki: good catch, this is definitely not desirable

murki: fixed

platform/jvm/replay/src/main/kotlin/io/bitdrift/capture/replay/internal/compose/ComposeTreeParser.kt
murki: There are some comments at the top of `SemanticsConfiguration.paintsSurface`, but this is probably the same discussion as https://github.com/bitdriftlabs/capture-sdk/pull/1161#discussion_r3866786331 we can follow up there, 
**tl;dr:** using `Shape` is a proxy signal, not a direct answer

murki: to be clear this check is only needed to "remove" transparency for opaque backgrounds inside a modal, without this check the logic was too broadly marking all surfaces inside the modal windows transparent. I've updated the comments to include more clarity. See the examples below where the bottom sheet was "too" transparent before

|before|after|
|---|---|
|Screenshot 2026-08-26 at 11 02 59 AM|Screenshot 2026-08-26 at 3 53 44 PM|

platform/jvm/replay/src/main/kotlin/io/bitdrift/capture/replay/internal/compose/ComposeTreeParser.kt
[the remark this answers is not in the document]
FranAguilera: I wouldn't bump yet. Reason is to prevent any unexpected issues for customers still on old version

platform/jvm/replay/src/main/kotlin/io/bitdrift/capture/replay/internal/compose/ComposeTreeParser.kt
FranAguilera: I validated this with Compose 1.7.6: `Shape` is not present so `paintsSurface` is always false and the modal is rendered as transparent in replay. 

Screenshot 2026-09-02 at 16 51 57

FranAguilera: the layout bounds for modal still present. Discussed offline that this is a compromise given most folks will be >= 1.10

FranAguilera (approved):

PR #1162, opened 2026-08-26
[iOS] Fix flaky tests

ArielDemarco:
# Overview
This PR attempts to reduce some flakyness by:
- Delete `QueueTimer` and its tests as they were failing. The class is not being used.
- Created an UUID for background `URLSession` in integration tests to prevent work owned by `nsurlsessiond` from surviving a test host and leaking into the next run when CI reuses a simulator.
- Prevent cases where the exponential backoff would take long enough for the test to reach the timeout.
- Added some logging/diagnostics to the network tests to report if the Rust runtime started a stream/attempted to write data/shut it down to find out if there's another source of flakiness in test (and also to make it easier to solve).
- `URLSessionIntegrationTests` calls `customSetUp` inside a loop, adding a new `MockLogging` into a singleton on every iteration. `invalidateAndCancel` is async, so callbacks from an earlier iteration can still land while the next one runs, and the mock fulfilled any expectation for any log. This explains timeout, multiple fulfillment failures, etc. To fix it, now we tag each request URL with the test name and an `UUID` and have the mock record and fulfill only for logs carrying that tag.


test/platform/swift/bridging/src/lib.rs
```diff
@@ -47,6 +48,23 @@ extern "C" fn create_benchmarking_configuration(dir_path: *const c_char) {
...
 
+#[unsafe(no_mangle)]
+extern "C" fn create_fast_retry_configuration(dir_path: *const c_char) {
+  let c_path: &CStr = unsafe { CStr::from_ptr(dir_path) };
+  let str_path: &str = c_path.to_str().unwrap();
+
+  let config = make_update(
+    vec![
+      ("api.initial_backoff_interval_ms", ValueKind::Int(1)),
+      ("api.max_backoff_interval_ms", ValueKind::Int(10)),
```
snowp: There are constants you can use from bd-runtime if you want to avoid hard coding it, check out how it's done in shared-core tests

snowp (approved):

test/platform/swift/bridging/src/lib.rs
ArielDemarco: Good call! i'll check it out!

ArielDemarco: Done! thanks!

PR #1164, opened 2026-08-27
andr;build: move Bazel test runtime to JDK 17 and align Robolectric

murki:
aligning bazel and gradle:
1. run tests in bazel using JDK 17 (instead of 11)
2. bump robolectric version for this and align them

`ci/jdk_wrapper.sh` stays on JDK 8 deliberately; it runs the Android `sdkmanager` from cmdline-tools 6.0, which requires it.

---


.bazelrc
```diff
@@ -47,7 +47,7 @@ build:noclippy --output_groups=-clippy_checks
 # `bazel test //...` should discover and build tests, not every build target that the wildcard
 # happens to match. Targets that are not tests remain available through `bazel build`.
 test --build_tests_only
-test --test_env=RUST_BACKTRACE=full --java_runtime_version=remotejdk_11
+test --test_env=RUST_BACKTRACE=full --java_runtime_version=remotejdk_17
```
snowp: Can we just remove this? Would 17 be the default?

snowp (approved):

.bazelrc
murki: i think so, easy to validate in CI

PR #1173, opened 2026-09-01
[Android] [gradle plugin] Support BITDRIFT_API_KEY (follow ups)

FranAguilera:
Follow up to https://github.com/bitdriftlabs/capture-sdk/pull/1169

Resolves BIT-9723, BIT-7392

- Make sure the env var api key check also applies to other commands like `bdUploadSourceMap` and `bdUploadSymbols`
- Ensure the expected latest version is installed
- Run related capture-plugin tests on CI
- Fix flaky test 

## Verification
- JUnit tests
- Manual verification via maven local publish (downloaded old cli version, passing empty api key, etc)

---

### DIFF
--- a/platform/jvm/capture-plugin/src/main/kotlin/io/bitdrift/capture/task/CLITask.kt
+++ b/platform/jvm/capture-plugin/src/main/kotlin/io/bitdrift/capture/task/CLITask.kt
@@ -8,6 +8,7 @@
 package io.bitdrift.capture.task
 
 import org.gradle.api.DefaultTask
+import org.gradle.api.GradleException
 import org.gradle.api.file.DirectoryProperty
 import org.gradle.api.provider.Property
 import org.gradle.api.tasks.Internal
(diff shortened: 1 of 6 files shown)


platform/jvm/capture-plugin/src/main/kotlin/io/bitdrift/capture/task/CLITask.kt
```diff
@@ -131,14 +131,40 @@ abstract class CLITask : DefaultTask() {
...
+        private const val DEPRECATED_API_KEY_ENV_VAR_NAME = "API_KEY"
+    }
 }
 
 class BDCLIDownloader(
     val executableFilePath: File,
+    private val download: (URI) -> ByteArray = { it.toURL().readBytes() },
 ) {
     val bdcliVersion = "0.2.23"
     val bdcliDownloadLoc: URI = URI.create("https://dl.bitdrift.io/bd-cli/$bdcliVersion/${downloadFilename()}/bd")
```
murki, line 170: the url also supports pointing to latest via e.g. https://dl.bitdrift.io/bd-cli/latest/bd-cli-mac-arm64.tar.gz/bd in case you think it's better to always attempt to use the newest version vs pinning, since we don't seem to update this script very often

murki (approved):

platform/jvm/capture-plugin/src/main/kotlin/io/bitdrift/capture/task/CLITask.kt
FranAguilera: I'd rather keep it as is to prevent outages if a newly published CLI version has issues.

PR #1174, opened 2026-09-01
ios: migrate Swift bridge Foundation usage to objc2

snowp:
Moves the Swift bridge’s focused Foundation paths to objc2-foundation: UserDefaults storage, NSString/NSData conversion helpers, and generic Foundation values. The compatibility C ABI and lossy malformed-NSString handling remain intact, while retaining the Arc-backed string API now present on main.


snowp:
objc2 seems problematic while we provide a static lib; closing

PR #1175, opened 2026-09-01
Avoid default log metadata callbacks

snowp:
Removes default platform metadata callbacks from plain logging. Android and iOS now pass only explicit custom timestamp and custom-field providers; absent providers use native time and empty fields, avoiding JNI/Objective-C round trips in the default path.


platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/providers/MetadataProvider.kt
```diff
@@ -9,16 +9,20 @@ package io.bitdrift.capture.providers
 
 import android.util.Log
 import io.bitdrift.capture.ErrorHandler
-import io.bitdrift.capture.IMetadataProvider
+import io.bitdrift.capture.ICustomFieldsProvider
+import io.bitdrift.capture.ITimestampProvider
 
-internal class MetadataProvider(
+internal class TimestampProvider(
```
FranAguilera: Can we rename the file accordingly?

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/LoggerImpl.kt
```diff
@@ -89,7 +91,7 @@ internal class LoggerImpl(
     configuration: Configuration,
     customFieldGetters: List<FieldGetter>,
     initialFields: Fields = emptyMap(),
-    dateProvider: DateProvider,
+    dateProvider: DateProvider?,
```
FranAguilera, line 94: nit: you could minimize changes if kept as non-null as pass the default one in Capture.kt

FranAguilera (approved):
Nice perf improvements. Just pre-approved with just a few nits to be addressed before merging

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/LoggerImpl.kt
snowp: It seemed cleaner to me to just propagate null here instead of having to do a instance check on SystemTimeProvider to map it back to null for the JNI case, since the goal here is to intentionally avoid using SystemTimeProvider to elide the JNI call during evaluation 

SystemTimeProvider is almost ready to be removed at this point but I didn't want to conflate updating the crash reporting code to also do the null check with this PR

PR #996, opened 2026-05-21
[Flutter] Alpha Support - Release 0.0.1

FranAguilera:
## Goal

Initial alpha flutter integration prototype. Plugin at `platform/capture_flutter/`, example app at `examples/flutter/` (pointing to local path dep for now).

Sample initialization in dart

```
 final success = await Capture.start( apiKey: _apiKey,apiUrl: _apiUrl,);
```

Adding a a local dep in `examples/flutter/pubspec.yaml`

```
dependencies:
  capture_flutter:
    path: ../../platform/capture_flutter 
```

For any customer interested to test this they can use this tag `flutter-prototype-0.0.1`

```
dependencies:
  capture_flutter:
    git:
      url: https://github.com/bitdriftlabs/capture-sdk.git
      ref: flutter-prototype-0.0.1
      path: platform/capture_flutter
```

## Verification

cd examples/flutter
flutter run

[Sample session here](https://timeline.bitdrift.dev/session/1c4eea0e-b6c6-4aca-8f80-65edd0f9690c?utm_source=sdk&utilization=0&expanded=1679665124000542747)

## Alpha releases

Flutter alpha releases are Git tags only.

**_NOTE: The initial alpha release was published manually with the core changes here as `flutter-prototype-0.0.1` tag but next release will follow this process below_**

For a new release:
1. Open a new PR and bump `platform/capture_flutter/pubspec.yaml`, update the README tag, and move reviewed notes from `platform/capture_flutter/ALPHA_RELEASES.md`'s `Next release` section into the new version entry.
2. Update the native Android and iOS Capture SDK dependencies when required.
3. Merge to `main`.
4. The `Publish Flutter Alpha Release Tag` action will run automatically creating the tag `flutter-prototype-<version>`.

## Basic CI check

- Also added basic check to run for each flutter PR change https://github.com/bitdriftlabs/capture-sdk/actions/runs/33615749345/job/100201048708?pr=996

Resolves BIT-9835, BIT-8641, BIT-8587, BIT-8586, BIT-8584

---

- [ ] `CHANGELOG.md`'s "Unreleased" section has been updated, if applicable.

PR #1186, opened 2026-09-03
Fix Flutter Android build: apply Kotlin Gradle plugin

iainfinlayson:
## Goal

`capture_flutter`'s Android module and the in-repo `examples/flutter` app both fail to build on `main`. Found while validating the Flutter alpha demo in `sa-public` off the merged #996.

## Root cause

Both `platform/capture_flutter/android/build.gradle.kts` and `examples/flutter/android/app/build.gradle.kts` use a `kotlin { compilerOptions { jvmTarget = ... } }` block, but neither declares `org.jetbrains.kotlin.android` in its own `plugins {}` block. Gradle only generates the typed `kotlin {}` DSL accessor for plugins applied within that same script's `plugins {}` block — not for ones applied transitively (e.g. by the Flutter Gradle plugin) — so both scripts fail to compile:

```
e: .../build.gradle.kts: None of the following candidates is applicable:
    fun DependencyHandler.kotlin(module: String, version: String? = ...): Any
    fun PluginDependenciesSpec.kotlin(module: String): PluginDependencySpec
e: .../build.gradle.kts: Unresolved reference 'compilerOptions'.
e: .../build.gradle.kts: Unresolved reference 'jvmTarget'.
```

Because `capture_flutter`'s own Kotlin sources (`CaptureFlutterPlugin.kt`) never compile as a result, **any Android app that depends on `capture_flutter` fails to build**, hitting:

```
cannot find symbol: class CaptureFlutterPlugin
```

## Fix

- Add `id("org.jetbrains.kotlin.android")` to the `plugins {}` block in both `platform/capture_flutter/android/build.gradle.kts` and `examples/flutter/android/app/build.gradle.kts`.
- Pin a Kotlin Gradle plugin version (`2.3.20`, matching `platform/jvm`) in `examples/flutter/android/settings.gradle.kts` — neither module declared one, and without a pin Gradle falls back to a default older than Flutter's minimum supported Kotlin version (1.8.10).

## Verification

```
cd examples/flutter && flutter build apk --release
```

- Before this branch (on `main`): fails with the errors above.
- After: `✓ Built build/app/outputs/flutter-apk/app-release.apk`.

Also confirmed the in-repo example app runs end-to-end on an arm64 emulator (SDK initializes, authenticates, welcome screen renders) once pointed at a valid API key.

---

- [ ] `CHANGELOG.md`'s "Unreleased" section has been updated, if applicable. (Skipped — build/tooling-only fix, no SDK behavior change, per repo convention.)

### DIFF
--- a/platform/capture_flutter/ALPHA_RELEASES.md
+++ b/platform/capture_flutter/ALPHA_RELEASES.md
@@ -23,6 +23,23 @@ TODO: Before publishing, update the native SDK dependencies in `android/build.gr
 
 - Nothing yet!
 
+## [0.0.3]
+[0.0.3]: https://github.com/bitdriftlabs/capture-sdk/tree/flutter-prototype-0.0.3
+
+### Both
+
+**Added**
+
+- Nothing yet!
+
+**Changed**
+
+- Nothing yet!
+
+**Fixed**
+
+- Fixed Android builds for apps using `capture_flutter` by applying the Kotlin Android Gradle plugin.
+
 ## [0.0.2]
 [0.0.2]: https://github.com/bitdriftlabs/capture-sdk/tree/flutter-prototype-0.0.2
 
--- a/examples/flutter/android/app/build.gradle.kts
+++ b/examples/flutter/android/app/build.gradle.kts
@@ -2,6 +2,7 @@ import org.jetbrains.kotlin.gradle.dsl.JvmTarget
 
 plugins {
     id("com.android.application")
+    id("org.jetbrains.kotlin.android")
     id("dev.flutter.flutter-gradle-plugin")
 }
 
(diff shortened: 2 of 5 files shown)


iainfinlayson:

FranAguilera:
Thanks for the fix @iainfinlayson. Will update ALPHA_RELEASE.md in flutter folder so a new tag is published after release

platform/capture_flutter/ALPHA_RELEASES.md
```diff
@@ -23,6 +23,23 @@ TODO: Before publishing, update the native SDK dependencies in `android/build.gr
 
 - Nothing yet!
 
+## [0.0.3]
+[0.0.3]: https://github.com/bitdriftlabs/capture-sdk/tree/flutter-prototype-0.0.3
```
FranAguilera, line 27: @iainfinlayson This is how we do release tag based, after merging this the tag will be published automatically via this job https://github.com/bitdriftlabs/capture-sdk/actions/workflows/publish_flutter_alpha.yaml

FranAguilera (approved):

PR #1196, opened 2026-09-08
[Android] [WebView] Open up manual webview instrumentation + modes for auto instrumentation

FranAguilera:
# Goal

Resolves BIT-9149

Plugin: 
- Provides a safer "auto" mode that will automatically instrument WebView that have enabled webview.settings.javaScriptEnabled already
- Deprecates existing `automaticWebViewInstrumentation` in favor of `webViewAutomaticInstrumentationMode`
- The 2 modes will be `JS_ENABLED` and `ALL` (full enables javascript automatically)

Capture-sdk:
- Open visibility of WebViewCapture.instrument(webView). When this method is called, if `javaScriptEnabled` is set to false it will enable it internally to be able to do full instrumentation

# Verification
- JUnit tests
- Update gradle-test-app
- Use gradle-test-app with the local published pluglign for e2e verification

Screenshot 2026-09-09 at 16 34 08

---


platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/webview/WebViewCapture.kt
```diff
@@ -26,54 +26,62 @@ import io.bitdrift.capture.experimental.ExperimentalBitdriftApi
 import io.bitdrift.capture.providers.ArrayFields
 import io.bitdrift.capture.providers.fieldsOf
 
+/** Describes how a specific WebView was instrumented for internal telemetry. */
+internal enum class WebViewInstrumentationMode {
+    AUTOMATIC_FULL,
```
ArielDemarco: Maybe, `AUTOMATIC_ALWAYS`?

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/webview/WebViewCapture.kt
```diff
@@ -102,22 +110,39 @@ internal object WebViewCapture {
...
 
         webview.markAsInstrumented()
     }
 
+    private fun shouldForceEnableJavaScript(
+        instrumentationMode: WebViewInstrumentationMode,
+        webview: WebView,
+    ): Boolean =
+        instrumentationMode != WebViewInstrumentationMode.AUTOMATIC_JAVASCRIPT_ENABLED_ONLY &&
+            !webview.settings.javaScriptEnabled
```
ArielDemarco: This means `MANUAL` forces enabling js isnt it? I think that's fine, but probably it's worth adding it to the docs

FranAguilera: Yup that's the intention. Will clarify docs to reflect this behavior

ArielDemarco: Awesome!

platform/jvm/capture-plugin/src/main/kotlin/io/bitdrift/capture/extension/InstrumentationExtension.kt
```diff
@@ -24,20 +24,32 @@ open class InstrumentationExtension
...
-         * [io.bitdrift.capture.webview.WebViewConfiguration] must be provided when calling
-         * `Capture.Logger.start()`. If `null` is passed, the injected calls will be no-op.
-         *
-         * **Experimental:** This API may change in future releases.
+         * @deprecated Use [automaticWebViewInstrumentationMode] with
+         * [WebViewAutomaticInstrumentationMode.JS_ENABLED_ONLY].
          */
+        @Deprecated(
+            message = "Use automaticWebViewInstrumentationMode = JS_ENABLED_ONLY.",
+            replaceWith = ReplaceWith("automaticWebViewInstrumentationMode.set(JS_ENABLED_ONLY)"),
```
murki: while we're here, are we changing the behavior of this boolean to make the true behave as JS_ENABLED_ONLY? if not I think that's the safest thing to do here, if they want the old behavior they are now forced to use `utomaticWebViewInstrumentationMode = FULL`

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/webview/WebViewCapture.kt
```diff
@@ -173,3 +206,10 @@ internal object WebViewCapture {
         }
     }
 }
+
+/** Describes how a specific WebView was instrumented for internal telemetry. */
+internal enum class WebViewInstrumentationMode {
+    AUTOMATIC_ALWAYS,
+    AUTOMATIC_JAVASCRIPT_ENABLED_ONLY,
```
murki, line 209: can we also rename these to match the FULL etc names in the plugin?

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/webview/WebViewCapture.kt
```diff
@@ -102,22 +104,39 @@ internal object WebViewCapture {
...
 
         webview.markAsInstrumented()
     }
 
+    private fun shouldForceEnableJavaScript(
+        instrumentationMode: WebViewInstrumentationMode,
+        webview: WebView,
+    ): Boolean =
+        instrumentationMode != WebViewInstrumentationMode.AUTOMATIC_JAVASCRIPT_ENABLED_ONLY &&
+            !webview.settings.javaScriptEnabled
```
murki: why do we need to check again for `!webview.settings.javaScriptEnabled` ? should it just be `instrumentationMode != WebViewInstrumentationMode.AUTOMATIC_JAVASCRIPT_ENABLED_ONLY`? also should we inline this method above when it's used? since it's just used once

FranAguilera: It was just to avoid setting it again if the underlying view already supports it . Yup inlined already (I extracted a method while iterating as I was reusing in other place but now it's no longer need)

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/webview/WebViewCapture.kt
```diff
@@ -173,3 +202,10 @@ internal object WebViewCapture {
         }
     }
 }
+
+/** Describes how a specific WebView was instrumented for internal telemetry. */
+internal enum class WebViewInstrumentationMode {
+    AUTOMATIC_ALWAYS,
```
murki, line 208: should this one follow the plugin enum name with full?

murki (approved):

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/webview/WebViewCapture.kt
FranAguilera: Uops good catch, forgot to update. Will do before merging alongside some gradle-test-app cleanup

FranAguilera: actually, i'd rather keep it like this as this is used to report the used mode from a customer and will be a bit confusing to the reader later given here we also have MANUAL

PR #1201, opened 2026-09-10
[Android] [Tooling] Extract test module + add test for loggerImpl JVM only

FranAguilera:
- extracted a test module with shared fakes
- Add a new test for the JVM implementation of LoggerImpl

---


FranAguilera:
low pri, flaky test are resolved since shared-core bump with fixes. We can retake this when we find a need again

PR #1208, opened 2026-09-15
[Android] Add Pre-Init memory buffer

FranAguilera:
# Goal

Resolves BIT-9500

Adding a PreInit in memory buffer for the logger calls that occur while the SDK is starting internally.

This is the current flow in prod

Screenshot 2026-09-16 at 17 44 25

And here with the proposal

Screenshot 2026-09-16 at 17 44 00

The implementation does:

- PreInitInMemoryLogger is added only when Capture state is LoggerState.Starting so all Capture.Logger.* calls are forwarded there
- The instance is removed once we reach the Started/StartFailure state
- Timestamps for each log are respected
- To guard against potential OOM if starts is stuck, when the in memory cached calls reached a limit of 512 KB the new events will be dropped and an internal log will be emitted

# Verification

Steps:
- In the gradle-test-app, go to setting and enable, `Start with simulated Delay` and `Start SDK on Background Thread`
Screenshot 2026-09-17 at 15 33 28

- Restart the app, go to SDK tab and trigger many calls (e.g. `Log Many messages`) etc
Screenshot 2026-09-17 at 15 34 27

- The internal initialization priorly would have miss those logs but now they are added into the buffer and replaying later. See [session](https://timeline.bitdrift.dev/session/3cca5d01-1899-493b-b7e6-9b4949fd5625?utm_source=sdk&utilization=0#timeline) 

 

---


platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/Capture.kt
```diff
@@ -388,15 +393,15 @@ object Capture {
...
+            get() = logger()?.takeUnless { it is PreInitInMemoryLogger }?.sessionId
 
         /**
          * The URL for the current ongoing session.
          * It's equal to `null` prior to the start of Capture SDK.
          */
         @JvmStatic
         val sessionUrl: String?
-            get() = logger()?.sessionUrl
+            get() = logger()?.takeUnless { it is PreInitInMemoryLogger }?.sessionUrl
```
FranAguilera: ignores the pre init buffer value as we are not initialized. Also choose to not break the non-nullable contract on the ILogger api

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/events/span/Span.kt
```diff
@@ -75,6 +75,20 @@ class Span internal constructor(
         ) { "" }
     }
 
+    /**
+     * Attaches the real logger to a span created before one existed yet (i.e. while the SDK was
+     * still starting), emitting its deferred start log and letting a subsequent end() call work.
+     */
+    internal fun attachLogger(logger: IInternalLogger) {
```
snowp, line 82: Would we want some kind of defensive programming here to avoid duplicate emission? Like a `deferred` flag or something that tells us whether we need to do this? The coordination here seems pretty loose

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/Capture.kt
snowp: I think this means that we'll break anyone doing
```
Capture.start()
Capture.sessionUrl() 
```
yeah?

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/Capture.kt
```diff
@@ -440,9 +445,9 @@ object Capture {
          */
         @JvmStatic
         fun createTemporaryDeviceCode(completion: (CaptureResult<String>) -> Unit) {
-            logger()?.also {
-                it.createTemporaryDeviceCode {
-                    mainThreadHandler.run { completion(it) }
+            logger()?.takeUnless { it is PreInitInMemoryLogger }?.also { logger ->
```
snowp: should we just have something like `initalizedLogger()` to avoid this predicate check in multiple places?

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/Capture.kt
```diff
@@ -916,6 +922,10 @@ object Capture {
                     )
                 }
 
+            val flushToNativeDuration =
+                measureTime {
+                    preInitInMemoryLogger.flushToNative(loggerImpl)
+                }
             default.set(LoggerState.Started(loggerImpl))
```
snowp, line 955: I think this is technically racy as we could get inbound logs after flushToNative ends and default.set is called? I think we'd need a lock or other work to make this better if we care about this narrow log drop chance

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/Capture.kt
```diff
@@ -937,6 +948,7 @@ object Capture {
 
             startResult.invokeCatchingOrThrowOnDebug(CaptureResult.Success(loggerImpl))
         } catch (throwable: Throwable) {
+            preInitInMemoryLogger.clear()
```
snowp, line 977: Is this necessary? won't it be deallocated as part of updating default? Seems fine but not sure if we need it 

If something can retain a handle to it then I imagine we'd want to make it drop inbounds as well if we wanted to avoid it doing all the log insertion and accounting

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/PreInitInMemoryLogger.kt
```diff
@@ -0,0 +1,357 @@
...
+        key: String,
+        value: String,
+    ) = add(BufferedCall.AddField(key, value))
+
+    override fun removeField(key: String) = add(BufferedCall.RemoveField(key))
+
+    override fun setFeatureFlagExposure(
+        name: String,
+        variant: String,
+    ) = add(BufferedCall.StringFeatureFlagExposure(name, variant))
```
snowp, line 68: technically this ought to also capture the timestamp but maybe not worth the complexity to add

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/PreInitInMemoryLogger.kt
```diff
@@ -0,0 +1,357 @@
...
+            // Attaches the real logger to the caller's own Span, instead of starting a new one.
+            override fun replay(logger: IInternalLogger) = span.attachLogger(logger)
+        }
+
+        data class HttpRequest(
+            val request: HttpRequestInfo,
+        ) : BufferedCall {
+            override val sizeBytes = sized(request.toString())
+
+            override fun replay(logger: IInternalLogger) = logger.log(request)
```
snowp, line 307: wouldn't this also need a timestamp override? or is that captured in the request info? same for the other convenience functions

snowp (commented):
Mostly makes sense to me, a few points and a few possible inconsistencies I spotted. We should decide whether we are okay with where this approach will get things slightly wrong in some cases or if we need to put in more effort for some of the edge cases

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/Capture.kt
FranAguilera: i think most folks may be checking nullability to infer valid session url, so here there is no much change from that logic to keep this to be non-null until sdk is fully started

One thing we could do is to mark those (Capture.Logger.sessionUrl, Capture.Logger.sessionId, etc) as deprecated and direct people from the deprecation note to rely on the existing StartResult callback that returns an object with the non-nullable sessionId,url, etc when there is a successful start

Screenshot 2026-09-18 at 17 04 26

wdyt?

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/Capture.kt
FranAguilera: good call, will add it

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/Capture.kt
FranAguilera: good point, will clean up

FranAguilera:
Moving to draft until all feedback is addressed

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/Capture.kt
FranAguilera: Yeah I was concerned about that and that's why we have another atomic ref within PreInit `drainTarget` which controls dispatch within PreInit.adds internals and that is independent of the default atomic ref from Capture.kt

```
fun flushToNative(logger: IInternalLogger) {
    drainTarget.set(logger)
    drainTo(logger)
}
```

And the internal add

```
private fun add(call: BufferedCall) {
    drainTarget.get()?.let {
        call.dispatch(it)
        return
    }
```

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/Capture.kt
FranAguilera: clear could still be useful for cases when someone holds a reference to the instance (e.g. a span api start call but never calling end manually

good idea on also dropping inbounds after that point too. will add

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/Capture.kt
snowp: Great yeah I think this address the log drop. There's still a chance of operation re-ordering for operations that come in during the drain period since they'd be interleaved with the other logs (workflow/processing order, not timeline sorted by time order). I think we could use a small lock to coordinate the draining of the buffer so that we continue to push entries into the queue until it's empty while avoiding locking in the regular hot-path. We can talk more in person (or you can suggest this to the LLM) if that's helpful 

The main thing I want to avoid is having any chance of

```
addField("a", "1")
addField("a", "2");
```

to result in a=1 being set for the duration of the session because it lines up with pre config drain in such a way that the second call is processed before the other.

platform/jvm/capture/src/test/kotlin/io/bitdrift/capture/PreInitInMemoryLoggerTest.kt
```diff
@@ -0,0 +1,336 @@
...
+            HttpResponseInfo(
+                request = request,
+                response = HttpResponse(result = HttpResponse.HttpResult.SUCCESS, statusCode = 200),
+                durationMs = 42L,
+            )
+
+        val beforeMs = System.currentTimeMillis()
+        preInitInMemoryLogger.log(request)
+        preInitInMemoryLogger.log(response)
+        val afterMs = System.currentTimeMillis()
```
snowp: This seems a bit weak to me, wont beforeMs and afterMs end up in the same ms in many cases? Probably fine just seems a bit brittle

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/PreInitInMemoryLogger.kt
snowp: This still seems open? Seems like ideally we should be passing a timestamp override for everything

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/Capture.kt
```diff
@@ -75,10 +77,17 @@ object Capture {
...
      */
-    fun logger(): ILogger? =
+    fun logger(): ILogger? = (default.get() as? LoggerState.Started)?.logger
+
+    /**
+     * Similar to [Capture.logger], but also returns the pre-init buffer while Capture is starting.
+     * This is required for internal SDK calls that shouldn't be dropped. E.g. Timber, network calls,
+     * etc
+     */
+    internal fun getInternalLogger(): IInternalLogger? =
```
snowp: Why wouldn't we expose this to all callers? Seems like if someone is calling logger() to get the handle they won't benefit?

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/Capture.kt
```diff
@@ -385,6 +397,10 @@ object Capture {
         /**
          * The Id for the current ongoing session.
          * It's equal to `null` prior to the start of Capture SDK.
+         *
+         * Alternatively, use the ILogger from start's startResult callback for a non-null
```
snowp: Isn't there a kdocs/javadocs way to reference the function here? It's a bit hard to parse that `start` is a particular function name

platform/jvm/capture/src/test/kotlin/io/bitdrift/capture/PreInitInMemoryLoggerTest.kt
snowp, line 1: could probably add some more tests here on the handoff to show that logs are correctly handed off

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/Capture.kt
FranAguilera: will update

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/PreInitInMemoryLogger.kt
FranAguilera: sorry forgot to comment, I've added an internal timestamp here to capture the call time for Httprequest

`override fun log(httpRequestInfo: HttpRequestInfo) = add(BufferedCall.HttpRequest(httpRequestInfo, getTimeStampInMs()))`

And the defintion

```
      data class HttpRequest(
          val request: HttpRequestInfo,
          val occurredAtMs: Long,
      ) : BufferedCall {
          override val sizeBytes = sized(request.toString())

          override fun dispatch(logger: IInternalLogger) {
              logger.logInternal(
                  type = LogType.SPAN,
                  level = LogLevel.DEBUG,
                  arrayFields = request.arrayFields,
                  matchingArrayFields = request.matchingArrayFields,
                  attributesOverrides = LogAttributesOverrides.OccurredAt(occurredAtMs),
              ) { request.name }
          }
      }
```

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/Capture.kt
FranAguilera: iirc there was a concern from @murki from a customer using the first non-null Capture.logger instance for the whole lifetime of the application on custom setup

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/Capture.kt
FranAguilera: Yup, will link to that

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/Capture.kt
murki: I think what @snowp  is saying is that `Capture.Logger() `should return the pre-init buffer as well? (which I agree), not that we should expose `IInternalLogger` publicly

murki: but it does feel scary that now we can make it possible for someone to hold a reference to Capture.logger() : PreInitInMemoryLogger and then they will never swap to the "real" logger

platform/jvm/capture-timber/src/main/kotlin/io/bitdrift/capture/timber/CaptureTree.kt
```diff
@@ -24,11 +26,11 @@ import timber.log.Timber
 open class CaptureTree internal constructor(
     private val internalLogger: ILogger?,
 ) : Timber.Tree() {
-    constructor() : this(Capture.logger())
+    constructor() : this(Capture.getInternalLogger())
```
murki: couldn't this cache a instance of `PreInitInMemoryLogger` if called early enough and never switch to the real logger?

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/network/okhttp/CaptureOkHttpEventListenerFactory.kt
```diff
@@ -60,7 +60,7 @@ class CaptureOkHttpEventListenerFactory internal constructor(
         responseFieldProvider: OkHttpResponseFieldProvider = DEFAULT_RESPONSE_FIELD_PROVIDER,
     ) : this(
         targetEventListenerFactory = targetEventListenerFactory,
-        logger = Capture.logger(),
+        logger = Capture.getInternalLogger(),
```
murki: couldn't this cache a instance of `PreInitInMemoryLogger` if called early enough and never switch to the real logger?

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/Capture.kt
murki: I think this a concern even in the current internal usage of `Capture. getInternalLogger()` ⚠️

FranAguilera: reverted back to previous Capture.logger returning also preinit while in starting state.

imho, we should deprecate Capture.logger() and provide testing utils if needed. Also from a while we've been exposing the ILogger valid instance from startResult callback so we can recommend customers to use that instead

platform/jvm/capture-timber/src/main/kotlin/io/bitdrift/capture/timber/CaptureTree.kt
FranAguilera: update to initial iteration, this is no longer cached

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/Capture.kt
FranAguilera: verified here with latest changes https://timeline.bitdrift.dev/session/5e1520a2-5290-4532-b48c-fba64bb12618?utm_source=sdk&utilization=0

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/PreInitInMemoryLogger.kt
```diff
@@ -0,0 +1,539 @@
...
+    private val failed = AtomicBoolean(false)
+    private val drainLock = Any()
+
+    // Preserve the non-null ILogger contract while initSdk is still in progress. The outer
+    // Capture.Logger getters hide these placeholders by returning null for PreInitInMemoryLogger,
+    // so we keep consumers of ILogger expecting non-null values
+    override val sessionId: String = UNKNOWN_VALUE
+    override val sessionUrl: String = UNKNOWN_VALUE
+    override val deviceId: String = UNKNOWN_VALUE
+    override val isTracingActive: Boolean = false
```
snowp: My agent flagged that with the current model of allowing consumers to hold on to the `PreInitInMemoryLogger` these would never start returning the real value, so we should consider having these also check `drainTarget` and forward to the real logger after startup so that callers eventually see the real value.

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/PreInitInMemoryLogger.kt
```diff
@@ -0,0 +1,539 @@
...
+    private val bufferedCalls = ConcurrentLinkedQueue<BufferedCall>()
+    private val bufferedBytes = AtomicInteger(0)
+    private val droppedCallCount = AtomicInteger(0)
+    private val drainTarget = AtomicReference<IInternalLogger?>()
+    private val failed = AtomicBoolean(false)
+    private val drainLock = Any()
+
+    // Preserve the non-null ILogger contract while initSdk is still in progress. The outer
+    // Capture.Logger getters hide these placeholders by returning null for PreInitInMemoryLogger,
+    // so we keep consumers of ILogger expecting non-null values
```
snowp: I think with our change to the way Capture.logger() returns the pre-init this comment is a bit misleading?

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/network/okhttp/CaptureOkHttpEventListenerFactory.kt
snowp: My agent flagged this: now that we're allowing logs to hit the PCB we'll also call this below 
```
    if (currentLogger == null || requestIgnorePolicy.shouldIgnore(call.request())) {
            return targetEventListener ?: EventListener.NONE
        }
        return CaptureOkHttpEventListener(
            logger = currentLogger,
            clock = clock,
            targetEventListener = targetEventListener,
            configuredPropagationMode = configuredPropagationMode,
            requestExtraFieldsProvider = requestFieldProvider,
            responseExtraFieldsProvider = responseFieldProvider,
        )
```
where we'd previously delay the logger. This means that we'd cache the default value of the runtime (w3c) and never re-evaluate this against the real runtime values. This might be an argument for skipping PCB integration for this call in particular since we can't guarantee the correct tracing config or at least we should avoid caching the default value for the duration of the `CaptureOkHttpEventListenerFactory`

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/PreInitInMemoryLogger.kt
```diff
@@ -0,0 +1,539 @@
...
+            return
+        }
+
+        bufferedCalls.add(call)
+
+        drainTarget.get()?.let(::drainTo)
+    }
+
+    private fun drainTo(logger: IInternalLogger) {
+        synchronized(drainLock) {
```
snowp, line 247: I think the way this needs to work is that we have to think about the PCB as having two states:

buffering (writes logs to buffer)
forwarding (drain is complete, logs flow)

This would use the https://en.wikipedia.org/wiki/Double-checked_locking pattern to check the state of the PCB to determine if a lock is required during writes:
- If we're already in forwarding, we skip the lock and proceed. 
- If we're not, we lock and check the state value again (as the state might have changed since we previously check)

The key here is that this allows the drainer to take the lock and grab all the pending logs. Only after dispatching the entire batch does it reacquire the lock to check for more calls. Once the drainer sees that the buffer is empty we can now transition the state to forwarding. Since the PCB and drainer coordinate with the lock at this point, the handoff now have a strong guarantee at the cost of taking a lock over a small critical region during startup and never again once the PCB finishes. 

The key thing that this ensures that all the logs in the PCB have been *processed* before we switch to forwarding, which prevents new calls from overtaking buffered calls. 

Sorry to be to particular here but I think getting log ordering as correct as possible is pretty key given what our SDK does

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/PreInitInMemoryLogger.kt
```diff
@@ -0,0 +1,539 @@
...
+        }
+
+        if (!tryReserveBytes(call.sizeBytes)) {
+            droppedCallCount.incrementAndGet()
+            return
+        }
+
+        bufferedCalls.add(call)
+
+        drainTarget.get()?.let(::drainTo)
```
snowp, line 243: This stuff would also go away with the double checked locking, as we move to a strong guaranteed ordering from a best effort that leaves us open for races

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/PreInitInMemoryLogger.kt
```diff
@@ -0,0 +1,539 @@
...
+    override fun updateOotbField(
+        key: String,
+        value: String,
+    ) = add(BufferedCall.UpdateOotbField(key, value))
+
+    override fun logInternalError(
+        throwable: Throwable?,
+        blocking: Boolean,
+        message: () -> String,
+    ) = add(BufferedCall.LogInternalError(throwable, blocking, message()))
```
snowp, line 177: Agent flagged this: seems like we're now evaluating message() without a try-catch like we have in the non-PCB flow

snowp (commented):
Took another pass, heavily assisted by my agent to validate some ideas. 

Let me know if you want to chat more in person about this stuff, I realize it's all a bit tricky. I had pretty good luck chatting to my agent about some of these options (using Astra Light), might help iterate on this faster

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/network/okhttp/CaptureOkHttpEventListenerFactory.kt
FranAguilera: good catch, will fix

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/PreInitInMemoryLogger.kt
FranAguilera: good point, added a DCL implementation [here](https://github.com/bitdriftlabs/capture-sdk/pull/1208/changes/9658a59f7568faa19f76352b90307f2f7b1ca321) (manually verified in worse case situations)

btw will polish tests soon

FranAguilera: actually just simplified it a bit more and removed the drainLock given `LoggerImpl.initSdk` will happen only once for the caller thread (guarded via Atomic state ref in Capture.kt)

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/LoggerImpl.kt
```diff
@@ -724,10 +726,19 @@ internal class LoggerImpl(
...
-                        sdkConfiguredDuration.loggerImplBuildDuration.toDouble(DurationUnit.MILLISECONDS).toString(),
+                        sdkConfiguredDuration.loggerImplBuildDuration
+                            .toDouble(DurationUnit.MILLISECONDS)
+                            .toString(),
+                    "_pre_init_flush_to_native_duration_ms" to
+                        sdkConfiguredDuration.flushPreInitToNativeDuration
+                            .toDouble(DurationUnit.MILLISECONDS)
+                            .toString(),
                     "_session_replay_enabled" to isSessionReplayEnabled.toString(),
+                    "_pre_init_dropped_call_count" to preInitDroppedCallCount.toString(),
```
FranAguilera: @snowp Also added the pre init dropped call count here so we can query easily if needed later

snowp: nice this stuff is all useful, thanks

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/network/okhttp/CaptureOkHttpEventListenerFactory.kt
```diff
@@ -33,18 +32,19 @@ import okhttp3.EventListener
...
     private val requestFieldProvider: OkHttpRequestFieldProvider,
     private val responseFieldProvider: OkHttpResponseFieldProvider,
 ) : EventListener.Factory {
-    private val configuredPropagationMode by lazy {
-        TracePropagationMode.fromRuntimeValue(
-            runtimeProvider.getRuntimeStringConfigValue(RuntimeStringConfig.TRACE_PROPAGATION_MODE),
-        )
-    }
-    private val requestIgnorePolicy by lazy { RuntimeOkHttpRequestIgnorePolicy(runtimeProvider) }
+    private val configuredPropagationMode: TracePropagationMode
```
snowp: This is fixed but my agent flagged a similar thing in CaptureOkHttpTracingInterceptor with 

```
private val requestIgnorePolicy by lazy {
    RuntimeOkHttpRequestIgnorePolicy(runtimeProvider)
}
```

snowp (commented):
Just one small thing otherwise this lgtm! Thanks for iterating

platform/jvm/capture/src/main/kotlin/io/bitdrift/capture/network/okhttp/CaptureOkHttpEventListenerFactory.kt
FranAguilera: uops great catch, will update

snowp (approved):
lgtm, thanks for the iterations!

PR #1211, opened 2026-09-16
Centralize Taplo configuration

snowp:
Moves Taplo configuration ownership to the monorepo root and makes the local formatter command consume that shared policy. The corresponding monorepo change updates formatting commands to use the shared policy.


FranAguilera (approved):

snowp:
Closing for now: TOML formatting remains monorepo-local.

